PR #41, opened 2022-08-04
There needs to be a way to execute Rule CG0664

alexfurmenkov:
The idea of this PR is to make it possible validate the variables order against the library metadata. The suggested approach is to use a new rule type (implementing it as an operation might be too complex).

From the validation perspective the algorithm is:
1. Create a df that has one row and maps the column name to its order.
2. Expect that a rule contains only one condition holding the operator without a target and comparator.
3. Request metadata from the library knowing the dataset class and domain.
4. Create a list of rule conditions for each variable, use variable name as a target and "ordinal" as a comparator.
5. Validate the resulting rule against the df from step 1.

### DIFF
--- a/engine/enums/rule_types.py
+++ b/engine/enums/rule_types.py
@@ -8,6 +8,9 @@ class RuleTypes(BaseEnum):
     RANGE_CHECK = "Range & Limit"
     DATASET_METADATA_CHECK = "Dataset Metadata Check"
     DATASET_METADATA_CHECK_AGAINST_DEFINE = "Dataset Metadata Check against Define XML"
+    VARIABLE_ORDER_AGAINST_LIBRARY_METADATA_CHECK = (
+        "VARIABLE_ORDER_AGAINST_LIBRARY_METADATA_CHECK"
+    )
     VARIABLE_METADATA_CHECK_AGAINST_DEFINE = (
         "Variable Metadata Check against Define XML"
     )
--- a/engine/rules_engine.py
+++ b/engine/rules_engine.py
@@ -38,6 +38,7 @@ from engine.utilities.utils import (
     is_split_dataset,
     get_corresponding_datasets,
     serialize_rule,
+    search_in_list_of_dicts,
 )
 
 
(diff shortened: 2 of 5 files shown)


engine/enums/rule_types.py
```diff
@@ -8,6 +8,9 @@ class RuleTypes(BaseEnum):
     RANGE_CHECK = "Range & Limit"
     DATASET_METADATA_CHECK = "Dataset Metadata Check"
     DATASET_METADATA_CHECK_AGAINST_DEFINE = "Dataset Metadata Check against Define XML"
+    VARIABLE_ORDER_AGAINST_LIBRARY_METADATA_CHECK = (
+        "VARIABLE_ORDER_AGAINST_LIBRARY_METADATA_CHECK"
```
nhaydel, line 12: This should be a plain text string like the other rule types:
`Variable Order Check` or something similar

engine/rules_engine.py
```diff
@@ -398,6 +399,56 @@ def validate_dataset_contents_against_library_metadata(
...
+                f"standard={self.standard}, version={self.standard_version}, class={dataset_class}, domain={domain}"
+            )
+
+        # fill rule conditions with expected variable ordinal
+        targets: List[str] = []
+        comparator: dict = {}
+        for variable in domain_metadata["datasetVariables"]:
+            targets.append(variable["name"])
+            comparator[variable["name"]] = variable["ordinal"]
+        self.rule_processor.create_list_of_conditions_for_each_target(rule, targets)
```
nhaydel, line 441: I believe these methods should return something instead of modifying the rule, but that can be handled in a different work item if we want to go that route since you did not add that in this PR

nhaydel:
Would it be easier to implement as an operation if we had an operation for getting the array of columns, an operation for getting the array of variables in the domain and a business rules operator to check if two arrays are equal? I think those all could be useful for other rules as well and wouldn’t require a new rule type

engine/enums/rule_types.py
alexfurmenkov: Oh, yes, I forgot about that

engine/rules_engine.py
alexfurmenkov: We already use these methods with the assumption that they change the rule by reference. We can of course change them to return a modified rule in a separate PR but let's first understand if this change benefits us

alexfurmenkov:
> Would it be easier to implement as an operation if we had an operation for getting the array of columns, an operation for getting the array of variables in the domain and a business rules operator to check if two arrays are equal? I think those all could be useful for other rules as well and wouldn’t require a new rule type

@nhaydel let me ensure I understand you right. You are suggesting to have 2 separate operations: first will return a list of dataset columns in accordance to their order and the second will get the similar list of columns not from the dataset but from the library metadata? Finally, `equql_to` operator will be used to check the equality of these lists. Am I missing something?

nhaydel:
> > Would it be easier to implement as an operation if we had an operation for getting the array of columns, an operation for getting the array of variables in the domain and a business rules operator to check if two arrays are equal? I think those all could be useful for other rules as well and wouldn’t require a new rule type
> 
> @nhaydel let me ensure I understand you right. You are suggesting to have 2 separate operations: first will return a list of dataset columns in accordance to their order and the second will get the similar list of columns not from the dataset but from the library metadata. Finally, `equal_to` operator will be used to check the equality of these lists. Am I missing something?

Yes, that is what I'm suggesting. It would avoid us having to modify the conditions

alexfurmenkov:
> > > Would it be easier to implement as an operation if we had an operation for getting the array of columns, an operation for getting the array of variables in the domain and a business rules operator to check if two arrays are equal? I think those all could be useful for other rules as well and wouldn’t require a new rule type
> > 
> > 
> > @nhaydel let me ensure I understand you right. You are suggesting to have 2 separate operations: first will return a list of dataset columns in accordance to their order and the second will get the similar list of columns not from the dataset but from the library metadata. Finally, `equal_to` operator will be used to check the equality of these lists. Am I missing something?
> 
> Yes, that is what I'm suggesting. It would avoid us having to modify the conditions

Okay, let me try this approach and compare the pros and cons

alexfurmenkov:
@nhaydel I agree with your idea, it should make things much easier. I think I will create a new branch from my refactoring PR (https://github.com/cdisc-org/cdisc-rules-engine/pull/75) since both of them modify the same place in the code.

alexfurmenkov:
@nhaydel the only drawback is that the reports will be reported for all rows of a dataset. I think that it the future it can be mitigated by specifying the errors limit as we did in the Azure version.

nhaydel:
> @nhaydel the only drawback is that the errors will be reported for all rows of a dataset. I think that it the future it can be mitigated by specifying the errors limit as we did in the Azure version.

I believe if the rule specifies dataset sensitivity only one error will be reported

alexfurmenkov:
@nhaydel closing this PR, created a fresh one - https://github.com/cdisc-org/cdisc-rules-engine/pull/80

Issue #66, opened 2022-08-12
Add JSON output format


DmitryMK:
Currently the only outputs format for the report is XLSX. Add an option to control output format and add JSON format.

DmitryMK:
The functionality was implemented in PR 69.

DmitryMK:
The functionality was implemented in PR 69.

PR #69, opened 2022-08-14
Adding JSON output format

DmitryMK:
Merging into the development branch now, as we previously discussed that after 1.0.0 we will stop merging into main directly.

This PR adds 2 options to address https://github.com/cdisc-org/cdisc-rules-engine/issues/66 :
--output-format
--raw-report

By default it will save JSON with the same structure/fields as the Excel report. But for development purposes it might be needed to get results as they are reported by the engine, the --raw-report flag can be used for this purpose.

I created a GenericReport class from the ExcelReport, which is not inherited by both ExcelReport and JsonReport classes. The main reason is to be consistent, so that in case we change how the report looks like in Excel, it will be consistently updated for the JSON.


cdisc_rules_engine/utilities/excel_report.py
```diff
@@ -162,3 +51,21 @@ def get_excel_export(self, define_version, cdiscCt, standard, version) -> Workbo
...
+                self._args.standard,
+                self._args.version.replace("-", "."),
+            )
+            with open(output_name, "wb") as f:
+                f.write(excel_workbook_to_stream(report_data))
+        except Exception as e:
+            logger.error(e)
+            raise e
+        finally:
+            self._template.close()
```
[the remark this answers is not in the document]
DmitryMK: It might be a bad design, but it is opened in ReportFactory (line 41 - self._data_service.read_data(..)). I kept it this way for backward compatibility with tests, but I can rewrite ti, so that the template is opened within a class instance and not passed as a parameter.

core.py
```diff
@@ -90,6 +73,21 @@ def cli():
     default=generate_report_filename(datetime.now().isoformat()),
     help="Report output file destination",
 )
+@click.option(
+    "-of",
+    "--output-format",
+    default="XLSX",
+    type=click.Choice(["JSON", "XLSX"], case_sensitive=False),
```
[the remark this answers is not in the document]
DmitryMK: Fixed

cdisc_rules_engine/utilities/report_factory.py
```diff
@@ -0,0 +1,57 @@
...
+                    self._args.report_template, "rb"
+                )
+            return ExcelReport(
+                self._data_path,
+                self._results,
+                self._elapsed_time,
+                self._args,
+                self._template,
+            )
+        elif output_type == "JSON":
```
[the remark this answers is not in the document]
DmitryMK: Fixed

cdisc_rules_engine/utilities/report_factory.py
```diff
@@ -0,0 +1,57 @@
...
+        self._results = results
+        self._elapsed_time = elapsed_time
+        self._args = args
+        self._data_service = data_service
+
+    def get_report_service(self):
+        output_type = self._args.output_format.upper()
+        if output_type == "XLSX":
+            if self._data_service != None:
+                self._template = self._data_service.read_data(
```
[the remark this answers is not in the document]
DmitryMK: Fixed

cdisc_rules_engine/utilities/base_report.py
```diff
@@ -0,0 +1,167 @@
+from typing import List, TextIO, Union
+from openpyxl import Workbook
+from cdisc_rules_engine.models.rule_validation_result import RuleValidationResult
+from cdisc_rules_engine.models.validation_args import Validation_args
+from cdisc_rules_engine.services.data_services.local_data_service import (
+    LocalDataService,
+)
+
+
+class BaseReport:
```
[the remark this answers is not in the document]
DmitryMK: Updated, also used abstractmethod decorators for the methods which need to be implemented.

cdisc_rules_engine/utilities/excel_report.py
```diff
@@ -20,128 +22,15 @@ def __init__(
         data_path: str,
         validation_results: List[RuleValidationResult],
         elapsed_time: float,
-        report_template: TextIO,
+        args: Validation_args,
+        _template: TextIO,
```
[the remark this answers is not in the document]
DmitryMK: Fixed

core.py
```diff
@@ -121,6 +121,13 @@ def validate(
 
     python core.py -s SDTM -v 3.4 -d /path/to/datasets
     """
+
+    # Validate conditional options
+    logger = logging.getLogger("validator")
+    if raw_report == True and output_format.lower() != "json":
```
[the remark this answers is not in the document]
DmitryMK: Fixed

cdisc_rules_engine/utilities/base_report.py
```diff
@@ -0,0 +1,167 @@
...
+                rules_report.append(rules_item)
+
+        return sorted(
+            rules_report,
+            key=lambda x: x[0] if (self._item_type == "list") else x["rule_id"],
+        )
+
+    @abstractmethod
+    def get_export() -> Union[dict, Workbook]:
+        raise NotImplementedError
```
alexfurmenkov, line 163: `abstractmethod` decorator and raising a `NotImplementedError` are 2 different ways to do the same thing but they should not be combined. 

The difference is that raising a `NotImplementedError` is the old style whereas `abstractmethod` decorator is the new style. A method marked with `abstractmethod` decorator must not contain any implementation (raising an error is also an implementation), the method can only contain `pass` keyword or a docstring. 

In terms of our project, we tend to use the new style but some examples of raising a `NotImplementedError` still exist in the project and my suggestion would be to unify the style later, not in this PR.

cdisc_rules_engine/utilities/report_factory.py
```diff
@@ -0,0 +1,58 @@
...
+        self._data_path = data_path
+        self._results = results
+        self._elapsed_time = elapsed_time
+        self._args = args
+        self._data_service = data_service
+
+    def get_report_service(self):
+        output_type = self._args.output_format.upper()
+        if output_type == ReportTypes.XLSX.value:
+            if self._data_service != None:
```
alexfurmenkov, line 41: self._data_service can't be None since it is a mandatory argument in the class init. We can remove this condition safely.

cdisc_rules_engine/utilities/base_report.py
```diff
@@ -0,0 +1,167 @@
...
+        rules_report = []
+        for validation_result in self._results:
+            rules_item = {
+                "rule_id": validation_result.id,
+                "version": "1",
+                "message": validation_result.message,
+                "status": "SUCCESS"
+                if validation_result.execution_status == "success"
+                else "SKIPPED",
+            }
```
alexfurmenkov, line 150: `Please` use ExecutionStatus enum on lines 148, 9

cdisc_rules_engine/utilities/report_factory.py
DmitryMK: It is done for backward compatibility with tests, which provide None as a data_service value. I understand, that probably it will be better to rewrite test, rather than having this condition.

cdisc_rules_engine/utilities/base_report.py
DmitryMK: Aleksei for the explanation, will update it. I think that flake8 will catch it as well, so if we have it, that might help to avoid such issues.

cdisc_rules_engine/utilities/report_factory.py
alexfurmenkov: Yes, or in test you just can pass a mock

cdisc_rules_engine/utilities/base_report.py
DmitryMK: Updated

cdisc_rules_engine/utilities/report_factory.py
DmitryMK: Updated the test and remove the check for None from the report_factory.

cdisc_rules_engine/utilities/base_report.py
DmitryMK: Updated it for lines 147, 148, 149.

cdisc_rules_engine/utilities/json_report.py
```diff
@@ -0,0 +1,41 @@
+from datetime import datetime
+from typing import List, TextIO
+from cdisc_rules_engine.models.rule_validation_result import RuleValidationResult
+from cdisc_rules_engine.utilities.generic_report import GenericReport
+
+
+class JsonReport(GenericReport):
```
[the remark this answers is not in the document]
DmitryMK: Added test_json_export.py

alexfurmenkov (approved):
Approving. Please change the destination branch to main since we have agreed to merge all open PRs and then start creating PRs to development.

nhaydel (approved):
Sorry for the late response, didn't realize I requested changes on this. Can we merge this into development though

alexfurmenkov:
While the PR was hanging, we have again agreed to merge into development. Can you please change the destination branch and do a git pull from development into this branch?

DmitryMK:
@alexfurmenkov will merge to development. I have a problem with xmlschema logging - it prints a lot of messages and I think this was introduced in this branch. So I will merge once I fix this issue.

DmitryMK:
@alexfurmenkov @nhaydel I have addressed flake8 issues for the files which were updated in this branch (there are other files which are failing flake8 checks). Mainly I just want to confirm that it is fine that I moved `logging.getLogger(XXX).disabled = True` to logging_service_factory.py.
The problem was that flake8 issued an error that import must be at top level because these statements were between imports in run_validation.py. So I had to move it to a new location.

alexfurmenkov:
Got it

PR #75, opened 2022-08-17
Refactor perform_rule_operations method

alexfurmenkov:
I saw several areas of improvement of `perform_rule_operations` method:

1. The operations in data processor have different interfaces and this is why we maintain several mappings like `domain_operator_map`, `study_operator_map` etc. Besides having the mappings we also have a branching logic to properly call the operation depending on its area: domain, study, dictionary. I have unified the operations interface to depend on OperationParams object which holds all necessary input parameters for the operation. Now, all operations are called with the same input parameters and no branching logic and mappings are required.
2. The second step to get rid of these huge mappings is to rename the operation function names and invoke them like `getattr(data_processor, operation_name)` (as we do with operators in business rules). In this case we should raise an error if the operation is not found. From the user and rule perspective, no changes are expected.
3. Handling operation results also contains if/else branches and caused by the fact that all operations return different data types:`elif isinstance(result, np.ndarray)`, `elif isinstance(result, dict)`. I understand that it is almost impossible to make all operations return the same data type since they serve different purposes but my suggestion in terms of this PR is to make all operations return the data type that can be assigned to a dataframe column without additional processing, so no branching logic is needed when handling the result.


cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -262,121 +200,184 @@ def valid_meddra_term_references(self, dataframe, target, domain, **kwargs):
...
         valid_code_term_pairs = self.cache.get(cache_key)
         if not valid_code_term_pairs:
-            terms: dict = self.cache.get(dictionaries_path)
+            terms: dict = self.cache.get(params.meddra_path)
             valid_code_term_pairs = MedDRATerm.get_code_term_pairs(terms)
             self.cache.add(cache_key, valid_code_term_pairs)
         column = str(uuid4()) + "_pairs"
-        dataframe[column] = list(zip(dataframe[columns[0]], dataframe[columns[1]]))
-        result = dataframe[column].isin(valid_code_term_pairs[term_type])
+        params.dataframe[column] = list(
```
sityaevDI, line 298: [we can use list comprehension and save couple msec](https://stackoverflow.com/questions/5790860/and-vs-list-and-dict-which-is-better), but it more of coding style for me

cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -262,121 +200,184 @@ def valid_meddra_term_references(self, dataframe, target, domain, **kwargs):
...
-        dataframe[column] = dataframe[code_strings].agg("/".join, axis=1)
-        result = dataframe[column].isin(valid_term_hierarchies)
+        params.dataframe[column] = params.dataframe[code_strings].agg("/".join, axis=1)
+        result = params.dataframe[column].isin(valid_term_hierarchies)
         return result
 
-    def valid_meddra_code_term_pairs(self, dataframe, target, domain, **kwargs):
-        dictionaries_path: str = kwargs.get("meddra_path")
-        if not dictionaries_path:
+    def valid_meddra_code_term_pairs(self, params: OperationParams) -> pd.Series:
```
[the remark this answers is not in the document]
alexfurmenkov: Currently, some method names in data processor are the actual operation names in the rules which are currently used since we are using getattr approach. Changing the function name will also lead to changing the rules which is a major change.

cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -262,121 +200,184 @@ def valid_meddra_term_references(self, dataframe, target, domain, **kwargs):
...
+        if not params.meddra_path:
             raise ValueError("Can't execute the operation, no meddra path provided")
         variable_pair_map = {
-            f"{domain}{MedDRAVariables.SOC.value}": (
+            f"{params.domain}{MedDRAVariables.SOC.value}": (
```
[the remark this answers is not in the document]
alexfurmenkov: Good idea, I think we can make it an enum. Will do, thanks

cdisc_rules_engine/utilities/data_processor.py
alexfurmenkov: Not sure:

```
timeit.timeit("list(zip([1, 2, 3], [4, 5, 6]))")
Out[9]: 0.4584283690000035
timeit.timeit("[x for x in zip([1, 2, 3], [4, 5, 6])]")
Out[10]: 0.5141287360000035
```

cdisc_rules_engine/utilities/data_processor.py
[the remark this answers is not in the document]
alexfurmenkov: It was hard to create an enum out of this mapping since it requires input parameters, so I have extracted it into a separate method `_get_columns_by_meddra_variable_name`

cdisc_rules_engine/models/operation_params.py
```diff
@@ -0,0 +1,26 @@
+from dataclasses import dataclass
+from typing import List
+
+import pandas as pd
+
+
+@dataclass
+class OperationParams:
```
nhaydel, line 8: I like this. As a rule we should probably create dataclasses for all complex function parameters

cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -41,196 +42,136 @@ def __init__(self, data_service=None, cache: CacheServiceInterface = None):
             data_service or DataServiceFactory(config, self.cache).get_data_service()
         )
 
-    @staticmethod
-    def calc_min(dataframe, target, grouping: List = None, **kwargs):
-        if not grouping:
-            return dataframe[target].min()
+    def min(self, params: OperationParams) -> Union[int, float, pd.DataFrame]:
```
nhaydel, line 45: PR looks really good. My only  suggestion would be that we may want to do something like we did with the data services, and make operations a class by themselves and allow users to register operations. That way the language itself is extensible. It is quite a big change so I could see it being a new issue if we want to do it

nhaydel (approved):
Approved with suggestions

PR #76, opened 2022-08-18
Reduce version to 0.3.0

nhaydel:
Reduce version


nhaydel:
Merging to get new version released

PR #80, opened 2022-08-19
There needs to be a way to execute Rule CG0664

alexfurmenkov:
This PR adds 2 new operations for validating column order against library metadata: `get_column_order_from_dataset` and `get_column_order_from_library`. 

Rule example:
```
{
   "core_id":"MockRule",
   "standards":[
      {
         "Name":"SDTMIG",
         "Version":"3.3"
      }
   ],
   "classes":{
      "Include":[
         "Events"
      ]
   },
   "domains":{
      "Include":[
         "AE"
      ]
   },
   "sensitivity": "Dataset",
   "operations":[
      {
         "operator":"get_column_order_from_dataset",
         "domain":"AE",
         "id":"$column_order_from_dataset"
      },
      {
         "operator":"get_column_order_from_library",
         "domain":"AE",
         "id":"$column_order_from_library"
      }
   ],
   "conditions": {
      "any":[
         {
            "name":"get_dataset",
            "operator":"not_equal_to",
            "value":{
               "target":"$column_order_from_dataset",
               "comparator":"$column_order_from_library"
            }
         }
      ]
   },
   "actions":[
      {
         "name":"generate_dataset_error_objects",
         "params":{
            "message":"Order of variables is invalid"
         }
      }
   ],
   "output_variables":[
      "$column_order_from_dataset",
      "$column_order_from_library"
   ]
}
```

The idea is that `get_column_order_from_dataset` operation returns a list of dataset columns in accordance to their order and `get_column_order_from_library` gets the similar list of columns not from the dataset but from the library metadata. Finally, `not_equal_to` operator is used to check the equality of these lists.


cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -327,6 +375,29 @@ def _get_columns_by_meddra_variable_name(
...
         return variable_pair_map.get(params.target)
 
+    def _get_metadata_for_class_and_domain(
+        self, dataset_class: str, domain: str, standard: str, standard_version: str
+    ) -> Optional[dict]:
+        """
+        Gets metadata for the given class and domain from cache.
+        The cache stores CDISC Library metadata.
+        """
+        standard_details: dict = (
```
nhaydel, line 385: Since this is a standalone application now and the cache may not always be populated do we think it's a good idea to start making requests to the library if the data is not found in cache? I don't think that would have to be updated in this PR, but an issue should be created for it if we want to start doing that

nhaydel (approved):

nhaydel:
Based on some information I got from the standards developers. Checking order is not as simple as getting it from the domain. Here is the logic we agreed upon that should work in all cases:
Given a domain and that domains class. Check the parent **MODEL** for ordering using the logic as follows:
1. If the model has the domain, pull the expected order from the model domain
2. If the model does not have the domain, pull the expected order from the model class. If the class is a general observations class (Findings, Events, Interventions) Add the Identifier Variables from the general observations class to the beginning of the list of variables and the Timing Variables from the general observations class to the end of the list of variables.
3. Verify that the order of variables matches the order specified in the resulting list.
Note: Some variables are permissible so they could be excluded and the variable order still be correct.

Ex:
```
// Validating variable order for a domain in a general observation class
// Suppose this is the variable order specified in the model class
{
  --TESTCD: 1
  --STRESC: 2
}
// And this is the variable order specified in the general observations class
{
  IDVAR: 1
  TIMINGVAR: 2
}

// Then the correct order of variables to validate against would be:
{
  IDVAR: 1
  --TESTCD: 2
  --STRESC: 3
 TIMINGVAR: 4
}

Now assuming that TESTCD is permissible the following variable order is still valid
[IDVAR, LBSTRESC, TIMINGVAR]
This is because the ordinals associated with those variables are in order: [1, 3, 4]
```

alexfurmenkov:
> Based on some information I got from the standards developers. Checking order is not as simple as getting it from the domain. Here is the logic we agreed upon that should work in all cases: Given a domain and that domains class. Check the parent **MODEL** for ordering using the logic as follows:
> 
> 1. If the model has the domain, pull the expected order from the model domain
> 2. If the model does not have the domain, pull the expected order from the model class. If the class is a general observations class (Findings, Events, Interventions) Add the Identifier Variables from the general observations class to the beginning of the list of variables and the Timing Variables from the general observations class to the end of the list of variables.
> 3. Verify that the order of variables matches the order specified in the resulting list.
>    Note: Some variables are permissible so they could be excluded and the variable order still be correct.
> 
> Ex:
> 
> ```
> // Validating variable order for a domain in a general observation class
> // Suppose this is the variable order specified in the model class
> {
>   --TESTCD: 1
>   --STRESC: 2
> }
> // And this is the variable order specified in the general observations class
> {
>   IDVAR: 1
>   TIMINGVAR: 2
> }
> 
> // Then the correct order of variables to validate against would be:
> {
>   IDVAR: 1
>   --TESTCD: 2
>   --STRESC: 3
>  TIMINGVAR: 4
> }
> 
> Now assuming that TESTCD is permissible the following variable order is still valid
> [IDVAR, LBSTRESC, TIMINGVAR]
> This is because the ordinals associated with those variables are in order: [1, 3, 4]
> ```

Thanks for the clarification, I would only like to understand where can I take the parent **MODEL**? As I understand, it is not the standard details that I am using right now.

nhaydel:
@alexfurmenkov we have not been caching the model metadata because people would be using implementations instead of models. However, every implementation guide (SDTMIG 3.4, SDTMIG 3.3, etc...) implements a model. The model it implements can be found in the model link in the response when requesting the implementation guide from the library:
```
{
  "implementedBy": [],
  "model": {
    "href": "/mdr/sdtm/2-0",
    "title": "Study Data Tabulation Model Version 2.0",
    "type": "Foundational Model"
  },
  "priorVersion": {
    "href": "/mdr/sdtmig/3-3",
    "title": "Study Data Tabulation Model Implementation Guide: Human Clinical Trials",
    "type": "Implementation Guide"
  },
  "self": {
    "href": "/mdr/sdtmig/3-4",
    "title": "Study Data Tabulation Model Implementation Guide: Human Clinical Trials",
    "type": "Implementation Guide"
  }
}
```
We likely will have to add some data to our cache, and pull data from the models. I would suggest updating the cache feeder in a separate issue, and having that issue block this PR.

alexfurmenkov:
> @alexfurmenkov we have not been caching the model metadata because people would be using implementations instead of models. However, every implementation guide (SDTMIG 3.4, SDTMIG 3.3, etc...) implements a model. The model it implements can be found in the model link in the response when requesting the implementation guide from the library:
> 
> ```
> {
>   "implementedBy": [],
>   "model": {
>     "href": "/mdr/sdtm/2-0",
>     "title": "Study Data Tabulation Model Version 2.0",
>     "type": "Foundational Model"
>   },
>   "priorVersion": {
>     "href": "/mdr/sdtmig/3-3",
>     "title": "Study Data Tabulation Model Implementation Guide: Human Clinical Trials",
>     "type": "Implementation Guide"
>   },
>   "self": {
>     "href": "/mdr/sdtmig/3-4",
>     "title": "Study Data Tabulation Model Implementation Guide: Human Clinical Trials",
>     "type": "Implementation Guide"
>   }
> }
> ```
> 
> We likely will have to add some data to our cache, and pull data from the models. I would suggest updating the cache feeder in a separate issue, and having that issue block this PR.

Okay, thanks. I think that we should implement this in terms of this PR and Issue since it is about supporting this rule. I will work on it.

alexfurmenkov:
@nhaydel we need to request the model for each standard, right?

nhaydel:
> @nhaydel we need to request the model for each standard, right?

yes. The products endpoint also provides a list of models if you want to just preemptively load them all

alexfurmenkov:
> > @nhaydel we need to request the model for each standard, right?
> 
> yes. The products endpoint also provides a list of models if you want to just preemptively load them all

@nhaydel updated the logic to cache model metadata and use it in the operation. What do you think, should we update existing pkl files with model metadata?

cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -170,6 +183,38 @@ def variable_names(self, params: OperationParams) -> set:
...
+        # get variables metadata from the standard model
+        variables_metadata: List[
+            dict
+        ] = self._get_variables_metadata_from_standard_model(
+            dataset_class, params.domain, params.standard, params.standard_version
+        )
+
+        # create a list of variable names in accordance to the "ordinal" key
+        variables_metadata.sort(key=lambda item: item["ordinal"])
+        variable_names_list = [
```
nhaydel, line 213: There is one more step to getting the correct order. If the class is one of Interventions, Findings, or Events then the  General Observation class variables need to be added to the variable names list. This should be done by adding the variables with role "Identifier" to the beginning of the list and variables with role "Timing" to the end of the list. If we are not caching the role, we should be.

nhaydel: Also, this operation probably should actually do the comparison, because some variables can be omitted. So it should  check whether or not the dataframe columns are sorted based on the supplied variables ordinals. Not sure if you are doing this elsewhere though

alexfurmenkov: Okay, I understand, I will modify the logic. 

The actual comparison is done in "conditions", the operations just return the variable list in a given order and the conditions compare the columns. You can check the example rule in conftest.py. Is this approach okay for you? I want the operation to just return a list and not compare anything so that we could reuse its result for further validation scenarios if needed.

nhaydel: That approach is ok with me if it can handle missing variables that are allowed to be missing. For example if the returned order is ["STUDYID", "DOMAIN", "USUBJID", "VAR1", "VAR2", "VAR3", "VAR4"] but VAR3 is permissible then this is still a valid variable order for the dataset to have ["STUDYID", "DOMAIN", "USUBJID", "VAR1", "VAR2", "VAR4"]. To be clear I don't think this operation/check needs to know what variables are permissible or not, but it should be able to tell if any set of variables is in order based on their ordinals.

alexfurmenkov: Updated the logic

cdisc_rules_engine/operations/operations_factory.py
```diff
@@ -1,29 +1,32 @@
-from cdisc_rules_engine.operations.distinct import Distinct
```
alexfurmenkov, line 1: Nothing critical here, just sorted according to PEP

cdisc_rules_engine/utilities/data_processor.py
nhaydel: I don't see how you would write an rule that allows for permissible variables to be missing. Can you provide an example?

alexfurmenkov: There is no such option right now, I would suggest implementing it in a separate PR not to complicate this one

nhaydel: That works for me. Approved

nhaydel (approved):

PR #82, opened 2022-08-21
Add progress bar

DmitryMK:
* Adding progress bar
* Adding --disable-progressbar option.
* Updating option names to use - instead of _.

I used imap_unordered instead of map, which should not cause problems, as ordering of the results is done in the reporting part.
Removed pool.close and poo.join as now pool is managed with context manager (with statement).


scripts/run_validation.py
```diff
@@ -173,13 +175,32 @@ def run_validation(args: namedtuple):
...
-    pool = Pool(args.pool_size)
-    results = pool.map(
-        partial(validate_single_rule, shared_cache, data_path, datasets, args),
-        rules,
-    )
-    pool.close()
-    pool.join()
+    results = []
+
+    # run each rule in a separate thread
```
alexfurmenkov, line 180: Not thread but a process, it is important to understand the difference

alexfurmenkov:
Please edit the misleading comment, the rest is fine

README.md
```diff
@@ -28,15 +28,16 @@ Run `core.py validate --help` to see the list of validation options.
 * -ca, --cache TEXT                          Relative path to cache files containing pre loaded metadata and rules
 * -p, --pool_size INTEGER                    Number of parallel processes for validation
```
nhaydel, line 29: pool size is not pool-size

scripts/run_validation.py
```diff
@@ -173,13 +175,32 @@ def run_validation(args: namedtuple):
...
+    with Pool(args.pool_size) as pool:
+        if args.disable_progressbar is True:
+            for rule_result in pool.imap_unordered(
```
nhaydel, line 183: can we not just say results = pool.imap_unordered...

DmitryMK: We need to have it in a for loop, as we need to iterate the progress bar.

DmitryMK: I'm wrong, this is the part which avoids progressbar. Still imap_unordered returns an iterable rather than a list. I've tried it as `results = pool.imap_unordered...` and the process was just frozen at some point (although there were no issues).

scripts/run_validation.py
DmitryMK: fixed

README.md
DmitryMK: fixed

alexfurmenkov (approved):

nhaydel (approved):

PR #84, opened 2022-08-22
Remove unnecessary config variables, add objects to __init__

alexfurmenkov:


cdisc_rules_engine/config/config.py
```diff
@@ -20,10 +20,6 @@ def __new__(cls):
                 "REDIS_HOST_NAME",
                 "REDIS_ACCESS_KEY",
                 "CDISC_LIBRARY_API_KEY",
-                "CDISC_LIBRARY_URL",
-                "DICTIONARIES_TABLE_NAME",
-                "WHODRUG_TERMS_TABLE_NAME",
-                "MEDDRA_TERMS_TABLE_NAME",
                 "DATA_SERVICE_TYPE",
```
alexfurmenkov, line 23: CDISC_LIBRARY_URL got changed to a hardcoded value, table names should not be set here

nhaydel:
Looks good. Aren't we merging into the development branch though?

cdisc_rules_engine/config/config.py
DmitryMK: Does it makes sense to keep CDISC Library API customizable? In Parexel we work with the CDISC Library via a mirror.

DmitryMK: It will be a separate PR, so it is more like a question whether we should do it later or not.

alexfurmenkov: Yes, we can implement it later in a separate PR

alexfurmenkov:
> Looks good. Aren't we merging into the development branch though?

Will change the target, thanks

alexfurmenkov:
@nhaydel changed

nhaydel (approved):

PR #85, opened 2022-08-22
Move dummy data service

nhaydel:
This PR fixes the issue `ImportError: cannot import name 'DummyDataService'` when installing the python module.

Steps to test:

1. Clone and build a wheel from this branch: `python -m build`
2. Create a new virtual environment
3. Install the new build in the virtual environment: `pip install .\dist\cdisc_rules_engine-1.0.0-py3-none-any.whl`
4. Verify you can import Dummy Data Service in the virtual environment


cdisc_rules_engine/services/data_services/data_service_factory.py
```diff
@@ -1,4 +1,4 @@
-from cdisc_rules_engine.dummy_services.dummy_data_service import DummyDataService
+from cdisc_rules_engine.services.data_services.dummy_data_service import DummyDataService
```
alexfurmenkov, line 1: Since DummyDataService is in module init, it can be imported from services.data_services directly like

`from cdisc_rules_engine.services.data_services import DummyDataService`

cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -7,7 +7,7 @@
 import pandas as pd
 
 from cdisc_rules_engine.config import config
-from cdisc_rules_engine.dummy_services.dummy_data_service import DummyDataService
+from cdisc_rules_engine.services.data_services.dummy_data_service import DummyDataService
```
alexfurmenkov, line 10: Same

tests/unit/test_dummy_data_service.py
```diff
@@ -3,7 +3,7 @@
 import pytest
 
 from cdisc_rules_engine.dummy_models.dummy_dataset import DummyDataset
-from cdisc_rules_engine.dummy_services.dummy_data_service import DummyDataService
+from cdisc_rules_engine.services.data_services.dummy_data_service import DummyDataService
```
alexfurmenkov, line 6: Same

cdisc_rules_engine/services/data_services/data_service_factory.py
nhaydel: In order to do that the import order in the __init__.py had to be changed. I made that change, hopefully black doesn't complain about it

PR #93, opened 2022-08-29
Move cache population methods to importable class

nhaydel:
This PR moves the cache population methods to a class that implementations such as CORE can import to populate their own cache


cdisc_rules_engine/services/cache/cache_populator_service.py
```diff
@@ -0,0 +1,251 @@
...
+)
+from cdisc_rules_engine.services.cdisc_library_service import CDISCLibraryService
+from cdisc_rules_engine.utilities.utils import (
+    get_library_variables_metadata_cache_key,
+    get_standard_details_cache_key,
+)
+
+
+class CachePopulator:
+    def __init__(self, cache: CacheServiceInterface, api_key: str):
```
alexfurmenkov, line 23: I would suggest to make CDISCLibraryService instance an input parameter to inject this dependency.

cdisc_rules_engine/services/cache/cache_populator_service.py
```diff
@@ -0,0 +1,251 @@
...
+        self.api_key = api_key
+        self.cache = cache
+        self.library_service = CDISCLibraryService(self.api_key, self.cache)
+
+    async def load_cache_data(self):
+        """
+        This function populates an in memory cache with all data necessary for running
```
alexfurmenkov, line 30: Why in memory only? Is it a typo?

cdisc_rules_engine/services/cache/cache_populator_service.py
```diff
@@ -0,0 +1,251 @@
...
+        except LibraryResourceNotFoundException:
+            return None
+        return {
+            "cache_key": get_library_variables_metadata_cache_key(
+                standard_type, standard_version
+            ),
+            **variables_metadata,
+        }
+
+    def save_rules_locally(self, cache_path: str):
```
alexfurmenkov, line 213: Can the public methods be moved up so that they were declared before all encapsulated methods?

alexfurmenkov (approved):

cdisc_rules_engine/services/cache/cache_populator_service.py
nhaydel: Ahh it's the result of copying the method from the old script. Will update

PR #97, opened 2022-09-01
Draft: resolves #96:  more flexible dictionary installation

### LINKED ISSUE #96: refactor dictionary factories
We need to provide way to extend dictionary terms parsing and split logic of parsing terms and saving them

sityaevDI:


.flake8
```diff
@@ -1,6 +1,7 @@
 [flake8]
 max-line-length = 88
 max_complexity = 10
+ignore = E203
```
alexfurmenkov, line 4: Why?

sityaevDI: black and flake8 have conflicts about white spaces before colons

PR #98, opened 2022-09-02
abstract factory for dictionary terms factories

sityaevDI:
split logic of parsing and caching terms
ignore E203 PEP error

### DIFF
--- a/cdisc_rules_engine/models/dictionaries/abstract_factory.py
+++ b/cdisc_rules_engine/models/dictionaries/abstract_factory.py
@@ -11,7 +11,7 @@ from cdisc_rules_engine.services.factory_interface import FactoryInterface
 
 
 class AbstractTermsFactory(FactoryInterface):
-    _service_map: dict = {
+    _registered_services_map: dict = {
         DictionaryTypes.MEDDRA.value: MedDRATermsFactory,
         DictionaryTypes.WHODRUG.value: WhoDrugTermsFactory,
     }
@@ -25,13 +25,14 @@ class AbstractTermsFactory(FactoryInterface):
             raise ValueError("Service name must not be empty!")
         if not issubclass(service, TermsFactoryInterface):
             raise TypeError("Implementation of TermsFactoryInterface required!")
-        cls._service_map[name] = service
+        cls._registered_services_map[name] = service
 
     def get_service(self, name: str, **kwargs) -> TermsFactoryInterface:
-        if name not in self._service_map:
+        if name not in self._registered_services_map:
             raise ValueError(
-                f"Service name must be in  {list(self._service_map.keys())}, "
+                f"Service name must be in"
+                f" {list(self._registered_services_map.keys())}, "
                 f"given service name is {name}"
             )
-        factory = self._service_map.get(name)
+        factory = self._registered_services_map.get(name)
         return factory(data_service=self.data_service)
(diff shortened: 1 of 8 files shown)


cdisc_rules_engine/models/dictionaries/abstract_factory.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @classmethod
+    def register_service(cls, name: str, service: Type[TermsFactoryInterface]) -> None:
+        if not name:
+            raise ValueError("Service name must not be empty!")
+        if not issubclass(service, TermsFactoryInterface):
+            raise TypeError("Implementation of TermsFactoryInterface required!")
+        cls._service_map[name] = service
+
+    def get_service(self, name: str, **kwargs):
```
alexfurmenkov, line 30: add return type

cdisc_rules_engine/models/dictionaries/get_dictionary_terms.py
```diff
@@ -0,0 +1,16 @@
+from . import DictionaryTypes, TermsFactoryInterface
+from .abstract_factory import AbstractTermsFactory
+from cdisc_rules_engine.services.data_services import BaseDataService
+
+
+def get_dictionary_terms_from_folder(
```
alexfurmenkov, line 6: Let's call it "install_dictionary_terms_from_folder"

alexfurmenkov (approved):

alexfurmenkov (approved):

cdisc_rules_engine/models/dictionaries/abstract_factory.py
```diff
@@ -0,0 +1,37 @@
...
+    _service_map: dict = {
+        DictionaryTypes.MEDDRA.value: MedDRATermsFactory,
+        DictionaryTypes.WHODRUG.value: WhoDrugTermsFactory,
+    }
+
+    def __init__(self, data_service: BaseDataService):
+        self.data_service = data_service
+
+    @classmethod
+    def register_service(cls, name: str, service: Type[TermsFactoryInterface]) -> None:
```
nhaydel, line 23: These are factories, so it might be more clear to name these functions: register_factory, get_factory and the map: _factory_map

cdisc_rules_engine/models/dictionaries/abstract_factory.py
```diff
@@ -0,0 +1,37 @@
...
+    DictionaryTypes,
+    TermsFactoryInterface,
+)
+from cdisc_rules_engine.models.dictionaries.meddra import MedDRATermsFactory
+from cdisc_rules_engine.models.dictionaries.whodrug import WhoDrugTermsFactory
+from cdisc_rules_engine.services.data_services import BaseDataService
+from cdisc_rules_engine.services.factory_interface import FactoryInterface
+
+
+class AbstractTermsFactory(FactoryInterface):
```
nhaydel, line 13: Should the MedDRATermsFactory and WhoDrugTermsFactory inherit from AbstractTermsFactory?

nhaydel (changes requested):
Few minor comments

cdisc_rules_engine/models/dictionaries/abstract_factory.py
nhaydel: Nevermind, I see this is more of a factory factory (A factory that produces factories.)

cdisc_rules_engine/models/dictionaries/abstract_factory.py
sityaevDI: I suggest to name it like`_registered_services_map` because _factory_map and _service_map don't describe field purpose as it should. What do you think?

nhaydel: That works for me

PR #100, opened 2022-09-05
Expand operation functionality using plugin architecture

nhaydel:
This PR adds functionality to register operations similar to how we register data services. There are still tests to add, but the functionality is there and all existing tests are passing.


cdisc_rules_engine/operations/operation_interface.py
```diff
@@ -0,0 +1,47 @@
...
+from cdisc_rules_engine.models.operation_params import OperationParams
+from abc import ABC, abstractmethod
+
+from cdisc_rules_engine.services.cache.cache_service_interface import (
+    CacheServiceInterface,
+)
+from cdisc_rules_engine.services.data_services.base_data_service import BaseDataService
+
+
+class OperationInterface(ABC):
```
alexfurmenkov, line 11: Interface must not contain any implementation

cdisc_rules_engine/operations/operations_factory.py
```diff
@@ -0,0 +1,82 @@
...
+        "variable_exists": VariableExists,
+        "variable_value_count": VariableValueCount,
+        "variable_names": VariableNames,
+        "valide_meddra_code_references": ValidMeddraCodeReferences,
+        "valid_whodrug_references": ValidWhodrugReferences,
+        "valid_meddra_term_references": ValidMeddraTermReferences,
+        "valid_meddra_code_term_pairs": ValidMeddraCodeTermPairs,
+    }
+
+    def __init__(self):
```
alexfurmenkov, line 53: What is the purpose of defining an empty init? Can we just remove its definition?

cdisc_rules_engine/utilities/rule_processor.py
```diff
@@ -262,7 +253,7 @@ def perform_rule_operations(
         return dataset_copy
 
     def _execute_operation(
-        self, operation_to_call: Callable, operation_params: OperationParams
+        self, operation_params: OperationParams, dataset: pd.DataFrame
```
alexfurmenkov, line 256: dataset exists in OperationParams

cdisc_rules_engine/operations/operation_interface.py
```diff
@@ -0,0 +1,47 @@
...
+        self.cache = cache_service
+        self.data_service = data_service
+        self.evaluation_dataset = dataset
+
+    @abstractmethod
+    def execute(self) -> pd.DataFrame:
+        """Perform operation calculations."""
+        pass
+
+    def _handle_operation_result(self, result) -> pd.DataFrame:
```
alexfurmenkov, line 29: An interface must not contain any implementation. Please create an abstract class and move _handle_operation_result there. I have also noticed that all operations duplicate  _handle_operation_result call. I think it would be great to call this method only once.

cdisc_rules_engine/utilities/rule_processor.py
nhaydel: The dataset in operation params is the dataset we are performing the operation on. More often than not, the dataset we are performing the operation on is different from the original dataset. So in order to always return a data frame from operation execute methods I opted to append the result to the original dataset

cdisc_rules_engine/operations/operation_interface.py
nhaydel: Would it be better just to change the operation interface class into a BaseOperation class? I don’t see a need for an interface and a base class

cdisc_rules_engine/utilities/rule_processor.py
alexfurmenkov: Got it, then it is a good idea to rename the parameter to be like `original_dataset`

cdisc_rules_engine/operations/operation_interface.py
alexfurmenkov: I think we can. Then this base class can have a public `execute` method which will have 2 steps: execute the operation logic, handle the result. "execute the operation logic" part can be abstract and the concrete operation classes will implement it. This way, `_handle_operation_result` will be called only once.

alexfurmenkov (approved):

cdisc_rules_engine/operations/base_operation.py
```diff
@@ -0,0 +1,51 @@
...
+        pass
+
+    def execute(self) -> pd.DataFrame:
+        result = self._execute_operation()
+        return self._handle_operation_result(result)
+
+    def _handle_operation_result(self, result) -> pd.DataFrame:
+        if self.params.grouping:
+            # Handle grouped results
+            result = result.rename(
```
sityaevDI, line 36: Maybe we can move grouping to separate method to clear handling method

cdisc_rules_engine/operations/valid_meddra_code_references.py
```diff
@@ -0,0 +1,35 @@
+from cdisc_rules_engine.operations.base_operation import BaseOperation
+from uuid import uuid4
+from cdisc_rules_engine.models.dictionaries.meddra.meddra_variables import (
+    MedDRAVariables,
+)
+from cdisc_rules_engine.models.dictionaries.meddra.terms.meddra_term import MedDRATerm
+
+
+class ValidMeddraCodeReferences(BaseOperation):
```
sityaevDI, line 9: I suggest MedDRACodeReferencesValidator name to such classes. For me valid means representation of valid structure, not a validation operation.

cdisc_rules_engine/operations/valid_meddra_code_term_pairs.py
```diff
@@ -0,0 +1,113 @@
...
+    ) -> Optional[Tuple[str, Tuple[str, str]]]:
+        """
+        Extracts target name from params and
+        returns associated term type and dataset columns.
+        """
+        variable_pair_map = {
+            f"{self.params.domain}{MedDRAVariables.SOC.value}": (
+                TermTypes.SOC.value,
+                (
+                    f"{self.params.domain}{MedDRAVariables.SOCCD.value}",
```
sityaevDI, line 45: `{self.params.domain}{MedDRAVariables.SOCCD.value}` can be moved to variable to avoid duplication

sityaevDI (changes requested):

cdisc_rules_engine/operations/valid_meddra_code_term_pairs.py
nhaydel: updated

cdisc_rules_engine/operations/valid_meddra_code_references.py
nhaydel: updated

cdisc_rules_engine/operations/dy.py
```diff
@@ -0,0 +1,16 @@
+from cdisc_rules_engine.operations.base_operation import BaseOperation
+from datetime import datetime
+
+
+class DY(BaseOperation):
```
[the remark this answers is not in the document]
nhaydel: updated

cdisc_rules_engine/operations/base_operation.py
nhaydel: updated

sityaevDI (approved):

PR #102, opened 2022-09-05
Extendable data reader factory

alexfurmenkov:


cdisc_rules_engine/services/data_readers/data_reader_factory.py
```diff
@@ -1,7 +1,39 @@
...
+class DataReaderFactory(FactoryInterface):
+    _service_map = {
+        "xpt": XPTReader,
+    }
+
+    def __init__(self):
+        self._default_service_name: str = "xpt"
 
+    @classmethod
+    def register_service(cls, name: str, service: Type[DataReaderInterface]):
```
nhaydel, line 19: These methods should probably be called register reader and get_reader. Also the _service_map should be reader_map

nhaydel (approved):
Approved with suggestions to rename methods and variable

cdisc_rules_engine/services/data_readers/data_reader_factory.py
alexfurmenkov: We tend to make all factories implement the same interface for convenience. The interface defines the methods as register_service, get_service. Good catch about reader_map, will do.

nhaydel: Sounds good. I will update the operations factory to use get_service, and register_service method

alexfurmenkov: Yeah, just inherit them from interface

PR #104, opened 2022-09-06
Add plugin loader

alexfurmenkov:
This PR adds a point for loading plugins which is automatically invoked during initialization of "engine" folder. 

A plugin is expected to be a Python package which is installed into venv and defines entry points in setup.py (https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-package-metadata):
```
setup(
    ...
    entry_points={'core.plugins.data_readers': 'csv = csv_reader:CSVReader'},
    ...
)
```

where:
`core.plugins.data_readers` a group that plugin is related to, right now I have defined the next groups: data_readers and data_services
`csv` a plugin name which will be used when registering it in the factory
`csv_reader:CSVReader` is a reference to the actual class that implements the plugin logic. This class must implement the interface of the factory objects.

The loader finds all plugins with relevant entrypoint groups and registers each of them in the corresponding factory.


sityaevDI (approved):

nhaydel:
Can we add the OperationsFactory to this before it is merged?

PR #130, opened 2022-09-15
Config and data service improvements

alexfurmenkov:
Recently I have been trying to extend the engine with a plugin that implements AWS data service and figured out that data service interface and base class can be improved in the following ways:

1. Some obligatory to implement methods like `get_variables_metadata` and `get_define_xml_contents` have not been listed in the interface. Added them.
2. Methods `join_split_datasets` and `get_dataset_by_type` do not really depend on the concrete implementation and can be the same for all concrete data services, so I have moved them to the base class.
3. `_async_get_dataset` was refactored to use asyncio instead of multithreading since we try to follow asyncio approach.

Besides, I have noticed a couple of things that can help make the engine configurable from outside:

1. Added `ConfigInterface` to remove the dependency from `ConfigService` class and allow external apps like plugins implement the configuration as they wish.
2. Parametrized `RulesEngine` class to accept config object that implements the interface.

### DIFF
--- a/cdisc_rules_engine/services/data_services/base_data_service.py
+++ b/cdisc_rules_engine/services/data_services/base_data_service.py
@@ -6,6 +6,7 @@ from typing import Callable, List, Optional, Iterable
 import numpy as np
 import pandas as pd
 
+from cdisc_rules_engine.config import ConfigService
 from cdisc_rules_engine.constants.domains import AP_DOMAIN_LENGTH
 from cdisc_rules_engine.interfaces import DataServiceInterface, CacheServiceInterface
 from cdisc_rules_engine.models.dataset_types import DatasetTypes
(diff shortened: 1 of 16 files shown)


cdisc_rules_engine/services/data_services/base_data_service.py
```diff
@@ -138,3 +205,14 @@ def _replace_nans_in_numeric_cols_with_none(dataset: pd.DataFrame):
...
             lambda x: x.replace({np.nan: None})
         )
+
+    async def _async_get_dataset(
+        self, function_to_call: Callable, dataset_name: str, **kwargs
+    ) -> pd.DataFrame:
+        """
+        Asynchronously executes passed function_to_call.
+        """
+        loop = asyncio.get_event_loop()
```
nhaydel, line 215: Is this slower for the local data service?

nhaydel (approved):
Approved. Let's just make sure that local validation still works and is performant if you haven't already

PR #150, opened 2022-10-04
Implement to representation method in rule class

alexfurmenkov:


cdisc_rules_engine/models/rule.py
```diff
@@ -172,26 +168,13 @@ def _to_db_dict(self) -> dict:
...
-            db_dict["datasets"] = self.datasets
+            representation["datasets"] = self.datasets
         if self.output_variables:
-            db_dict["output_variables"] = self.output_variables
+            representation["output_variables"] = self.output_variables
         if self.operations:
-            db_dict["operations"] = self.operations
-        return db_dict
-
-    def _ensure_valid_record_structure(self):
```
nhaydel, line 186: I still think we want some representation validation here.

alexfurmenkov: I think we can follow the concept of serializers as the next iteration of our `RepresentationInterface`.

The idea is simple - we have a separate hierarchy of objects called serializers. Their responsibility is to accept an object of class X as an input argument and implement a unified interface to "display" this object in a certain way(s). Besides representation, they can have a method like `is_valid` that validates the fields of a wrapped object (this is how Django Rest Framework validates incoming requests).

So, with the rule class it can be like:
```
rule_serializer = RuleSerializer(rule_obj)
rule_serializer.data -> {dict representation of the rule}
rule_serializer.is_valid -> bool
```

alexfurmenkov: @nhaydel updated the code in ADO PR to use the serializer, please check it as you have time.

nhaydel: I like this approach. It also allows us to make serializers plugins so instead of having a rule DB model, we can just have a COSMOS serializer. But that can be discussed at a later time.

nhaydel (approved):

PR #160, opened 2022-10-10
Updates to DY operation

nhaydel:
This PR updates the DY operation to apply the following logic:
Get the RFSTDTC values from DM
Get the date portion of whatever DTC variable is given to the operation
Calculate Study Day: --DY = (date portion of --DTC) - (date portion of RFSTDTC) + 1 if --DTC is on or after RFSTDTC, --DY = (date portion of --DTC) - (date portion of RFSTDTC) if --DTC precedes RFSTDTC.

note: This function returns 0 for invalid dates instead of None to support operations comparing an integer to the result of this operation.

### DIFF
--- a/cdisc_rules_engine/operations/day_data_validator.py
+++ b/cdisc_rules_engine/operations/day_data_validator.py
@@ -1,16 +1,40 @@
 from cdisc_rules_engine.operations.base_operation import BaseOperation
 from datetime import datetime
+import numpy as np
 
 
 class DayDataValidator(BaseOperation):
     def _execute_operation(self):
-        dtc_value = self.params.dataframe[self.params.target].map(
-            datetime.fromisoformat
-        )
-        rfstdtc_value = self.params.dataframe["RFSTDTC"].map(datetime.fromisoformat)
+        dtc_value = self.params.dataframe[self.params.target].map(self.parse_timestamp)
 
-        # + 1 if --DTC is on or after RFSTDTC
-        delta = (dtc_value - rfstdtc_value).map(
-            lambda x: x.days if x.days < 0 else x.days + 1
-        )
-        return delta
+        # Always get RFSTDTC column from DM dataset.
+        dm_datasets = [
+            dataset for dataset in self.params.datasets if dataset["domain"] == "DM"
+        ]
+        if not dm_datasets:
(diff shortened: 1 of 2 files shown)


sityaevDI (approved):

cdisc_rules_engine/operations/day_data_validator.py
```diff
@@ -1,16 +1,40 @@
...
+import numpy as np
 
 
 class DayDataValidator(BaseOperation):
     def _execute_operation(self):
-        dtc_value = self.params.dataframe[self.params.target].map(
-            datetime.fromisoformat
-        )
-        rfstdtc_value = self.params.dataframe["RFSTDTC"].map(datetime.fromisoformat)
+        dtc_value = self.params.dataframe[self.params.target].map(self.parse_timestamp)
```
gcampion-cdisc, line 8: Does this require a "domain" parameter be supplied to the operation or will it work with the current domain?

cdisc_rules_engine/operations/day_data_validator.py
```diff
@@ -1,16 +1,40 @@
...
-            lambda x: x.days if x.days < 0 else x.days + 1
-        )
-        return delta
+        # Always get RFSTDTC column from DM dataset.
+        dm_datasets = [
+            dataset for dataset in self.params.datasets if dataset["domain"] == "DM"
+        ]
+        if not dm_datasets:
+            # Return none for all values if dm is not provided.
+            return [0] * len(self.params.dataframe[self.params.target])
```
gcampion-cdisc, line 16: What will happen if DM exists, but the DM.RFSTDTC variable is missing?
What will happen if the --DTC variable is missing in the current domain?

nhaydel: If DM exists but DM.RFSTDTC variable is missing, the rule will be skipped because of a KeyError being thrown
If the --DTC variable is missing, the rule will be skipped for the current domain due to a KeyError being thrown

cdisc_rules_engine/operations/day_data_validator.py
nhaydel: If no domain is supplied it will use the current domain.

gcampion-cdisc (approved):

PR #161, opened 2022-10-11
Serializers for terms

sityaevDI:


alexfurmenkov (approved):

nhaydel:
What is the benefit of serializers in this instance?

PR #170, opened 2022-10-13
Cross Domain Variable Presence

nhaydel:
This PR creates a new operator "variable_count" which counts the number of times a value in the target column of the target dataset appears as a variable in the study. For example if there is a LNKGRP column with values like ["DOMAIN", "AESEQ", ...] The operator will return a series containing the number of times each value appears in the study

### DIFF
--- a/tests/unit/test_operations/test_variable_count.py
+++ b/tests/unit/test_operations/test_variable_count.py
@@ -1,0 +1,46 @@
+from cdisc_rules_engine.config.config import ConfigService
+from cdisc_rules_engine.operations.variable_count import VariableCount
+from cdisc_rules_engine.models.operation_params import OperationParams
+import pandas as pd
+from cdisc_rules_engine.services.cache.cache_service_factory import CacheServiceFactory
+
+
+def test_variable_count(mock_data_service, operation_params: OperationParams):
+    config = ConfigService()
+    cache = CacheServiceFactory(config).get_cache_service()
+    dataset_path = "study/bundle/blah"
+    datasets_map = {
+        "AE": pd.DataFrame.from_dict(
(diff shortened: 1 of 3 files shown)


tests/unit/test_operations/test_variable_count.py
```diff
@@ -0,0 +1,46 @@
...
+    mock_data_service.get_dataset.side_effect = lambda name: datasets_map.get(
+        name.split("/")[-1]
+    )
+    mock_data_service.join_split_datasets.side_effect = lambda func, files: pd.concat(
+        [func(f) for f in files]
+    )
+    operation_params.domain = "RELREC"
+    operation_params.dataframe = datasets_map["RELREC"]
+    operation_params.datasets = datasets
+    operation_params.target = "LNKGRP"
```
gcampion-cdisc, line 40: It looks like the variable_count function is still more complex than what we are expecting to support rule cg0022. From what I can tell from this test, it is looking at the list of variables _within_ LNKGRP and getting a count for each variable.  Instead, we need it only to do a count of the instances of the "LNKGRP" variable itself. Here the result would be "1". If we used "DOMAIN" instead of LNKGRP, the result should be 2. 

Also note, this should work with domain-specific variables like --SEQ. If we ran this test using "--SEQ" instead of LNKGRP, the result should be 2.

gcampion-cdisc (changes requested):

tests/unit/test_operations/test_variable_count.py
nhaydel: Ahh ok, I will update the function to just count variable appearance in a study

nhaydel: updated

gcampion-cdisc (approved):

PR #176, opened 2022-10-19
Add -r argument for selecting individual rules to run

nhaydel:
This PR adds the -r flag for specifying individual rules to run:

`python core.py validate -d XPT/ -s sdtmig -v 3-4 -l info -r CDISC.SDTMIG.CG0344`

The -r flag can be used more than once to specify multiple rules

### DIFF
--- a/scripts/run_validation.py
+++ b/scripts/run_validation.py
@@ -141,21 +141,31 @@ def run_validation(args: Validation_args):
     set_log_level(args.log_level.lower())
     cache_path: str = f"{os.path.dirname(__file__)}/../{args.cache}"
     data_path: str = f"{os.path.dirname(__file__)}/../{args.data}"
-
     # fill cache
     CacheManager.register("RedisCacheService", RedisCacheService)
     CacheManager.register("InMemoryCacheService", InMemoryCacheService)
     manager = CacheManager()
     manager.start()
     shared_cache = get_cache_service(manager)
+    engine_logger.info("Populating cache")
     shared_cache = fill_cache_with_provided_data(shared_cache, cache_path)
 
     # install dictionaries if needed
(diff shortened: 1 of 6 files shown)


scripts/run_validation.py
```diff
@@ -141,21 +141,31 @@ def run_validation(args: Validation_args):
...
+        for rule in args.rules
+    ]
+    if not args.rules:
+        engine_logger.info(
+            f"No rules specified. Running all rules for {args.standard} \
+                version {args.version}"
+        )
+        rules = shared_cache.get_all_by_prefix(
+            get_rules_cache_key(args.standard, args.version.replace(".", "-"))
+        )
```
alexfurmenkov, line 168: I have a couple of ideas here:

First, what do you think about rewriting it to be like:
```
if args.rules:
    rules = [
        shared_cache.get(
            get_rules_cache_key(args.standard, args.version.replace(".", "-"), rule)
        )
        for rule in args.rules
    ]
else:
    engine_logger.info(
        f"No rules specified. Running all rules for {args.standard} \
            version {args.version}"
    )
    rules = shared_cache.get_all_by_prefix(
        get_rules_cache_key(args.standard, args.version.replace(".", "-"))
    )
```

Second, I think we need to display a warning message, not an info on line 162. But it is totally up to you.

Third, on lines 155 - 160, the more rules are given in `args.rules`, the more times we query the cache. Wouldn't it be more performant to refactor it to a single query saying "give me all rules whose IDs are in the list of args.rules"?

nhaydel: I like these suggestions. Will do

nhaydel: updated

alexfurmenkov (approved):

PR #178, opened 2022-10-25
Plugin extendable logging

alexfurmenkov:


cdisc_rules_engine/services/__init__.py
```diff
@@ -1,4 +1,4 @@
 from cdisc_rules_engine.config import config
-from cdisc_rules_engine.services.logging_service_factory import LoggingServiceFactory
+from cdisc_rules_engine.services.logging import LoggingServiceFactory
 
-logger = LoggingServiceFactory.get_logger(config)
+logger = LoggingServiceFactory(config).get_service()
```
nhaydel, line 4: We use this logger in the run_validation.py file. In that file we have the option to disable the logger. https://github.com/cdisc-org/cdisc-rules-engine/blob/development/scripts/run_validation.py#L126 Can we add a method to the interface and logger for disabling logging?

alexfurmenkov: Updated, implemented the same interface as logging library provides:
image

nhaydel (approved):

nhaydel:
Approved. If another release is needed, can you merge to main and release from main?

alexfurmenkov:
> Approved. If another release is needed, can you merge to main and release from main?

Sure, will do.

PR #179, opened 2022-10-25
Extend data service interface

alexfurmenkov:
The main purpose of adding this method is to delete unnecessary JSONBlobDataService class and substitute it with `get_raw_dataset_metadata` method


cdisc_rules_engine/models/dataset_types.py
```diff
@@ -4,5 +4,6 @@
 class DatasetTypes(BaseEnum):
     CONTENTS = "contents"
     METADATA = "metadata"
+    RAW_METADATA = "raw_metadata"
     JSON_METADATA = "json_metadata"
```
nhaydel, line 8: Can we delete this dataset_type as well?

alexfurmenkov: Sure, thanks for noting it

PR #182, opened 2022-10-26
fix deprecated method call

alexfurmenkov:


scripts/run_validation.py
```diff
@@ -137,7 +137,7 @@ def get_rules(cache: CacheServiceInterface, args) -> List[dict]:
         ]
         rules = cache.get_all(keys)
     else:
-        engine_logger.warn(
+        engine_logger.warning(
```
alexfurmenkov, line 140: warn method is deprecated in logging module, so I haven't included it into the interface.

nhaydel (approved):

PR #184, opened 2022-10-27
Fix cache path

alexfurmenkov:
The PR fixes the issue of missing folder with cache files when starting a validation using the CLI.

The operation: `cache_path: str = f"{os.path.dirname(__file__)}/../{args.cache}"` returns a path like `/Users/Aleksei_Furmenkov/Desktop/core/scripts/../resources/cache` when executed in `scripts/run_validation.py`. 

But the issue is that there is no folder `scripts` when building the CLI version (at least on MacOS):
image

As a solution, I suggest concatenating given cache path and current directory in `core.py` file, not in `scripts/run_validation.py`.

@nhaydel can you please run the CLI using `main` branch and check if the issue exists on your machine? Me and Denis have Macs, sorry.


nhaydel:
I downloaded the windows zip file from the last release. I did not have any issue running it

alexfurmenkov:
> @alexfurmenkov if you want to do a mini release (4.9.1?) from this branch I can test it on windows

@nhaydel okay, will do. Good idea

nhaydel:
Sorry for the late response. This new cache path worked for me. Tested on version: 0.4.9.1

nhaydel (approved):

PR #185, opened 2022-10-28
Update whodrug validation logic to match hierarchy validation

nhaydel:
This PR adds logic for validating the whodrug hierarchy similar to how we validate meddra hierarchy. I demo'd this to a whodrug representative at the CDISC US interchange and got approval for the check logic. It will support rules like the following:
```
"conditions": {
      "all": [
        {
          "name": "get_dataset",
          "operator": "not_equal_to",
          "value": {
            "target": "$valid_whodrug_codes",
            "comparator": true
          }
        }
      ]
    },
    "operations": [
      {
        "id": "$valid_whodrug_codes",
        "operator": "whodrug_code_hierarchy"
      }
    ],
```
Additional fields were added to the Whodrug terms to support lookup via an identifier between Drug Names -> ATC Class and ATC Class -> ATC Text

ATC Class is used to determine appropriate Drug Names and ATC Text because there is a 1 to many mapping of ATC Text to ATC Class and a 1 to 1 mapping of Drug Name to ATC Class.


cdisc_rules_engine/models/dictionaries/whodrug/whodrug_terms_factory.py
```diff
@@ -72,4 +74,24 @@ def __create_term_objects_from_file(
             # create a term object for each line and append it to the mapping
             for line in file:
                 term_obj: BaseWhoDrugTerm = model_class.from_txt_line(line)
-                code_to_term_map[term_obj.type].append(term_obj)
+                code_to_term_map[term_obj.type][term_obj.get_identifier()] = term_obj
+
+    @staticmethod
+    def get_code_hierarchies(term_map: dict) -> Set[str]:
```
alexfurmenkov, line 80: Please move this method to operation class

cdisc_rules_engine/models/dictionaries/whodrug/whodrug_terms_factory.py
```diff
@@ -28,7 +29,9 @@ def __init__(self, data_service: DataServiceInterface):
             WhodrugFileNames.INA_FILE_NAME.value: AtcText,
         }
 
-    def install_terms(self, directory_path: str) -> Dict[str, List[BaseWhoDrugTerm]]:
+    def install_terms(
+        self, directory_path: str
+    ) -> Dict[str, Dict[str, BaseWhoDrugTerm]]:
```
alexfurmenkov, line 34: We need to have the same return type in both of the factories: MedDra and WhoDrug. This will allow us depend on the interface, not on the implementation. Having WhoDrug factory interface changed, we also need to refactor MedDra factory. Otherwise, my suggestion would be to rollback WhoDrug factory.

cdisc_rules_engine/models/dictionaries/whodrug/atc_classification.py
```diff
@@ -10,13 +10,23 @@ class AtcClassification(BaseWhoDrugTerm):
...
         self.parentCode: str = record_params["parentCode"]  # Drug Record Number
+        self.checkDigit: str = record_params["checkDigit"]
+        self.parentSequenceNumber: str = record_params["parentSequenceNumber"]
 
     @classmethod
     def from_txt_line(cls, line: str) -> "AtcClassification":
         return cls(
             {
                 "parentCode": line[:6].strip(),
+                "parentSequenceNumber": line[6:7],
```
alexfurmenkov, line 21: We can just get check digit by index, without slicing

cdisc_rules_engine/models/dictionaries/whodrug/whodrug_terms_factory.py
nhaydel: Meddra already returns a dictionary of terms

alexfurmenkov:
Please increment the version, the rest is fine

alexfurmenkov (approved):

PR #188, opened 2022-11-03
Add list rules command

nhaydel:
This PR adds a command to list rules for a proposed desktop application. It is meant to print rules for a given standard and version


core.py
```diff
@@ -191,8 +194,31 @@ def update_cache(ctx: click.Context, cache_path: str, apikey: str):
     cache_populator.save_variables_metadata_locally(cache_path)
 
 
+@click.command()
+@click.option(
+    "-c",
+    "--cache_path",
+    default="resources/cache",
```
alexfurmenkov, line 201: Should we create a constant or enum for default paths?

alexfurmenkov (approved):

PR #202, opened 2022-11-16
Allow absolute paths

nhaydel:
This PR adds the ability to use an absolute path or a relative path to specify data directory. 

Minor tweaks to how we decide to load ct cache data. Needs to be updated further.


alexfurmenkov (approved):
cool

PR #204, opened 2022-11-18
Allow validation of individual datasets, not the whole folder

alexfurmenkov:
This PR allows validating individual datasets, not the whole folder. Since we have switched to supporting absolute paths, the idea is to support a new argument `--dataset-path -dp` that can be multiple. So, the validation command can look like:

`core validate ... -dp=/Users/Aleksei_Furmenkov/PycharmProjects/cdisc-rules-engine/resources/data/ec.xpt
-dp=/Users/Aleksei_Furmenkov/PycharmProjects/cdisc-rules-engine/resources/data/ae.xpt ...`

Original `--data -d` is preserved.

Having this implemented also required changing reports a bit. Previously, we just printed the directory path but we can't do it anymore since individual datasets can be stored in different directories. The following changes have been made:

Excel:
Before
image

After
Screenshot 2022-11-18 at 12 43 05 PM

JSON:
Before
```
{
  "conformance_details": {
    "data_path": "/Users/Aleksei_Furmenkov/PycharmProjects/cdisc-rules-engine/resources/data",
    "report_date": "2022-11-18T12:50:57",
    "runtime": 11.54
  },
  ...
}
```

After
```
{
  "conformance_details": {
    "data_path": [
      "/Users/Aleksei_Furmenkov/PycharmProjects/cdisc-rules-engine/resources/data/aw.xpt",
      "/Users/Aleksei_Furmenkov/PycharmProjects/cdisc-rules-engine/resources/data/ae.xpt",
      "/Users/Aleksei_Furmenkov/PycharmProjects/cdisc-rules-engine/resources/data/ec.xpt"
    ],
    "report_date": "2022-11-18T12:50:57",
    "runtime": 11.54
  },
  ...
}
```


nhaydel:
If a user validate 100 data files by specifying the -d flag, will all 100 appear in that data location field?

alexfurmenkov:
> If a user validate 100 data files by specifying the -d flag, will all 100 appear in that data location field?

Yes, they will. How do you think we should handle it? Maybe should we keep everything in the JSON report and display first N in Excel report with appending something like "..."

nhaydel:
> > If a user validate 100 data files by specifying the -d flag, will all 100 appear in that data location field?
> 
> Yes, they will. How do you think we should handle it? Maybe should we keep everything in the JSON report and display first N in Excel report with appending something like "..."?

I think that is fine for now, we may want a new datasets tab but we can create a new issue for that if people want it

alexfurmenkov:
> > @alexfurmenkov can we at least limit it to 5 files in the excel output for now?
> 
> Yeah, will do

@nhaydel solved merge conflicts and limited excel report to 5 paths. Should we include this PR into 0.5 release?

nhaydel (approved):

PR #205, opened 2022-11-18
Add verbose output flag for printing rule ids as they are completed

nhaydel:
This PR adds a verbose output flag, -vo, to the core validate command that will print rule ids as they are completed.

Additionally, this PR fixes a small bug with the operations tests that was caused by supplying an empty dataframe to the operation constructor.


scripts/run_validation.py
```diff
@@ -61,6 +61,8 @@ def validate_single_rule(cache, path, datasets, args, rule: dict = None):
         for dataset in datasets
     ]
     results = list(itertools.chain(*results))
+    if args.verbose_output:
+        print(f"{rule['core_id']} validation complete")
```
alexfurmenkov, line 65: Shouldn't it be a logger.info call?

alexfurmenkov (approved):

scripts/run_validation.py
nhaydel: I will update. I'll also bump the version to version 0.5 for the release today

PR #211, opened 2022-11-22
Dataset-JSON support (partial)

JozefAerts:
Dataset-JSON support, but still having some trouble with DataReaderFactory, I need to hardcode that the service name is "json" instead of the default service name "xpt".


Test_JSONFiles_MSG/ta.json
```diff
@@ -0,0 +1,12 @@
+{
```
nhaydel, line 1: generally, test resources should go here: https://github.com/cdisc-org/cdisc-rules-engine/tree/development/tests/resources

cdisc_rules_engine/services/data_readers/data_reader_factory.py
```diff
@@ -20,6 +24,7 @@ def register_service(cls, name: str, service: Type[DataReaderInterface]):
         """
         Registers a new service in internal _service_map
         """
+        print("registering data reader service with name = ", name)
```
nhaydel, line 27: no need to register the json data reader because it is already added to the reader_map. Registering services is for external users to extend the tool without having to make a PR

cdisc_rules_engine/services/data_services/local_data_service.py
```diff
@@ -168,29 +177,37 @@ def _async_get_datasets(
...
+        # print("local_data_service.py: reading metadata from file = ", file_path)
         file_size = os.path.getsize(file_path)
         file_metadata = {
             "path": file_path,
             "name": extract_file_name_from_path_string(file_path),
             "size": file_size,
         }
         with open(file_path, "rb") as f:
-            contents_metadata = DatasetMetadataReader(f.read()).read()
+            if not self.use_json:
```
nhaydel, line 188: We should, implement an interface for dataset metadata readers, similar to how we have for data readers. This way we can call something like _metadata_reader_factory.get_service() and return the appropriate metadata reader.

cdisc_rules_engine/services/data_services/local_data_service.py
```diff
@@ -168,29 +177,37 @@ def _async_get_datasets(
...
+                contents_metadata = DatasetMetadataReader(f.read()).read()
+            # 2022-11-18: FOR USE WITH Dataset-JSON only
+            else:
+                contents_metadata = DatasetJSONMetadataReader(file_path).read()
```
nhaydel, line 192: We should be able to detect whether or not it is a json file or not based on the file extension

core.py
```diff
@@ -104,6 +104,17 @@ def cli():
     help="Disable progress bar",
 )
 @click.option("--rules", "-r", multiple=True)
+# Addition Jozef
```
nhaydel, line 107: I think we can avoid this by checking the file extension when reading the file

cdisc_rules_engine/services/data_readers/dataset_json_reader.py
```diff
@@ -0,0 +1,58 @@
...
+)
+
+# Testing only
+# test_file_path = "D:\\CDISC_CORE_Engine\\Python_work_Jozef\\lb.json"
+
+
+class DatasetJSONReader(DataReaderInterface):
+    def from_file(self, file_path):
+        # print("dataset_json_reader: starting reading_from_file = ", file_path)
+        f = open(file_path)
```
nhaydel, line 15: you can read files in python by setting up a context like:
```
with open(file_path) as f:
    data = json.load(f)
```
This ensures that the file is closed once the scope is exited.

cdisc_rules_engine/services/data_readers/dataset_json_reader.py
```diff
@@ -0,0 +1,58 @@
...
+        # returns JSON object as
+        # a dictionary
+        data = json.load(f)
+        # print("data = ", data)
+        # check for "clinicalData"
+        start_data_string = "clinicalData"
+        if 'clinicalData' in data:
```
nhaydel, line 22: is this check necessary if we are not changing start_data_string?

cdisc_rules_engine/services/data_readers/dataset_json_reader.py
```diff
@@ -0,0 +1,58 @@
...
+        # print("Creating DataFrame for dataset with ItemGroupOID = ", item_group_oid)
+        # we need the column names
+        # TODO: need also support for "referenceData" for trial design domains
+        meta_data = data[start_data_string]["itemGroupData"][item_group_oid]["items"]
+        column_names = []
+        for x in meta_data:
+            column_names.append(x["name"])
+        # print("With column names = \n", column_names)
+        # Set up an empty DataFrame, only containing the columns
+        df = pd.DataFrame(columns=column_names)
```
nhaydel, line 41: Is it possible to use pandas from_dict method here to create a df from the dictionary data: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.from_dict.html or a slightly modified version of the dictionary?

nhaydel:
You can change the pr to commit into development. All work is committed to development and then we make a release PR to merge into main when a release is ready

cdisc_rules_engine/services/data_readers/dataset_json_reader.py
JozefAerts: The reason is that in Dataset-XML most datasets have "clinicalData" as the object name, but the trial design datasets (like TS, TA, TI, TE, etc.) have "referenceData" as the object name. We can however not rely on the dataset name for that (there are exceptions to the rule that trial design datasets start with a "T").
It can however be that the are some lines are not necessary (are from early) attempts, so I will check again.

cdisc_rules_engine/services/data_readers/dataset_json_reader.py
JozefAerts: Thanks! This is still a bit a strange construct for me, coming from the Java world ...

JozefAerts:
obsolete, issues fixed and new pull request made

PR #213, opened 2022-11-23
Fixed issue with hardcoded use of json for input files

JozefAerts:


JozefAerts:
Still had a bug when reading metadata for XPT.

PR #219, opened 2022-11-29
Dataset-JSON integration

JozefAerts:
Further testing was performed with some code clean up.


JozefAerts:
It might indeed that there were none against the one from last week, but I thought I had cancelled the one of last week, as it looked it had an issue, but then wasn't one.
Still too unexperienced with all this.
If there is no change, and you already merged last week, we can indeed close this pull request.
I will anyway still do a lot of testing with the merged version.
I am also thinking about a feature to read SDTM in CSV format (used a lot in the academic world, as they don't need to submit to FDA), but it might just be that I don't have the time in the next weeks.

nhaydel:
I have not merged the PR from last week. I have not merged any PR related to DatasetJSON @JozefAerts

nhaydel:
@JozefAerts it looks like you are trying to merge the wrong branch from your fork. Other PRs were from a DatasetJSON branch on your fork

JozefAerts:
Seems to be done incorrectly, starting from development instead of from my own branch ...

PR #222, opened 2022-11-30
Add list-ds-metadata command

alexfurmenkov:
The PR adds a command that accepts a list of absolute dataset paths and returns a list of metadata information for each given dataset.

Input 
`core.py list-dataset-metadata -dp=path_1 -dp=path_2 -dp=path_3 ...`

Output
```
[
   {
      "domain":"AE",
      "filename":"ae.xpt",
      "full_path":"/Users/Aleksei_Furmenkov/PycharmProjects/cdisc-rules-engine/resources/data/ae.xpt",
      "size":"38000",
      "label":"Adverse Events",
      "modification_date":"2020-08-21T09:14:26"
   },
   ...
]
```


core.py
```diff
@@ -1,25 +1,28 @@
 import asyncio
```
alexfurmenkov, line 1: Just sorted imports here since we have a lot of them here

scripts/list_dataset_metadata.py
```diff
@@ -0,0 +1,39 @@
+from typing import Tuple, List
```
alexfurmenkov, line 1: I've created a new file for list_dataset_metadata command because in my opinion `core.py` should only contain a set of entrypoints that just delegate the call to the command controllers.

nhaydel: agreed.

cdisc_rules_engine/services/data_services/local_data_service.py
```diff
@@ -89,6 +89,7 @@ def get_raw_dataset_metadata(self, dataset_name: str, **kwargs) -> DatasetMetada
             label=contents_metadata["dataset_label"],
             modification_date=contents_metadata["dataset_modification_date"],
             filename=file_metadata["name"],
+            full_path=file_metadata["path"],
```
nhaydel, line 92: Why does the metadata path need to be different from the file path?

core.py
```diff
@@ -310,10 +313,49 @@ def list_rule_sets(ctx: click.Context, cache_path: str):
...
 
+@click.command()
+@click.option(
+    "-dp",
+    "--dataset-path",
+    required=True,
+    multiple=True,
+)
+@click.pass_context
+def list_ds_metadata(ctx: click.Context, dataset_path: Tuple[str]):
```
nhaydel, line 324: can we call this function list_dataset_metadata or just list_metadata? I feel like the ds will confuse users

alexfurmenkov: Okay

cdisc_rules_engine/services/data_services/local_data_service.py
alexfurmenkov: Metadata path is the same as the file metadata, maybe I misunderstood your question?

core.py
alexfurmenkov: @nhaydel renamed

cdisc_rules_engine/services/data_services/local_data_service.py
nhaydel: My question is why do we need a filename variable and a full_path variable? Shouldn't they be the same, or at least shouldn't we be able to get the filename from the full path

PR #225, opened 2022-12-06
Parametrize progress bar

alexfurmenkov:
While working on a custom UI for the CORE engine we have identified the need to display the progress percentage as we had in the web version. 

Despite the fact that the CLI prints the progress bar to stdout by default, it is not very convenient to parse this progress bar on the desktop frontend and extract the percentage from it. So, in this PR I suggest parametrizing the progress bar in the following way - introduce a new parameter `-p, --progress` that can accept any of the following values:

- bar (default) - prints a progress bar like [████████████████████████████--------] 78%
- percents - prints only percents on new line like:
```
23
24
34
...
```
- disabled - disables the progress output (similarly to what `--disable-progressbar` flag does)

To preserve current behavior of the CLI, by default a progress bar is printed and instead of `--disable-progressbar` parameter we can pass `-p disabled`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,6 @@ From the root of the project run the following command:
 Clone the repository and run `core.py --help` to see the full list of commands.
 
 Run `core.py validate --help` to see the list of validation options.
-
 ```
   -ca, --cache TEXT               Relative path to cache files containing pre
                                   loaded metadata and rules
@@ -57,8 +56,12 @@ Run `core.py validate --help` to see the list of validation options.
   -r, --rules TEXT
   -vo, --verbose-output           Specify this option to print rules as they
                                   are completed
+  -p, --progress [bar|disabled|percents]
+                                  Defines how to display the validation
+                                  progress. By default a progress bar like
+                                  "[████████████████████████████--------]
+                                  78%"is printed.
   --help                          Show this message and exit.
-
 ```
 
 #### Validate folder
--- a/README.md
+++ b/README.md
@@ -52,7 +52,6 @@ Run `core.py validate --help` to see the list of validation options.
                                   files
   --meddra TEXT                   Path to directory with MedDRA dictionary
                                   files
-  --disable-progressbar           Disable progress bar
   -r, --rules TEXT
   -vo, --verbose-output           Specify this option to print rules as they
                                   are completed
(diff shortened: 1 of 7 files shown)


nhaydel:
Doesn't the -vo flag already do this?

alexfurmenkov:
> Doesn't the -vo flag already do this?

Can you please share the output of the command using this flag?

alexfurmenkov:
> @alexfurmenkov your assessment is fair, I would like to add verbose_output as an option for the --progress flag.

Sure, will do. Can you please clarify what is expected from this option?

nhaydel:
> > @alexfurmenkov your assessment is fair, I would like to add verbose_output as an option for the --progress flag.
> 
> Sure, will do. Can you please clarify what is expected from this option?

If provided it should log:
`[Level 51 2022-12-06 08:10:55,627 - console_logger.py:54] - CDISC.SDTMIG.CG0177 validation complete` as each rule is completed. It should log even if logging is disabled

alexfurmenkov:
> > > @alexfurmenkov your assessment is fair, I would like to add verbose_output as an option for the --progress flag.
> > 
> > 
> > Sure, will do. Can you please clarify what is expected from this option?
> 
> If provided it should log: `[Level 51 2022-12-06 08:10:55,627 - console_logger.py:54] - CDISC.SDTMIG.CG0177 validation complete` as each rule is completed. It should log even if logging is disabled

So, you'd like to replace existing flag with this parameter, right?

alexfurmenkov:
> > > > > @alexfurmenkov your assessment is fair, I would like to add verbose_output as an option for the --progress flag.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > Sure, will do. Can you please clarify what is expected from this option?
> > 
> > 
> > > 
> > 
> > 
> > > If provided it should log: `[Level 51 2022-12-06 08:10:55,627 - console_logger.py:54] - CDISC.SDTMIG.CG0177 validation complete` as each rule is completed. It should log even if logging is disabled
> > 
> > 
> > So, you'd like to replace existing flag with this parameter, right?
> 
> Yes, like you did for percentage and progress bar, there should be an option for verbose that prints log statements for each rule completed

Added

nhaydel (approved):

Issue #221, opened 2022-11-29
Rule blocked: CORERULES-15 CORERULES-16


gcampion-cdisc:
This bug was created in an ADO PBI but was not carried to Github until now

**Sample Request**
[cg0037.json.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/10115132/cg0037.json.txt)

**Links to related JIRA Tickets**
- https://dev.azure.com/cdisc-org/cdisc-core/_workitems/edit/4708
- https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-15
- https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-16

**Rule Information**
- **Standard**: SDTMIG
- **Rule ID**: CG0037, CG0039
- **Rule Description**: --SOCCD != --BDSYCD

**Describe the bug**
Rule Sample:
```yaml
- name: --SOCCD
  operator: not_equal_to
  value: --BDSYCD
```
The response is unexpected because we are comparing null not_equal_to null, which should return false instead of true.

**Error returned from Rule Engine**
```json
{
  "MH": [
    {
      "domain": "MH",
      "errors": [
        {
          "row": 4,
          "seq": 1,
          "uSubjId": "CDISC006",
          "value": {
            "MHBDSYCD": null,
            "MHSOCCD": null,
            "MHTERM": "ALZHEIMER'S DISEASE"
          }
        }
      ],
      "executionStatus": "success",
      "message": "Primary analysis was used but MHSOCCD and MHBDSYCD are not equal",
      "variables": [
        "MHBDSYCD",
        "MHSOCCD",
        "MHTERM"
      ]
    }
  ]
}
```

**Expected behavior**
- When doing "equal_to" and "not_equal_to" and other comparison type operators, empty "" string values should be handled the same way null numeric values are handled.
- The attached request should return no results in the "errors" array.
- The comparisons should return false for both "equal_to" and "not_equal_to" whenever both comparators are empty "" or null.
- The comparisons should follow this truth table:
https://cdisc-org.github.io/conformance-rules-editor/#/boolean_logic?id=equal_to-not_equal_to

**Notes**
I believe [#3968](https://dev.azure.com/cdisc-org/2c53d232-12ce-401e-9084-84689e2d7372/_workitems/edit/3968)  partially resolved the issue, but maybe the issue was only resolved for the case of empty "" string values. In this rule, we have null numeric values instead.

nhaydel:
[...]

PR #241, opened 2023-01-04
Update DY calculation to merge dm with evaluation dataset and calcula…

nhaydel:
…te day difference

This PR updates the day-difference calculation, to properly calculate the day difference.

Steps for calculation:

1. Merge DM dataset with the evaluation dataset on usubjid
2. Calculate the day difference between the associated RFSTDTC from the DM domain and the target value in the the evaluation dataset.

Note: This operation now assumes that both the evaluation dataset and the dm dataset have a USUBJID column. If this is not present the operation will raise an error and the rule using the operation will be skipped

### DIFF
--- a/cdisc_rules_engine/operations/day_data_validator.py
+++ b/cdisc_rules_engine/operations/day_data_validator.py
@@ -5,7 +5,10 @@ import numpy as np
 
 class DayDataValidator(BaseOperation):
     def _execute_operation(self):
-        dtc_value = self.params.dataframe[self.params.target].map(self.parse_timestamp)
+
+        dtc_value = self.evaluation_dataset[self.params.target].map(
+            self.parse_timestamp
+        )
 
         # Always get RFSTDTC column from DM dataset.
         dm_datasets = [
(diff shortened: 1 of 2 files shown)


gcampion-cdisc (commented):
I tried it with this check:
```json
  "Check": {
    "all": [
      {
        "name": "--DY",
        "operator": "not_equal_to",
        "value": "$val_dy"
      }
    ]
  },
```

And this data:
```json
        "EXDTC": [
          "2012-11-29",
          "2012-11-30",
          "2012-12-01",
          "2012-12",
          "2012-12",
          "2012-12-04T01",
          "2012-12-05"
        ],
        "EXDY": [-1, 1, 2, 0, null, 7, 6],
```
It gives an error for records # 4 (0), # 5 (null), # 6 (7)
I was not expecting it give an error for # 5 (null).

This is my reasoning:
1. The val_dy should be calculated as null/None
2. null not_equal_to null should evaluate to false/pass:
https://cdisc-org.github.io/conformance-rules-editor/#/boolean_logic?id=equal_to-not_equal_to

I am guessing my reasoning is failing at step 1 because the val_dy calculation is failing altogether and that triggers the error without going to step 2?  If that is the case, this might be okay. I don't know whether it is better to error out (and leave it to the author to explicitly include the additional check logic) or simply return null.
If it is failing at step 2, then this is a problem.

Otherwise, things looks good.

nhaydel:
> I tried it with this check:
> 
> ```json
> 
>   "Check": {
> 
>     "all": [
> 
>       {
> 
>         "name": "--DY",
> 
>         "operator": "not_equal_to",
> 
>         "value": "$val_dy"
> 
>       }
> 
>     ]
> 
>   },
> 
> ```
> 
> 
> 
> And this data:
> 
> ```json
> 
>         "EXDTC": [
> 
>           "2012-11-29",
> 
>           "2012-11-30",
> 
>           "2012-12-01",
> 
>           "2012-12",
> 
>           "2012-12",
> 
>           "2012-12-04T01",
> 
>           "2012-12-05"
> 
>         ],
> 
>         "EXDY": [-1, 1, 2, 0, null, 7, 6],
> 
> ```
> 
> It gives an error for records # 4 (0), # 5 (null), # 6 (7)
> 
> I was not expecting it give an error for # 5 (null).
> 
> 
> 
> This is my reasoning:
> 
> 1. The val_dy should be calculated as null/None
> 
> 2. null not_equal_to null should evaluate to false/pass:
> 
> https://cdisc-org.github.io/conformance-rules-editor/#/boolean_logic?id=equal_to-not_equal_to
> 
> 
> 
> I am guessing my reasoning is failing at step 1 because the val_dy calculation is failing altogether and that triggers the error without going to step 2?  If that is the case, this might be okay. I don't know whether it is better to error out (and leave it to the author to explicitly include the additional check logic) or simply return null.
> 
> If it is failing at step 2, then this is a problem.
> 
> 
> 
> Otherwise, things looks good.

When given a null value the dy calculation returns the empty string and null != "" is the reason you are getting an error

gcampion-cdisc:
Ah, I didn't realize this issue is still open here:
https://github.com/cdisc-org/cdisc-rules-engine/issues/221

Since that's the case, this PR is good.

gcampion-cdisc (approved):

nhaydel:
> Ah, I didn't realize this issue is still open here:
> 
> https://github.com/cdisc-org/cdisc-rules-engine/issues/221
> 
> 
> 
> Since that's the case, this PR is good.

A PR has been open for that issue for a while: https://github.com/cdisc-org/business-rules/pull/57

PR #268, opened 2023-02-07
Get dataset variables as well as class variables

nhaydel:
This PR updates the library column order operation to generate the list using the  datasetVariables for a domain. If the dataset is a general observations class, the Identifier and Timing variables are added.

This PR also adds a new operation "get_model_column_order", which gets the dataset variables for a domain from the model, and appends Identifier and Timing variables as necessary. Additionally, if the dataset is apart of the findings about class, the findings variables and findings about variables are also added to the list of allowed variables

### DIFF
--- a/cdisc_rules_engine/operations/library_model_column_order.py
+++ b/cdisc_rules_engine/operations/library_model_column_order.py
@@ -77,11 +77,10 @@ class LibraryModelColumnOrder(BaseOperation):
         domain_details = self._get_model_domain_metadata(
             model_details, self.params.domain
         )
-        class_title = domain_details["_links"]["parentClass"]["title"]
-        class_details = self._get_model_class_metadata(model_details, class_title)
+        class_name = self._get_class_name_from_ig(standard_details)
+        class_details = self._get_model_class_metadata(model_details, class_name)
         variables_metadata = domain_details.get("datasetVariables", [])
         variables_metadata.sort(key=lambda item: item["ordinal"])
-        class_name = class_details.get("name")
         if class_name in DETECTABLE_CLASSES or class_name == FINDINGS_ABOUT:
             # if the class is one of Interventions, Findings, Events, or Findings About
             # -> get class variables instead of datasetVariables add
@@ -110,7 +109,7 @@ class LibraryModelColumnOrder(BaseOperation):
             gen_obs_class_metadata: dict = self._get_model_class_metadata(
                 model_details, GENERAL_OBSERVATIONS_CLASS
             )
-            identifiers_metadata, timing_metadata = self._sort_class_variables_by_role(
+            identifiers_metadata, timing_metadata = self._group_class_variables_by_role(
                 gen_obs_class_metadata["classVariables"]
             )
             # Identifiers are added to the beginning and Timing to the end
@@ -134,10 +133,21 @@ class LibraryModelColumnOrder(BaseOperation):
             model_version = ""
         return model_type, model_version
 
+    def _get_class_name_from_ig(self, standard_details) -> Tuple:
+        # Get domain and class details for domain. This logic is specific
+        # to SDTM based standards. Needs to be expanded for other models
+        for c in standard_details.get("classes"):
+            domain_details = search_in_list_of_dicts(
+                c.get("datasets", []), lambda item: item["name"] == self.params.domain
+            )
+            if domain_details:
+                return c.get("name")
+        return None
+
     def _get_model_class_metadata(
         self,
         model_details: dict,
-        class_title: str,
+        class_name: str,
     ) -> dict:
         """
         Extracts metadata of a certain class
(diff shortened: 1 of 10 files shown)


nhaydel:
>  Which issue(s) are linked to this PR? I'm not sure what is meant by classVariables vs datasetVariables.
> 
> 
> 
> A Special Purpose domain like DM that is defined in the model should only look at dataset variables for that specific domain.
> 
> A Findings domain like LB should look at the Findings class variables for Topic and Qualifier variables and the General Observations class variables for Identifier and Timing variables.
> 
> 
> 
> I'm not sure where a dataset would need to look at both classVariables and datasetVariables.

This is not linked to a specific issue, Jozef just brought this up when testing a rule. Since the variables allowed in a domain are the union of the variables for that class and the variables in the dataset we get both sets. I believe in most cases they will be the same set. In the case of general observations classes the timing and identifier variables are also added

JozefAerts:
The rules I was testing was SEND-266 (see https://jira.cdisc.org/browse/CORERULES-702) "Additional SDTM 1.5 Timing variables may be used in any domains based on the general observation class except when restricted either by domain specific assumptions or listed in the SENDIG as prohibited for use in SEND" (remark that "except when" can currently not be implemented) which is very similar to CG0013 "Variable = Model List of Allowed Variables for Observation Class", coming from Section 2.5 in the SDTMIG-v3.4 "sponsors may not add any ..." (see attached screenshot).
Unbenannt

nhaydel:
> 2\. descendant subclasses

@gcampion-cdisc what is meant by "match on"

i will update the logic as follows:
If a dataset is a non general observations class - use dataset variables
if it is a general observation class - use Identifiers + class variables + Timing.

Shouldn't the class variables contain Topic and qualifier variables already?

cdisc_rules_engine/operations/library_model_column_order.py
```diff
@@ -0,0 +1,190 @@
...
+        variables_metadata.sort(key=lambda item: item["ordinal"])
+        class_name = class_details.get("name")
+        if class_name in DETECTABLE_CLASSES or class_name == FINDINGS_ABOUT:
+            # if the class is one of Interventions, Findings, Events, or Findings About
+            # -> get class variables instead of datasetVariables add
+            # General Observation class variables to variables metadata
+            variables_metadata = class_details.get("classVariables", [])
+            variables_metadata.sort(key=lambda item: item["ordinal"])
+
+            if class_name == FINDINGS_ABOUT:
```
gcampion-cdisc, line 92: Is FINDINGS_ABOUT actually being detected anywhere? I don't see it in the BaseDataService.get_dataset_class.
If it needs to be added, we need to be careful that "Findings about" domains are still included when the scope specifies "findings"
Does this need to be a new ticket?

cdisc_rules_engine/operations/library_model_column_order.py
```diff
@@ -0,0 +1,190 @@
...
+            raise ValueError(
+                f"Model domain metadata is not found in CDISC Library. "
+                f"standard={self.params.standard}, "
+                f"version={self.params.standard_version}, "
+                f"domain={domain_name}"
+            )
+
+        return domain_details
+
+    def _sort_class_variables_by_role(
```
gcampion-cdisc, line 175: I would rename this to "_group_class_variables_by_role"

gcampion-cdisc (changes requested):

cdisc_rules_engine/operations/library_model_column_order.py
nhaydel: In this operation the class is determined via library metadata. So if the domain appears in the findings about class it will be considered a findings about domain

gcampion-cdisc: In the model, General Observations classes (Interventions, Events, Findings, and Findings About) don't have domains like they do in the IG.
Special Purpose, Trial Design, Study Reference, and Relationship classes have domains.

https://dev.cdisclibrary.org/browser/#/mdr/sdtm/2-0/classes/FindingsAbout

nhaydel: @gcampion-cdisc if the model doesn't have the domain, then the class name would have to come from the ig correct? We can use the ig to get the class for the domain and then get the class variables from the model

nhaydel: Updated to this logic:
1. Get domains parent class from IG
2. Get model class variables using the class found in the IG

if this logic is not sufficient then the engine will have to look at the actual data to determine the class. If that is the case we need an issue with clear criteria on how to determine if a dataset belongs to the FINDINGS_ABOUT class

cdisc_rules_engine/operations/library_model_column_order.py
nhaydel: updated.

cdisc_rules_engine/operations/library_model_column_order.py
gcampion-cdisc: It shouldn't use the ig. If the domain name isn't found in one of the other model domains, it should use the topic variable the same way as the get_dataset_class method. That method should also be updated so that if a Findings class has an "--OBJ" variable, it should be considered "Findings About".

I would also update the `_contains_topic_variable` method to not use `ends_with`, but use 2-char wildcard matching instead. ABTRT is a topic variable, but ABCTRT is NOT a topic variable.

nhaydel: Updated. Also update the _contains_topic_variable method to search the dataset for domain_name+topic_variable

gcampion-cdisc (approved):

PR #321, opened 2023-03-16
Iss 315 move cdisc library conformance rules generator to GitHub

gerrycampion:
Moves ADO cdisc-library-conformance-rules-generator repo to cdisc-rules-engine in GH.

Made the following transformations to the generator:
- Deleted unused endpoint test files to fix broken unit tests
- Moved function app files to root
- Converted pipeline to workflow
- Merged unit tests
- Imported relative cdisc-rules-engine instead of using requirements package
- Linted
- updated hosts extension bundle version
- remove enums and use enums from cdisc-rules-engine

Made following changes to cdisc-rules-engine:
- Updated automated-CI linter versions to match pre-commit linter versions
- Created secret AZURE_SP (service principal)
- Merged gitignores
- Merged requirements

To test you can double check:
- Workflow has been tested and run successfully. View the `Deploy Azure Rule Tester Endpoint` Actions. To rerun yourself:
  `gh workflow run "Deploy Azure Rule Tester Endpoint" --ref ISS-315-move-cdisc-library-conformance-rules-generator-to-github`
- Confirm that the function app code has updated
- Confirm that rule-editor can still access the execute rule endpoint successfully


cdisc_rule_tester/models/package.py
```diff
@@ -0,0 +1,30 @@
+import json
```
nhaydel, line 1: I don't think we need this file anymore

cdisc_rule_tester/rule_builder.py
```diff
@@ -0,0 +1,80 @@
+from cdisc_library_client import CDISCLibraryClient
```
nhaydel, line 1: This is also no longer needed

requirements.txt
```diff
@@ -1,20 +1,24 @@
-pytest==7.1.2
-pytest-cov==3.0.0
-pandas==1.3.5
+azure-functions
```
nhaydel, line 1: The requirements.txt for the test function should be separate from the engine

cdisc_rule_tester/models/rule.py
```diff
@@ -0,0 +1,223 @@
+import yaml
```
nhaydel, line 1: I don't think this is necessary anymore either because we are using the model from the engine

cdisc_rule_tester/services/config_service.py
```diff
@@ -0,0 +1,40 @@
+import os
```
nhaydel, line 1: This is no longer needed as well

nhaydel (changes requested):
Some files to remove.
Also we should separate the requirements.txt for each app
Do you plan on removing the engine from the requirements and using the repository code for the engine or will that be a separate issue?

gerrycampion:
> Do you plan on removing the engine from the requirements and using the repository code for the engine or will that be a separate issue?

I had already removed the engine from the requirements to use the repo code instead. Or am I missing something?

cdisc_rule_tester/services/config_service.py
gerrycampion: is logging_service still needed?

nhaydel: You're right, I don't think we need logging_service as well

nhaydel (approved):

PR #337, opened 2023-03-24
Parent lib model col order Operation added

gerrycampion:
New operation. Given a supp dataset, the operation will find the parent domain, find the class of the parent domain, and return the possible model variables (in order) for the parent domain.
This will allow us to write a rule to check whether a supp qnam value is in the list of parent model variables.

To test, verify and run the unit tests.

### DIFF
--- a/cdisc_rules_engine/operations/library_model_column_order.py
+++ b/cdisc_rules_engine/operations/library_model_column_order.py
@@ -26,17 +26,16 @@ class LibraryModelColumnOrder(BaseOperation):
         The lists with column names are sorted
         in accordance to "ordinal" key of library metadata.
         """
+        return self._get_variable_names_list(self.params.domain, self.params.dataframe)
 
+    def _get_variable_names_list(self, domain, dataframe):
         # get variables metadata from the standard model
         variables_metadata: List[
             dict
-        ] = self._get_variables_metadata_from_standard_model(
-            self.params.domain, self.params.dataframe
-        )
-
+        ] = self._get_variables_metadata_from_standard_model(domain, dataframe)
         # create a list of variable names in accordance to the "ordinal" key
         variable_names_list = self._replace_variable_wildcards(
-            variables_metadata, self.params.domain
+            variables_metadata, domain
         )
         return list(OrderedDict.fromkeys(variable_names_list))
 
--- a/cdisc_rules_engine/operations/library_model_column_order.py
+++ b/cdisc_rules_engine/operations/library_model_column_order.py
@@ -30,15 +30,19 @@ class LibraryModelColumnOrder(BaseOperation):
         # get variables metadata from the standard model
         variables_metadata: List[
             dict
-        ] = self._get_variables_metadata_from_standard_model()
+        ] = self._get_variables_metadata_from_standard_model(
+            self.params.domain, self.params.dataframe
+        )
 
         # create a list of variable names in accordance to the "ordinal" key
-        variable_names_list = [
-            var["name"].replace("--", self.params.domain) for var in variables_metadata
-        ]
+        variable_names_list = self._replace_variable_wildcards(
+            variables_metadata, self.params.domain
+        )
         return list(OrderedDict.fromkeys(variable_names_list))
 
-    def _get_variables_metadata_from_standard_model(self) -> List[dict]:
+    def _get_variables_metadata_from_standard_model(
+        self, domain, dataframe
+    ) -> List[dict]:
         """
         Gets variables metadata for the given class and domain from cache.
         The cache stores CDISC Library metadata.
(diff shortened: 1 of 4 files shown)


cdisc_rules_engine/operations/library_model_column_order.py
```diff
@@ -30,15 +30,19 @@ def _execute_operation(self):
         # get variables metadata from the standard model
         variables_metadata: List[
             dict
-        ] = self._get_variables_metadata_from_standard_model()
+        ] = self._get_variables_metadata_from_standard_model(
+            self.params.domain, self.params.dataframe
```
nhaydel, line 34: Why are you passing instance variables into the method?

cdisc_rules_engine/operations/parent_library_model_column_order.py
```diff
@@ -0,0 +1,52 @@
...
+        Length of Series is equal to the length of given dataframe.
+        The lists with column names are sorted
+        in accordance to "ordinal" key of library metadata.
+        """
+        if "RDOMAIN" not in self.params.dataframe:
+            return []
+        rdomains = self.params.dataframe["RDOMAIN"].unique()
+        rdomain_names_list = {}
+        for rdomain in rdomains:
+            parent_datasets = [
```
nhaydel, line 28: Instead of having to loop over the datasets for each rdomain, can't you get the list of datasets corresponding to an rdomain by looping over the datasets and generating a map of rdomain -> parent datasets. Then for each value in the map you can get the appropriate metadata?

nhaydel (changes requested):
One comment on performance of determining the parent_datasets associated with each rdomain

cdisc_rules_engine/operations/parent_library_model_column_order.py
gerrycampion: I hadn't thought about this. Since this operation is primarily for supps, which should always have the same rdomain throughout, should I just take the first RDOMAIN value instead of generating variable mappings for all parent domains?

nhaydel: If you think this is an appropriate assumption to make, but I don't think we need to make that assumption

gerrycampion: I switched to setdefault to avoid the extra loop through the records and to avoid getting variable lists for unused domains

gerrycampion: Ah, I thought your concern was with looping through the dataset rows an extra time here: `self.params.dataframe["RDOMAIN"].unique()`

I don't think we want to reduce the loops through the dataset list at the cost of doing unnecessary metadata lookups. Typically this Operation would involve:
`1 single rdomain` * `a study of ~25 datasets.`
That's small. I could create a dict of domains -> datasets, which would be useful to have globally across the entire validation, but the gains here would still be marginal

nhaydel: My concern is this loop here: https://github.com/cdisc-org/cdisc-rules-engine/pull/337/files#diff-91c2b79e0fca32264501ab5b7b25327d9c1a67f387c70b3127c95db0703502daR32 If we call it for each RDOMAIN we'd have to loop through datasets per rdomain. I guess we can assume there will usually only be 1 rdomain, but we can avoid running that loop all together by just building a map domains -> datasets. While it would be a slight gain overall, I think that map would be particularly useful here to avoid relying on the assumption that there is only 1 single rdomain in the dataset

cdisc_rules_engine/operations/parent_library_model_column_order.py
```diff
@@ -0,0 +1,43 @@
...
+        2    ["STUDYID", "DOMAIN", ...]
+        ...
+
+        Length of Series is equal to the length of given dataframe.
+        The lists with column names are sorted
+        in accordance to "ordinal" key of library metadata.
+        """
+        domain_to_datasets = self._get_domain_to_datasets()
+        rdomain_names_list = {}
+        return Series(
```
nhaydel, line 24: Sorry I didn't mention this before, but what happens if the series returned is shorter than the dataset being evaluated?

gerrycampion: I don't think this can happen unless RDOMAIN column is missing, so I added logic to handle that

cdisc_rules_engine/operations/parent_library_model_column_order.py
```diff
@@ -26,7 +26,9 @@ def _execute_operation(self):
                 rdomain,
                 self._get_parent_variable_names_list(domain_to_datasets, rdomain),
             )
-            for rdomain in self.params.dataframe.get("RDOMAIN", [])
+            for rdomain in self.params.dataframe.get(
+                "RDOMAIN", [None] * self.params.dataframe.shape[0]
```
nhaydel, line 30: NIT: This is perferably `len(self.params.dataframe)`

PR #343, opened 2023-03-28
Capitalize USUBJID and SEQ in results, update tests

nhaydel:
Steps to test:

1. Run unit tests
2. Run a validation and ensure usubjid and seq are populated in the report


gerrycampion (approved):
There is also a comment in models.actions.py that has the old uSubjId and seq casing

PR #350, opened 2023-03-29
Uses new report template and fixes missing details

gerrycampion:
To test:
Run a new validation and ensure the new report contains appropriate values in all columns. Likely FDA Rule Id and PMDA Rule Id columns will be empty unless this is run with FDA or PMDA rules.


cdisc_rules_engine/models/rule_validation_result.py
```diff
@@ -2,16 +2,36 @@
...
         self.executability: str = rule.get("executability")
-        self.message: str = rule.get("message")
+        actions = rule.get("actions")
+        self.message: str = None
+        if actions and len(actions) == 1:
+            self.message = actions[0].get("params", {}).get("message")
         self.execution_status: str = get_execution_status(results)
         self.results = results
 
+    def _get_rule_ids(self, rule: Rule, org: str):
```
nhaydel, line 22: Can you add a return type here?

cdisc_rules_engine/services/reporting/base_report.py
```diff
@@ -37,9 +37,8 @@ def get_summary_data(self) -> List[List]:
         information:
         return [
             "Dataset",
-            "RuleID",
+            "CORE-ID",
             "Message",
-            "Executability",
```
nhaydel, line 42: Are we supposed to be removing executability? I don't remember seeing it in the acceptance criteria

gerrycampion: Question for @nickdedonder
Your provided template has `Executability` removed from the issue summary. It still exists in the Issue details. Is this intentional?

gerrycampion: @nickdedonder 
Sorry, to clarify, this was mislabeled as `Severity` in the previous version of the report template. So you actually removed the `Severity` column.

nickdedonder: yes, I deleted that as there was a decision not to use severity. We'll check in the future what do with it as this is only applicable for the PMDA. I don't want to have severity always present if I'm not running rules for the PMDA

nhaydel (approved):
LGTM

PR #356, opened 2023-04-03
Endpoint supports define.xml

gerrycampion:
To test, run the new test case.

You can also run this branch of the rule editor https://github.com/cdisc-org/conformance-rules-editor/tree/133-support-definexml-test-files locally against this branch and test submitting a define.xml file.

### DIFF
--- a/cdisc_rules_engine/services/data_services/dummy_data_service.py
+++ b/cdisc_rules_engine/services/data_services/dummy_data_service.py
@@ -121,7 +121,7 @@ class DummyDataService(BaseDataService):
             dataset_name=dataset_name, **params
         )
 
-    def get_define_xml_contents(self, dataset_name: str) -> bytes:
+    def get_define_xml_contents(self, dataset_name: str) -> str:
         return self.define_xml
 
     def has_all_files(self, prefix: str, file_names: List[str]) -> bool:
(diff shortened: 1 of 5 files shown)


cdisc_rules_engine/services/data_services/dummy_data_service.py
```diff
@@ -121,7 +122,7 @@ def get_dataset_by_type(
         )
 
     def get_define_xml_contents(self, dataset_name: str) -> bytes:
-        pass
+        return self.define_xml
```
nhaydel, line 125: Should this be converted to bytes before returning? I think by default it is a string

gerrycampion: Would the update I made be better since from_file_contents takes Union[bytes, str] ?

nhaydel: I think it would be better just to convert the string to bytes to conform with what local data service does at the moment. This way there is less room for the test endpoint to have failures that aren't caught by the local data service

nhaydel (changes requested):
I prefer only having one return type. It is cleaner and not really any more work to convert the string in dummy data service to bytes

nhaydel (approved):

PR #385, opened 2023-04-14
Only get rules if rule exists

nhaydel:
Steps to test:

Attempt to run a validation with a rule id that does not exist or is not published. Verify that the validation is still successful


gerrycampion:
If rules A,B,C are requested and only A & B exist, what will happen to rule C in the report?

nhaydel:
Rule C will not be selected for validation so it will not be run and not appear in the report

PR #386, opened 2023-04-17
Add Define Item Group Metadata Check rule type and test

nhaydel:
This PR adds a new rule type for getting the define item group metadata for a domain, and updates the variable exists operator logic to check the dataset contents rather than assuming the contents are in the current dataset.

Steps to test:

* Run the test command using the attached rule, dataset, and define.xml. Ensure the define.xml file is in the same directory as the dataset. (Note: Be sure to change the define file to an xml file and the define_rule and dataset file to .json files)
* Verify VS is the only reported erroneous dataset.
[define.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/11254813/define.txt)
[define_rule.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/11254814/define_rule.txt)
[dataset.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/11254816/dataset.txt)

### DIFF
--- a/cdisc_rules_engine/dataset_builders/base_dataset_builder.py
+++ b/cdisc_rules_engine/dataset_builders/base_dataset_builder.py
@@ -60,6 +60,28 @@ class BaseDatasetBuilder:
             for dataset in get_corresponding_datasets(self.datasets, self.domain)
         ]
 
+    def get_define_xml_item_group_metadata(self, domain: str) -> List[dict]:
+        """
+        Gets Define XML item group metadata
+        returns a list of dictionaries containing the following keys:
+            "define_dataset_name"
+            "define_dataset_label"
+            "define_dataset_location"
+            "define_dataset_class"
+            "define_dataset_structure"
+            "define_dataset_is_non_standard"
+            "define_dataset_variables"
+        """
+        directory_path = get_directory_path(self.dataset_path)
+        define_xml_path: str = f"{directory_path}/{DEFINE_XML_FILE_NAME}"
+        define_xml_contents: bytes = self.data_service.get_define_xml_contents(
+            dataset_name=define_xml_path
+        )
+        define_xml_reader = DefineXMLReader.from_file_contents(
+            define_xml_contents, cache_service_obj=self.cache
+        )
+        return define_xml_reader.extract_domain_metadata(domain)
+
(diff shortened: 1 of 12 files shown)


cdisc_rules_engine/utilities/decorators.py
```diff
@@ -37,19 +38,21 @@ def inner(*args, **kwargs):
     return decorator
 
 
-def cached(cache_key: str):
+def cached(cache_key: str):  # noqa: C901
```
gerrycampion, line 41: what does this mean

cdisc_rules_engine/operations/variable_exists.py
```diff
@@ -4,4 +4,5 @@
 class VariableExists(BaseOperation):
     def _execute_operation(self):
         # get metadata
-        return self.params.target in self.params.dataframe
+        dataframe = self.data_service.get_dataset(self.params.domain)
```
gerrycampion, line 7: Negative case seems to work correctly, but I can't get the positive case to work by renaming VSNOTEST to VSTESTCD in dataset.json. It still says $testcd_exists=False

The other parts of the rule appear to work.

gerrycampion (changes requested):

cdisc_rules_engine/dataset_builders/base_dataset_builder.py
```diff
@@ -60,6 +60,28 @@ def get_corresponding_datasets_names(self) -> List[str]:
...
+            "define_dataset_name"
+            "define_dataset_label"
+            "define_dataset_location"
+            "define_dataset_class"
+            "define_dataset_structure"
+            "define_dataset_is_non_standard"
+            "define_dataset_variables"
+        """
+        directory_path = get_directory_path(self.dataset_path)
+        define_xml_path: str = f"{directory_path}/{DEFINE_XML_FILE_NAME}"
```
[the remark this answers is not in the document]
gerrycampion: \
(gh removed the backslash)

cdisc_rules_engine/operations/variable_exists.py
nhaydel: Fixed with variable_exists operator update

Issue #390, opened 2023-04-21
Add support for define 2.0


gerrycampion:
AC
- Version of define.xml is extracted from the define.xml file. This will come from:
  /ODM/Study/MetaDataVersion/def:DefineVersion
- Define ODM components will be loaded according to the version. At a minimum, we should support define versions 2.0 and 2.1

Implementation details
There should be an abstract base class for the common define.xml components.
Each version of the define.xml should have its own concrete class that overrides version-specific functionality.

Add following test cases to `test_define_xml_reader.py` `test_define_variable_is_collected`:
```python
test_define_2_0_file_path: str = (
    f"{os.path.dirname(__file__)}/../resources/test_defineV20-SDTM.xml"
)
```

```python
        (
            test_define_2_0_file_path,
            "AE",
            "AEACN",
            True,
        ),
        (
            test_define_2_0_file_path,
            "AE",
            "AEBODSYS",
            False,
        ),
        (
            test_define_2_0_file_path,
            "LB",
            "LBCAT",
            False,
        ),
```

RamilCDISC:
The review for this issue is over. The changes are according to the Acceptance Criteria. The code gets the define xml version from /ODM/Study/MetaDataVersion/def:DefineVersion and ODM components are loaded according to the version. We have separate concrete class for each version and one common abstract base class. This meets with the AC hence passing from QA review. 

Following screenshot shows the successful test runs for the changes.

<img width="567" alt="390" src="https://github.com/cdisc-org/cdisc-rules-engine/assets/113539111/dc9dc546-55ce-4978-b7f1-af1850239cb6">

RamilCDISC:
### Validation Summary

### Validation Methodology:
The validation process incorporated rigorous testing methodologies to ensure the accuracy and reliability of the system. This included:
**Unit Testing:**
Unit Testing was done to verify the implementation of the Acceptance criteria. All those tests were then run using pytest to verify compliance with acceptance criteria.
**Code Review:**
The code changes in the code files were reviewed and were found to be in accordance with the Acceptance criteria. Also, the code was reviewed for following: 

- Code Quality Evaluation
- Logic and Functionality Review
- Error Handling and Exception Handling Assessment
- Security Considerations

**Regression Testing:**
Regression tests were run to validate that the changes did not affect any functionality of the overall system.

### Test Execution:
Following Tests cases were checked against the updated code
[...]

PR #391, opened 2023-04-21
ISS 304: added variable_is_null and define_variable_is_collected

gerrycampion:
I deviated slightly from the issue by creating `variable_is_null` instead of `is_not_null` to reduce confusing double negatives.
I renamed the Rule Type from `Define-XML` to `Define Item Metadata Check` to more closely parallel the `Define Item Group Metadata Check`. Currently, no rules existed with the `Define-XML` rule type.
Note that this has been tested for define 2.1, but not define 2.0. Test cases have been added and commented out for define 2.0. Refer to #390 to support define 2.0.

To test, run the test cases.
You can also create a rule using the source issue logic and a define.xml 2.1 file to test the full rule.

### DIFF
--- a/cdisc_rules_engine/operations/variable_is_null.py
+++ b/cdisc_rules_engine/operations/variable_is_null.py
@@ -1,0 +1,9 @@
+from cdisc_rules_engine.operations.base_operation import BaseOperation
+
+
+class VariableIsNull(BaseOperation):
+    def _execute_operation(self):
+        target_variable = self.params.target.replace("--", self.params.domain, 1)
+        if target_variable in self.params.dataframe:
+            series = self.params.dataframe[target_variable]
+            return series.mask(series == "").isnull().all()
--- a/cdisc_rules_engine/operations/variable_is_null.py
+++ b/cdisc_rules_engine/operations/variable_is_null.py
@@ -4,6 +4,7 @@ from cdisc_rules_engine.operations.base_operation import BaseOperation
 class VariableIsNull(BaseOperation):
     def _execute_operation(self):
         target_variable = self.params.target.replace("--", self.params.domain, 1)
-        if target_variable in self.params.dataframe:
-            series = self.params.dataframe[target_variable]
-            return series.mask(series == "").isnull().all()
+        if target_variable not in self.params.dataframe:
+            return True
+        series = self.params.dataframe[target_variable]
+        return series.mask(series == "").isnull().all()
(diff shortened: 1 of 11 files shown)


cdisc_rules_engine/operations/variable_is_null.py
```diff
@@ -0,0 +1,9 @@
+from cdisc_rules_engine.operations.base_operation import BaseOperation
+
+
+class VariableIsNull(BaseOperation):
+    def _execute_operation(self):
+        target_variable = self.params.target.replace("--", self.params.domain, 1)
```
nhaydel, line 6: We probably want to return a default value if the target_variable is not in the dataset.

tests/unit/test_define_xml_reader.py
```diff
@@ -11,6 +11,10 @@
 )
 from cdisc_rules_engine.services.define_xml_reader import DefineXMLReader
 
+test_define_2_0_file_path: str = (
```
nhaydel, line 14: Is it easy to have a test for version 2.0 and 2.1. If not I wouldn't worry about making changes in this ticket

nhaydel (changes requested):

nhaydel (approved):
Approved assuming CI passes

gerrycampion:
> Approved assuming CI passes

Any ideas why it is failing? The tests run successfully locally (though slowly). Seems like a timeout. Should I just try rerunning the CI later?

nhaydel:
> > Approved assuming CI passes
> 
> Any ideas why it is failing? The tests run successfully locally (though slowly). Seems like a timeout. Should I just try rerunning the CI later?

Seems to stall on this one: tests/unit/test_rule_tester/test_rule_tester.py, ran perfectly fine for me locally

nhaydel:
I am seeing a significant slowdown on the test_extract_variable metadata tests on this branch as opposed to the main branch

cdisc_rules_engine/services/define_xml_reader.py
```diff
@@ -273,6 +277,34 @@ def _get_variable_datatype(self, data_type):
...
+    # TODO: subclass define.xml versions
+    # https://github.com/cdisc-org/cdisc-rules-engine/issues/390
+    def _get_origin_type(self, itemdef):
+        if not itemdef.Origin:
+            return
+        version_origin_map = {
+            DefineXMLVersions.DEFINE_XML_2_0_0: lambda origin: origin.Type,
+            DefineXMLVersions.DEFINE_XML_2_1_0: lambda origin: origin[0].Type,
+        }
+        return version_origin_map.get(DefineXMLVersions(self.get_define_version()))(
```
nhaydel, line 289: I think the issue is here. Each time we call get_define_version() we are reading the file again. I think we should either save the version after reading it or find a quicker way  to get the version

nhaydel (changes requested):
We need to avoid calling _get_define_version for each variable in a define file/domain.

cdisc_rules_engine/services/define_xml_reader.py
gerrycampion: Pretty sure this is it. I thought about this, but forgot to revisit. Thanks for looking

nhaydel (approved):

PR #394, opened 2023-04-26
Import models that are imported inline by odmlib to support compiled …

nhaydel:
…define validation

Steps to test

1. Compile core using the pyinstaller command found in the README
2. Run a validation with directory including a define 2.1 file: ex `./dist/core/core validate -s sdtmig -v 3-4 -d msg/`
3. Verify the validation runs completely and an excel file is generated


gerrycampion (approved):

gerrycampion:
I've also created an issue here: https://github.com/swhume/odmlib/issues/14

PR #395, opened 2023-04-28
Add "PyPI Quickstart" section to README.md

Akollek:
In addition to the PyPI docs, this PR also includes a few other markdown formatting fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,17 +239,15 @@ dataset = DatasetVariable(data)
 
 ##### Step 3: Run the (relevant) rules
 
-Next, we need to actually run the rules. We can select which rules we want to run based on the domain we're checking and the domains in the "Include" and "Exclude" lists of the rule dictionary.
+Next, we need to actually run the rules. We can select which rules we want to run based on the domain of the data we're checking and the `"Include"` and `"Exclude"` domains of the rule.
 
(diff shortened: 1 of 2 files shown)


Akollek:
We've added a few steps outlining how we got validation working from within python, but please let us know if there are any omissions or easier ways of going about it!

I'm also aware that this readme gets pulled in as the help text for the command line executable, where the python docs are probably not so relevant. Let me know if you think we should move this docs to a subfolder and just link from the main readme.

mdeforestbrown:
Would be great to update the pypi documentation too for future users who might miss this README if they download the package directly.

Akollek:
> Would be great to update the pypi documentation too for future users who might miss this README if they download the package directly.

Good call, done! Also a point in favour of splitting out the python docs is that we could only show those in PyPI...

nhaydel:
This is great! Thanks!

nhaydel (changes requested):
Everything looks good, just need to fix the linter errors due to the changes in setup.py. You should just be able to run `black setup.py` to get those formatting changes fixed automatically.

Akollek:
Fixed :)

nhaydel (approved):

PR #402, opened 2023-05-04
Support CG0312

gerrycampion:
Related to editor PR: https://github.com/cdisc-org/conformance-rules-editor/pull/183

You can test by running the modified unit tests. Additionally, you can test the combined functionality with the rule editor. The following test package should yield 3 positive and 1 negative result.

Test data file: [CG0312.xlsx](https://github.com/cdisc-org/cdisc-rules-engine/files/11401405/CG0312.xlsx)
Define xml file: [defineV21-SDTM.xml.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/11401406/defineV21-SDTM.xml.txt)
Rule YAML:
```yaml
# Variable: DOMAIN
# Condition: Custom domain present in study
# Rule: DOMAIN not in reserved or modeled standard
Check:
  all:
    - name: define_dataset_is_non_standard
      operator: equal_to
      value: Yes
    - name: $domain_is_custom
      operator: equal_to
      value: false
Operations:
  - id: $domain_is_custom
    operator: domain_is_custom
Core:
  Id: CDISC.SDTMIG.CG0312
  Version: "1"
  Status: Draft
Description:
  Raise an error when custom domains use the reserved or modelled standard
  names
Outcome:
  Message: Custom domains use the reserved or modelled standard names.
Rule Type: Define Item Group Metadata Check
Sensitivity: Dataset
Authorities:
  - Organization: CDISC
    Standards:
      - Name: SDTMIG
        Version: "3.4"
        References:
          - Origin: SDTM and SDTMIG Conformance Rules
            Rule Identifier:
              Id: CG0312
              Version: "1"
            Version: "2.0"
            Citations:
              - Cited Guidance:
                  "'A custom domain may only be created if the data\
                  \ are different in nature and do not fit into an existing published\
                  \ domain.'"
                Document: IG v3.4
                Item: NA
                Section: "2.6"
      - Name: SDTMIG
        Version: "3.2"
        References:
          - Origin: SDTM and SDTMIG Conformance Rules
            Rule Identifier:
              Id: CG0312
              Version: "1"
            Version: "2.0"
            Citations:
              - Cited Guidance:
                  The following standard domains, listed in alphabetical
                  order by Domain Code, with their respective domain codes have been
                  defined or referenced by the CDISC SDS Team in this document. Note
                  that other domain models may be posted separately for comment after
                  this publication. See list in 2.5
                Document: IG v3.2
                Section: "2.5"
      - Name: SDTMIG
        Version: "3.3"
        References:
          - Origin: SDTM and SDTMIG Conformance Rules
            Rule Identifier:
              Id: CG0312
              Version: "1"
            Version: "2.0"
            Citations:
              - Cited Guidance:
                  A custom domain may only be created if the data are
                  different in nature and do not fit into an existing published domain.
                Document: IG v3.3
                Section: "2.6"
Scope:
  Classes:
    Include:
      - ALL
  Domains:
    Include:
      - ALL
```

### DIFF
--- a/cdisc_rules_engine/operations/domain_is_custom.py
+++ b/cdisc_rules_engine/operations/domain_is_custom.py
@@ -19,10 +19,8 @@ class DomainIsCustom(BaseOperation):
         standard_data: dict = self.cache.get(cache_key)
         if standard_data is None:
             cdisc_library_service = CDISCLibraryService(config, self.cache)
-            standard_data = set(
-                cdisc_library_service.get_standard(
-                    self.params.standard.lower(), self.params.standard_version
-                )
+            standard_data = cdisc_library_service.get_standard_details(
+                self.params.standard.lower(), self.params.standard_version
             )
-            self.cache.set(cache_key, standard_data)
+            self.cache.add(cache_key, standard_data)
         return self.params.domain not in standard_data.get("domains", {})
--- a/cdisc_rules_engine/operations/domain_is_custom.py
+++ b/cdisc_rules_engine/operations/domain_is_custom.py
@@ -1,0 +1,28 @@
+from cdisc_rules_engine.operations.base_operation import BaseOperation
+from cdisc_rules_engine.utilities.utils import (
+    get_standard_details_cache_key,
+)
+from cdisc_rules_engine.services.cdisc_library_service import CDISCLibraryService
+from cdisc_rules_engine import config
+
+
+class DomainIsCustom(BaseOperation):
+    def _execute_operation(self):
+        """
+        Gets standard details from cache and checks if
+        given domain is in standard domains.
+        If no -> the domain is custom.
+        """
+        cache_key = get_standard_details_cache_key(
+            self.params.standard, self.params.standard_version
+        )
+        standard_data: dict = self.cache.get(cache_key)
+        if standard_data is None:
+            cdisc_library_service = CDISCLibraryService(config, self.cache)
+            standard_data = set(
+                cdisc_library_service.get_standard(
+                    self.params.standard.lower(), self.params.standard_version
+                )
+            )
+            self.cache.set(cache_key, standard_data)
+        return self.params.domain not in standard_data.get("domains", {})
--- a/cdisc_rules_engine/services/data_services/dummy_data_service.py
+++ b/cdisc_rules_engine/services/data_services/dummy_data_service.py
@@ -127,7 +127,7 @@ class DummyDataService(BaseDataService):
             with open(dataset_name, "rb") as f:
                 return f.read()
 
-        return bytes(self.define_xml)
+        return self.define_xml.encode()
 
     def has_all_files(self, prefix: str, file_names: List[str]) -> bool:
         return True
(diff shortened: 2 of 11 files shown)


cdisc_rules_engine/operations/domain_is_custom.py
```diff
@@ -0,0 +1,28 @@
...
+        given domain is in standard domains.
+        If no -> the domain is custom.
+        """
+        cache_key = get_standard_details_cache_key(
+            self.params.standard, self.params.standard_version
+        )
+        standard_data: dict = self.cache.get(cache_key)
+        if standard_data is None:
+            cdisc_library_service = CDISCLibraryService(config, self.cache)
+            standard_data = set(
```
nhaydel, line 22: standard data should be a map right?

cdisc_rules_engine/services/data_services/dummy_data_service.py
```diff
@@ -127,7 +127,7 @@ def get_define_xml_contents(self, dataset_name: str) -> bytes:
             with open(dataset_name, "rb") as f:
                 return f.read()
 
-        return bytes(self.define_xml)
+        return self.define_xml.encode()
```
nhaydel, line 130: What is this change for?

tests/unit/test_rule_tester/test_rule_tester.py
```diff
@@ -206,7 +206,7 @@ def test_rule_with_define_xml(define_xml_variable_validation_rule: dict):
         }
     ]
 
-    with open(test_define_file_path, "rb") as file:
+    with open(test_define_file_path, "r") as file:
```
nhaydel, line 209: Why change this if we are expecting the file in bytes

cdisc_rules_engine/services/data_services/dummy_data_service.py
gerrycampion: I was getting this error when running from the rule tester: `string argument without an encoding.`
str.encode() provides a default encoding

cdisc_rules_engine/operations/domain_is_custom.py
gerrycampion: I fixed this along with the domain_label op where I copied from.

tests/unit/test_rule_tester/test_rule_tester.py
gerrycampion: expecting in str, so i updated the type

cdisc_rules_engine/services/data_services/dummy_data_service.py
nhaydel: Doesn't this change the return type of the method. I think this get define_xml method is expected to return bytes in the local data service. I think dummy data service should match

gerrycampion: no, they both do the same thing and return bytes.
`self.define_xml.encode()` is the same as `bytes(self.define_xml, 'utf-8')`

nhaydel (approved):

PR #403, opened 2023-05-05
Qa regression tests and issue383

RamilCDISC:
This PR adds a new folder QARegressionTests to the project, which contains regression tests for the application. The folder also includes the required resources for running the tests, which have been added to the resources folder in both the test folder and the root folder of the project.

Other than that same PR will be used to review the AC of issue 383.

To run the tests use pytest in the terminal. The command will be as follows
pytest QARegressionTests/cli_tests.py


nhaydel (changes requested):
It would be nice to move the regression tests to a folder under tests (i.e tests/regression)
Also it would be nice to break up the tests file to other files for each piece of functionality

i.e 
test_validation
test_test_command
test_list_rules_command
etc...

PR #405, opened 2023-05-08
Added support for define xml 2.0

gerrycampion:
To verify, you can run the new unit tests in test_define_xml_reader.py

### DIFF
--- a/cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
+++ b/cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
@@ -267,11 +267,13 @@ class BaseDefineXMLReader(ABC):
         }
         return variable_type_map.get(data_type, data_type)
 
+    @abstractmethod
     def _get_origin_type(self, itemdef):
-        return itemdef.Origin[0].Type if itemdef.Origin else None
+        pass
 
+    @abstractmethod
     def _get_variable_is_collected(self, itemdef):
-        return self._get_origin_type(itemdef) == "Collected" if itemdef.Origin else None
+        pass
 
     def _get_metadata_representation(self, metadata) -> dict:
         """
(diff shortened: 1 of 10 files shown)


cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
```diff
@@ -286,33 +267,11 @@ def _get_variable_datatype(self, data_type):
...
-        if not itemdef.Origin:
-            return
-        version_origin_map = {
-            DefineXMLVersions.DEFINE_XML_2_0_0: lambda origin: origin.Type,
-            DefineXMLVersions.DEFINE_XML_2_1_0: lambda origin: origin[0].Type,
-        }
-        return version_origin_map.get(DefineXMLVersions(self.get_define_version()))(
-            itemdef.Origin
-        )
+        return itemdef.Origin[0].Type if itemdef.Origin else None
```
nhaydel, line 271: NIT: I think we should keep version specific implementations out of the base class

cdisc_rules_engine/services/define_xml/define_xml_reader_factory.py
```diff
@@ -0,0 +1,102 @@
...
+            ),
+            None,
+        )
+        return define_version
+
+    @classmethod
+    def _from_namespace_version(
+        cls, namespace: str, version: str
+    ) -> BaseDefineXMLReader:
+        for define_xml_reader in cls._define_xml_readers:
```
nhaydel, line 97: This can be written using next and list comprehension similar to the above methods

tests/unit/test_define_xml_reader.py
```diff
@@ -9,8 +9,16 @@
...
 )
-from cdisc_rules_engine.services.define_xml_reader import DefineXMLReader
+from cdisc_rules_engine.services.define_xml.base_define_xml_reader import (
+    BaseDefineXMLReader,
+)
+from cdisc_rules_engine.services.define_xml.define_xml_reader_factory import (
+    DefineXMLReaderFactory,
+)
 
+test_define_2_0_file_path: str = (
```
nhaydel, line 19: Can we construct these file paths using os.path.join?

nhaydel (commented):
Left some very minor comments. Looks really good though.

nhaydel (approved):

PR #406, opened 2023-05-09
Add name and label referenced variable metadata operations

nhaydel:
This PR adds operations for getting referenced variable metadata from the library based on name or label

Steps to test:

1. Run the unit tests

### DIFF
--- a/cdisc_rules_engine/operations/label_referenced_variable_metadata.py
+++ b/cdisc_rules_engine/operations/label_referenced_variable_metadata.py
@@ -1,0 +1,23 @@
+from cdisc_rules_engine.operations.base_operation import BaseOperation
+import pandas as pd
+
+
+class LabelReferencedVariableMetadata(BaseOperation):
+    def _execute_operation(self):
+        """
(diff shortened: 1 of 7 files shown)


cdisc_rules_engine/operations/label_referenced_variable_metadata.py
```diff
@@ -0,0 +1,23 @@
+from cdisc_rules_engine.operations.base_operation import BaseOperation
+import pandas as pd
+
+
+class LabelReferencedVariableMetadata(BaseOperation):
```
gerrycampion, line 5: Wouldn't it be better to have a single Operation where we would pass in "name" or "label" as an additional parameter?

nhaydel: We could, but I figured it would be easier on the rule authors to know there are two reference operations than to remember a specific parameter for this operation. I also didn't think that parameter would be used elsewhere so I didn't think we should extend the schema

gerrycampion (approved):

PR #436, opened 2023-06-07
ISS 274 - New rule type Value Check against Define XML VLM

gerrycampion:
New rule type created: `Value Check against Define XML VLM`
New metadata columns created for VLM: `define_vlm_...`

I tried best to avoid a cartesian product when matching dataset rows with applicable VLM rows. I'm hoping a nested df.assign is better than a df.merge with `cross`.

To test:
- run the new unit test
- Running with the attached rule, data, and define.xml should produce the attached results (2 negatives).

[Rule.zip](https://github.com/cdisc-org/cdisc-rules-engine/files/11680466/Rule.zip)


tests/unit/test_dataset_builders/test_contents_define_vlm_dataset_builder.py
```diff
@@ -0,0 +1,81 @@
...
+        rule_processor=None,
+        data_processor=None,
+        dataset_path=None,
+        datasets=None,
+        domain=None,
+    ).build()
+    expected_df = pd.DataFrame.from_dict(expected)
+    expected_df.sort_index(axis=1, inplace=True)
+    result.sort_index(axis=1, inplace=True)
+    print(result.to_string())
```
nhaydel, line 80: NIT: you can remove this

nhaydel (approved):
Overall looks really good! I just noticed a print statement left in

PR #438, opened 2023-06-12
Add function to get ordered key variables for dataset

nhaydel:
This PR adds a function to get the key variables for a domain. Currently it is a single function to be used in the future by other issues.

Steps to test:
1. Run the unit tests

### DIFF
--- a/cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
+++ b/cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
@@ -2,7 +2,7 @@ from abc import abstractmethod, ABC
 from dataclasses import dataclass
 from typing import List, Optional
 from functools import cache
-
+from heapq import heappush, heappop
 
 from odmlib.define_loader import XMLDefineLoader
 from odmlib.loader import ODMLoader
@@ -60,6 +60,16 @@ class BaseDefineXMLReader(ABC):
         self.cache_service = cache_service_obj
         self.study_id = study_id
         self.data_bundle_id = data_bundle_id
+
+    @cache
+    def get_item_def_map(self) -> dict:
+        metadata = self._odm_loader.MetaDataVersion()
+        return {item.OID: item for item in metadata.ItemDef}
+
+    @cache
+    def get_value_list_def_map(self) -> dict:
+        metadata = self._odm_loader.MetaDataVersion()
+        return {value_list.OID: value_list for value_list in metadata.ValueListDef}
 
     def read(self) -> List[dict]:
         """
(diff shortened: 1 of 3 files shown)


tests/unit/test_define_xml_reader.py
```diff
@@ -290,3 +290,12 @@ def test_extract_variable_metadata_not_found():
...
         with pytest.raises(DomainNotFoundInDefineXMLError):
             reader.extract_variables_metadata(domain_name="not found domain")
+
+
+def test_get_domain_key_sequence():
+    with open(test_define_file_path, "rb") as file:
+        contents: bytes = file.read()
+        reader = DefineXMLReaderFactory.from_file_contents(contents)
+        ec_key_sequence: dict = reader.get_domain_key_sequence(domain_name="EC")
+        print(ec_key_sequence)
```
gerrycampion, line 300: remove print

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
```diff
@@ -320,3 +335,73 @@ def get_define_version(self) -> Optional[str]:
...
         return None
+
+    def get_domain_key_sequence(self, domain_name: str) -> List[str]:
+        """
+        Given a domain name, this function returns the key sequence variable names
+        in order of KeySequence number.
+
+        In the case that Supplemental qualifiers apply to a domain:
+
+        # TODO! How should this be handled - add to end of list
```
gerrycampion, line 346: is this still in question?

tests/unit/test_define_xml_reader.py
```diff
@@ -290,3 +290,12 @@ def test_extract_variable_metadata_not_found():
...
+        assert ec_key_sequence == ["STUDYID", "USUBJID", "ECTRT", "ECSTDTC", "ECREASOC"]
```
gerrycampion, line 301: Nice, it looks like this is doing EC how I would expect. I think you should also add a test for a SUPP dataset as well.
For example, the SUPPEC keys in this define.xml should be:
STUDYID, RDOMAIN, USUBJID, IDVAR, IDVARVAL, QNAM

gerrycampion (changes requested):

tests/unit/test_define_xml_reader.py
nhaydel: Good call, this does need some updating to work with supp datasets

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
gerrycampion, line 1: Sorry, I didn't notice this before, but don't we need to add to `_get_metadata_representation` for this, something like `define_dataset_key_sequence`?
Remember to add to the other define_dataset_ locations, like the builder comments and schema.

nhaydel: @gerrycampion I assumed that would be added in the future PRs where it is needed

gerrycampion (approved):

PR #441, opened 2023-06-15
382 add define-xml-path

mdeforestbrown:
Not 100% sure on the scope of this change, it seems we point to the define-xml from a few different places so there are quite a few changes here. Let me know if I'm on the right track and how best to add more tests for this.


core.py
```diff
@@ -68,6 +68,7 @@ def cli():
     multiple=True,
     help="Absolute path to dataset file",
 )
+@click.option("-dxp", "--define-xml-path", default="test", help="Path to Define-XML")
```
nhaydel, line 71: I don't think we want a default here. I think we should add required=False

nhaydel (changes requested):
I really like this. Only one small comment about the flag. Also there is one test failing, are you able to see that?

nhaydel:
@mdeforestbrown just need to fix the lint errors. You can install the linter in the github repo by running `pre-commit install` in the root of the repo

scripts/run_validation.py
```diff
@@ -57,6 +57,7 @@ def validate_single_rule(cache, datasets, args: Validation_args, rule: dict = No
         ct_package=args.controlled_terminology_package,
         meddra_path=args.meddra,
         whodrug_path=args.whodrug,
+        define_xml_path=args.define_xml_path,
```
nhaydel, line 60: Also, does the report get the correct define version if a define_xml_path is specified?

nhaydel (commented):
I have a couple questions. Assuming those are handled, it should be good to go

cdisc_rules_engine/dataset_builders/dataset_builder_factory.py
```diff
@@ -76,5 +76,6 @@ def get_service(
             kwargs.get("dataset_path"),
             kwargs.get("datasets"),
             kwargs.get("domain"),
+            kwargs.get("define_xml_path"),
```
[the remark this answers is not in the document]
mdeforestbrown: Oh yes my mistake, added.

scripts/run_validation.py
mdeforestbrown: Okay I have updated as discussed yesterday. 

If I move the define.xml to a separate folder and point to it with the path command it is now populating correctly for both the excel and the json report.

Screenshot 2023-07-21 at 10 32 27
Screenshot 2023-07-21 at 10 34 14

tests/QARegressionTests/test_core/test_test_command.py
```diff
@@ -28,6 +28,8 @@ def test_test_command_with_all_options(self):
                 "3.4",
                 "-dv",
                 os.path.join("tests", "resources", "report_test_data", "define.xml"),
+                "-dxp",
```
nhaydel, line 31: Does this test fail without this argument? I don't think we want it to

mdeforestbrown: > Does this test fail without this argument?

No it doesn't fail without this argument. I'd add a separate test but assumed it was covered by the other tests in this file without the argument?

nhaydel (approved):

PR #442, opened 2023-06-16
Add new keys to perform_rule_operations and fix regression tests

nhaydel:


tests/QARegressionTests/test_core/test_validate.py
```diff
@@ -174,22 +158,6 @@ def test_validate_less_than_minimum_options(self):
         self.assertEqual(result.exit_code, 2)
         self.assertIn("Error: Missing option", result.output)
 
-    def test_validate_invalid_value_to_option_version(self):
```
nhaydel, line 177: Removed because this check does not apply

tests/QARegressionTests/test_core/test_validate.py
```diff
@@ -63,22 +63,6 @@ def test_validate_required_v_option_present(self):
         )
         self.assertEqual(result.exit_code, 0)
 
-    def test_validate_both_d_and_data_options(self):
```
nhaydel, line 66: Removed because this check does not apply

tests/QARegressionTests/test_core/test_validate.py
RamilCDISC: I would highly recommend keeping this check intact. It is essential to consider scenarios where a user may inadvertently provide an invalid value to the -v option during a validation process. While we can modify the assertion part to accommodate such cases, it would be beneficial for the rule engine to handle these situations, and it is equally important to thoroughly test it as well. Preserving this check will contribute to the overall robustness and reliability of the system.

nhaydel: I agree, but it is checking functionality that we don't have. I think when the version checks are added it should be added back as well. As of right now it will always fail since that functionality has not been written

RamilCDISC (approved):

PR #443, opened 2023-06-16
Iss 295 validator crashes when defstructure is absent in the definexml

htu:
1. Download the following test files: 
[define-absent.xml.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/11772332/define-absent.xml.txt)
[define-present.xml.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/11772333/define-present.xml.txt)
[rule-send6.json.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/11772334/rule-send6.json.txt)
[test-dataset-send.json.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/11772336/test-dataset-send.json.txt)

2. Put the files in your local computer such as ./dist/issue295

3. Remove the file extension .txt from file names

4. Copy define-absent.xml to define.xml and then run the following commands (only difference is the -vx): 
`python core.py test -s sdtmig -v 3-4 -dp ./dist/issue295/test-dataset-send.json -r ./dist/issue295/rule-send6.json -dv 2.1  -vx y -l trace`
`python core.py test -s sdtmig -v 3-4 -dp ./dist/issue295/test-dataset-send.json -r ./dist/issue295/rule-send6.json -dv 2.1  -vx n -l trace`
You will get trace information in both commands. The first you will get "SUCCESS" and the second one with "SKIPPED" 

5. Copy define-present.xml to define.xml and then run the following commands: 
`python core.py test -s sdtmig -v 3-4 -dp ./dist/issue295/test-dataset-send.json -r ./dist/issue295/rule-send6.json -dv 2.1  -vx y -l trace`
`python core.py test -s sdtmig -v 3-4 -dp ./dist/issue295/test-dataset-send.json -r ./dist/issue295/rule-send6.json -dv 2.1  -vx n -l trace`
You should not get any trace information in both commands and both will get "SUCCESS" in the report.


cdisc_rules_engine/dataset_builders/base_dataset_builder.py
```diff
@@ -42,6 +42,7 @@ def build(self) -> pd.DataFrame:
         pass
 
     def get_dataset(self, **kwargs):
+        dataset = pd.DataFrame()
```
nhaydel, line 45: Is this line necessary

cdisc_rules_engine/dataset_builders/variables_metadata_with_define_dataset_builder.py
```diff
@@ -28,16 +29,26 @@ def build(self):
         define_variable_has_codelist,
         define_variable_codelist_coded_values
         """
+        vx = kwargs.get("validate_xml", "Y")
+        jm = kwargs.get("join_method", "left")
```
gerrycampion, line 33: I think we agreed that we don't want join_method to be an option here?

cdisc_rules_engine/rules_engine.py
```diff
@@ -240,7 +256,18 @@ def validate_rule(
                     self.standard, self.standard_version
                 )
             )
-            define_metadata: List[dict] = builder.get_define_xml_variables_metadata()
+            if vx in ("Y", "YES"):
```
nhaydel, line 259: I think we can isolate this check in the define_xml_reader class instead of checking it at this level. That way the engine doesn't care if the validate xml flag is set, it just runs rules

cdisc_rules_engine/rules_engine.py
```diff
@@ -155,6 +166,7 @@ def validate_single_rule(
                 error_obj.domain = dataset_domain
                 return [error_obj.to_representation()]
         except Exception as e:
+            logger.trace(e, __name__)
```
nhaydel, line 169: Do we need this trace?

cdisc_rules_engine/rules_engine.py
```diff
@@ -134,9 +136,18 @@ def validate_single_rule(
                 is_split_dataset(datasets, dataset_domain),
                 datasets,
             ):
-                result: List[Union[dict, str]] = self.validate_rule(
-                    rule, dataset_path, datasets, dataset_domain
-                )
+                result = None
+                if vx in ("Y", "YES"):
```
gerrycampion, line 140: just use self.validate_xml here instead of introducing a new vaguely-named intermediate variable

cdisc_rules_engine/services/logging/console_logger.py
```diff
@@ -52,3 +55,28 @@ def critical(self, msg: str, *args, **kwargs):
...
+            traceback_info = traceback.extract_tb(exc.__traceback__)
+            if traceback_info:
+                line_number = traceback_info[-1].lineno
+                caller_function = traceback_info[-1].name
+            print(f"Error: {caller_function}/{line_number}: {str(exc)}")
+            i = 0
+            for call_func in traceback_info:
+                i += 1
+                print(f"Trace({i}): {call_func}")
+        # self._logger.log(logging.CRITICAL + 2, msg, *args, **kwargs)
```
gerrycampion, line 82: remove commented code

cdisc_rules_engine/rules_engine.py
```diff
@@ -126,6 +127,7 @@ def validate_single_rule(
             f"Validating domain {dataset_domain}. "
             f"rule={rule}. dataset_path={dataset_path}. datasets={datasets}."
         )
+        vx = self.validate_xml
```
nhaydel, line 130: 1. Is this variable needed?
2. If it is can the name be more descriptive?

cdisc_rules_engine/rules_engine.py
nhaydel, line 140: It seems like these are just trying to bubble up exceptions that occur during xml validation. I think they should be logged and handled in the reader so we don't have this check for vx every time we have to get define metadata.

gerrycampion (changes requested):

cdisc_rules_engine/operations/base_operation.py
```diff
@@ -49,12 +49,12 @@ def __init__(
         self.evaluation_dataset = original_dataset
 
     @abstractmethod
-    def _execute_operation(self):
+    def _execute_operation(self, **kwargs):
```
nhaydel, line 52: If we add kwargs to these methods, I think we'd have to add the argument to all operations. If that's the case, I think it is better to store the data we need from kwargs in the class instance. I.e set these values as instance variables when we create the instance of the operation instead of passing them to the operation execution method

nhaydel (changes requested):

cdisc_rules_engine/dataset_builders/variables_metadata_with_define_dataset_builder.py
htu: Removed join_method.

core.py
```diff
@@ -359,7 +394,11 @@ def test(
     whodrug: str,
     meddra: str,
     rule: str,
+    validate_xml,
+    log_level,
 ):
+    vx = validate_xml.upper()
+    ll = log_level.lower()
```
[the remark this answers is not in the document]
htu: Removed the short variable names.

cdisc_rules_engine/rules_engine.py
htu: Rmoved short variable name.

cdisc_rules_engine/rules_engine.py
htu: Used self variable name.

cdisc_rules_engine/rules_engine.py
htu: I would leave it here since when the error message is shown only when logging is set to "error".

cdisc_rules_engine/rules_engine.py
htu: Could we do it in another ticket for up-lifting it to reader level?

cdisc_rules_engine/dataset_builders/base_dataset_builder.py
htu: Indeed, not needed.

scripts/test_rule.py
```diff
@@ -44,7 +44,8 @@ class CacheManager(SyncManager):
 
 
 def validate_single_rule(cache, path, args, datasets, rule: dict = None):
-    set_log_level("ERROR")
+    # set_log_level("ERROR")
```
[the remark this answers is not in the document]
htu: Removed the commented line.

scripts/test_rule.py
```diff
@@ -66,14 +68,14 @@ def validate_single_rule(cache, path, args, datasets, rule: dict = None):
...
-                    dataset.domain,
-                )
+            validated_result = engine.test_validation(
+                rule,
+                os.path.join(directory, dataset.filename),
+                datasets,
+                dataset.domain,
             )
+            results.append(validated_result)
+            # logger.info(f"Done validating {dataset.domain}")
```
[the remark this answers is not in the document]
htu: The logger is not availabe in this class. I was not sure whether I should add the logger here.

scripts/test_rule.py
```diff
@@ -86,7 +88,8 @@ def set_log_level(level: str):
 
 
 def test(args: TestArgs):
-    set_log_level("ERROR")
+    # set_log_level("ERROR")
```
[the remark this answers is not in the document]
htu: Done.

cdisc_rules_engine/services/logging/console_logger.py
htu: Done

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
```diff
@@ -93,13 +93,25 @@ def extract_domain_metadata(self, domain_name: str = None) -> dict:
...
+        vx = kwargs.get("validate_xml", "Y")
         logger.info(
             f"Extracting variables metadata from Define-XML. domain_name={domain_name}"
         )
-        metadata = self._odm_loader.MetaDataVersion()
-        domain_metadata = self._get_domain_metadata(metadata, domain_name)
+        metadata = None
+        domain_metadata = None
         variables_metadata = []
+        try:
```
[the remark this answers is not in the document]
htu: 1. I agree that it might need a new method to handle this. As to this validate-xml flag, it serves the purpose for now. 
2. The definition of validate-xml is to validate Define-XML or not when it misses def:Structure or in error:
      Y (skip define rule validation and continue for other rules)
       N (stop on error and report the error)

cdisc_rules_engine/operations/base_operation.py
htu: It seems that the rule instance variables are not passed to operation class. That is why I had to add it here.

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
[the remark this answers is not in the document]
nhaydel: It may serve it's purpose for this specific issue, but I don't think it is much more work to create a new method for the validation. I'm not saying you need to apply it in each function, but having it available will probably be useful in the future.

cdisc_rules_engine/rules_engine.py
nhaydel: I would rather not since it would just be knowingly adding tech debt, but if @gerrycampion agrees to do it later, I would be ok with creating a new issue for it

cdisc_rules_engine/rules_engine.py
nhaydel: How is that different from having to set the logging level to "trace"

cdisc_rules_engine/operations/base_operation.py
nhaydel: We should pass whatever instance variables we need to the operations class then

cdisc_rules_engine/rules_engine.py
htu: The trace information is only shown when it is set to trace. I think it might be better to enable trace when the log level is set to error. Do we want to know the trace information when it has error?

core.py
```diff
@@ -146,6 +146,17 @@ def cli():
         "is printed."
     ),
 )
+@click.option(
```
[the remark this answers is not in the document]
htu: Added choices.

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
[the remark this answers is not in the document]
htu: Added get_metadata_version

cdisc_rules_engine/rules_engine.py
gerrycampion: didnt see this before - i agree with nic that we don't need this outside the define_xml_reader

cdisc_rules_engine/dataset_builders/variables_metadata_with_define_dataset_builder.py
```diff
@@ -28,13 +29,22 @@ def build(self):
         define_variable_has_codelist,
         define_variable_codelist_coded_values
         """
+        validate_xml = kwargs.get("validate_xml", "Y")
+        logger = logging.getLogger()
```
nhaydel, line 33: We should use the logger from: https://github.com/cdisc-org/cdisc-rules-engine/blob/main/cdisc_rules_engine/services/logging/logging_service_factory.py#L35

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
```diff
@@ -108,10 +111,27 @@ def extract_variables_metadata(self, domain_name: str = None) -> List[dict]:
                     self._get_item_def_representation(itemdef[0], itemref, codelist_map)
                 )
         logger.info(f"Extracted variables metadata = {variables_metadata}")
+
         return variables_metadata
 
+    @cached("define-metadata-version")
```
nhaydel, line 117: Couple of notes on this function:

1. We've recently had some issues caching objects, so I think we should not cache here
2. We should not have to set an _exception value on the logger, for any reason. The log message should be formatted using the exception before calling the logger function.

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
```diff
@@ -108,10 +111,27 @@ def extract_variables_metadata(self, domain_name: str = None) -> List[dict]:
...
+    def get_metadata_version(self, **kwargs):
+        validate_xml = kwargs.get("validate_xml", "Y")
+        metadata = None
+        try:
+            metadata = self._odm_loader.MetaDataVersion()
+            logger.info(f"Define Metadata Version: {metadata}")
+        except Exception as e:
+            logger._exception = e
+            logger.error(f"{__name__}(VX={validate_xml})")
+            if validate_xml in ("Y", "YES"):
```
nhaydel, line 127: This line appears a lot across multiple files. Can we at least get it to be a method and call the method instead of having this check in multiple places

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
```diff
@@ -108,10 +111,27 @@ def extract_variables_metadata(self, domain_name: str = None) -> List[dict]:
                     self._get_item_def_representation(itemdef[0], itemref, codelist_map)
                 )
         logger.info(f"Extracted variables metadata = {variables_metadata}")
+
         return variables_metadata
 
+    @cached("define-metadata-version")
+    def get_metadata_version(self, **kwargs):
+        validate_xml = kwargs.get("validate_xml", "Y")
```
nhaydel, line 119: It also seems like we are using kwargs in a lot of places just to pass around the validate_xml flag. Can that not be set at object initialization so we don't have to have kwargs everywhere?

nhaydel (changes requested):

cdisc_rules_engine/dataset_builders/base_dataset_builder.py
```diff
@@ -75,26 +78,34 @@ def get_define_xml_item_group_metadata(self, domain: str) -> List[dict]:
...
         define_xml_reader = DefineXMLReaderFactory.from_file_contents(
-            define_xml_contents, cache_service_obj=self.cache
+            define_xml_contents,
+            cache_service_obj=self.cache,
+            validate_xml=self.validate_xml,
         )
         return define_xml_reader.extract_domain_metadata(domain)
 
-    def get_define_xml_variables_metadata(self) -> List[dict]:
+    def get_define_xml_variables_metadata(self, **kwargs) -> List[dict]:
```
nhaydel, line 94: Does this method still need kwargs?

cdisc_rules_engine/dataset_builders/base_dataset_builder.py
```diff
@@ -75,26 +78,34 @@ def get_define_xml_item_group_metadata(self, domain: str) -> List[dict]:
...
         define_xml_contents: bytes = self.data_service.get_define_xml_contents(
             dataset_name=define_xml_path
         )
         define_xml_reader = DefineXMLReaderFactory.from_file_contents(
-            define_xml_contents, cache_service_obj=self.cache
+            define_xml_contents,
+            cache_service_obj=self.cache,
+            validate_xml=self.validate_xml,
+        )
+        return define_xml_reader.extract_variables_metadata(
```
nhaydel, line 109: Same as above

cdisc_rules_engine/dataset_builders/define_variables_dataset_builder.py
```diff
@@ -4,7 +4,7 @@
 
 
 class DefineVariablesDatasetBuilder(BaseDatasetBuilder):
-    def build(self):
+    def build(self, **kwargs):
```
nhaydel, line 7: I think we need to remove kwargs from all builders

cdisc_rules_engine/rules_engine.py
```diff
@@ -240,7 +265,20 @@ def validate_rule(
...
                 )
             )
-            define_metadata: List[dict] = builder.get_define_xml_variables_metadata()
+            if self.is_validate_xml:
+                try:
+                    define_metadata: List[
+                        dict
+                    ] = builder.get_define_xml_variables_metadata(**kwargs)
+                except Exception as e:
+                    logger._exception = e
```
nhaydel, line 274: This exception setting should be removed.

cdisc_rules_engine/rules_engine.py
```diff
@@ -240,7 +265,20 @@ def validate_rule(
...
+                    logger.error()
+                    # logger.trace(e, __name__)
+            else:
+                define_metadata: List[dict] = builder.get_define_xml_variables_metadata(
```
nhaydel, line 278: Are you doing anything with this define metadata?

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
```diff
@@ -108,10 +116,27 @@ def extract_variables_metadata(self, domain_name: str = None) -> List[dict]:
                     self._get_item_def_representation(itemdef[0], itemref, codelist_map)
                 )
         logger.info(f"Extracted variables metadata = {variables_metadata}")
+
         return variables_metadata
 
+    def get_metadata_version(self, **kwargs):
```
nhaydel, line 122: We are not using kwargs here

cdisc_rules_engine/rules_engine.py
```diff
@@ -66,10 +66,14 @@ def __init__(
         self.ct_package = kwargs.get("ct_package")
         self.meddra_path: str = kwargs.get("meddra_path")
         self.whodrug_path: str = kwargs.get("whodrug_path")
+        self.validate_xml: str = kwargs.get("validate_xml")
 
     def get_schema(self):
         return export_rule_data(DatasetVariable, COREActions)
 
+    def is_validate_xml(self):
```
nhaydel, line 74: Does this need to be defined here and in the base_define_reader class? Maybe it should be a utility method

cdisc_rules_engine/services/reporting/excel_report.py
```diff
@@ -70,12 +70,12 @@ def get_export(
             wb["Conformance Details"]["B14"] = self._args.whodrug
         return wb
 
-    def write_report(self):
+    def write_report(self, **kwargs):
```
nhaydel, line 73: Do we need kwargs here?

nhaydel (changes requested):

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
nhaydel: This needs to be addressed still

cdisc_rules_engine/services/reporting/excel_report.py
```diff
@@ -86,6 +86,7 @@ def write_report(self):
             with open(self._output_name, "wb") as f:
                 f.write(excel_workbook_to_stream(report_data))
         except Exception as e:
+            logger._exception = e
```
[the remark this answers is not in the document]
nhaydel: This still needs to be changed back

nhaydel:
@htu A lot of my original comments are still unresolved. Also there are merge conflicts which should be resolved

cdisc_rules_engine/dataset_builders/define_variables_dataset_builder.py
nhaydel: This is still relevant

cdisc_rules_engine/rules_engine.py
```diff
@@ -134,9 +138,21 @@ def validate_single_rule(
                 is_split_dataset(datasets, dataset_domain),
                 datasets,
             ):
-                result: List[Union[dict, str]] = self.validate_rule(
-                    rule, dataset_path, datasets, dataset_domain
-                )
+                result = None
+                if self.is_validate_xml:
```
nhaydel, line 142: do we need to check this here? If an exception occurs won't it be logged in the outer `except` block and logged? If the goal is to create an empty results list, you will need to set result to an empty list when catching the exception, but I think we can just check this flag in the outer exception block

cdisc_rules_engine/rules_engine.py
```diff
@@ -187,9 +205,17 @@ def validate_rule(
...
         kwargs = {}
+        kwargs["validate_xml"] = self.validate_xml
+
+        builder = self.get_dataset_builder(
+            rule,
+            dataset_path,
+            datasets,
+            domain,
+        )
+        dataset = builder.get_dataset(**kwargs)
```
nhaydel, line 217: I don't think the builder.get_dataset should need the kwargs

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
```diff
@@ -60,18 +61,22 @@ def __init__(
         self.cache_service = cache_service_obj
         self.study_id = study_id
         self.data_bundle_id = data_bundle_id
+        self.validate_xml = validate_xml
 
-    def read(self) -> List[dict]:
+    def read(self, **kwargs) -> List[dict]:
```
nhaydel, line 66: is kwargs still needed here?

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
```diff
@@ -132,7 +155,7 @@ def extract_value_level_metadata(self, domain_name: str = None) -> List[dict]:
             f"Extracting value level metadata from Define-XML. "
             f"domain_name={domain_name}"
         )
-        metadata = self._odm_loader.MetaDataVersion()
+        metadata = self.get_metadata_version(**kwargs)
```
nhaydel, line 158: The method doesn't accept kwargs. It should be removed

nhaydel (changes requested):
Almost there, some more kwargs need to be removed and I had one question about how we are handling the validation result in the rules engine

cdisc_rules_engine/operations/base_operation.py
```diff
@@ -53,8 +53,8 @@ def _execute_operation(self):
         """Perform operation calculations."""
         pass
 
-    def execute(self) -> pd.DataFrame:
-        result = self._execute_operation()
+    def execute(self, **kwargs) -> pd.DataFrame:
```
nhaydel, line 56: I don't think we need kwargs here

cdisc_rules_engine/services/define_xml/base_define_xml_reader.py
```diff
@@ -118,9 +124,21 @@ def extract_variables_metadata(self, domain_name: str = None) -> List[dict]:
...
 
-    def extract_value_level_metadata(self, domain_name: str = None) -> List[dict]:
+    def get_metadata_version(self):
+        metadata = None
+        logger.info(f"Validate XML = {self.validate_xml}:{self.is_validate_xml}")
+        metadata = self._odm_loader.MetaDataVersion()
+        logger.info(f"Define Metadata Version: {metadata}")
+        return metadata
+
+    @cached("define-value-level-metadata")
```
nhaydel, line 137: We should not cache this because it causes issues

cdisc_rules_engine/utilities/reporting_utilities.py
```diff
@@ -7,7 +7,7 @@
 )
 
 
-def get_define_version(dataset_paths: List[str]) -> Optional[str]:
+def get_define_version(dataset_paths: List[str], **kwargs) -> Optional[str]:
```
nhaydel, line 10: same as above

scripts/test_rule.py
```diff
@@ -67,14 +68,14 @@ def validate_single_rule(cache, path, args, datasets, rule: dict = None):
...
-                    dataset.domain,
-                )
+            validated_result = engine.test_validation(
+                rule,
+                os.path.join(directory, dataset.filename),
+                datasets,
+                dataset.domain,
             )
+            results.append(validated_result)
+            engine_logger.info(f"Done validating {dataset.domain}")
```
nhaydel, line 78: I don't think we should log here

nhaydel (changes requested):
Very close, just a few more changes

nhaydel (approved):
Looks good, just need to resolve merge conflicts

nhaydel:
More merge conflicts from your other PR. Once fixed I will merge. @htu

cdisc_rules_engine/rules_engine.py
```diff
@@ -173,6 +177,8 @@ def validate_single_rule(
             error_obj: ValidationErrorContainer = self.handle_validation_exceptions(
                 e, dataset_path, dataset_path
             )
+            if self.is_validate_xml:
```
nhaydel, line 180: This line is breaking validation. We should not set a rule to success if it has an execution error. I believe what we want to do instead is just raise an error during validation if the is_validate_xml is set to true.

nhaydel (changes requested):
There are some breaking changes and I don't know if this actually implements the AC from the ticket because

There is no message in the results indicating that the rule was skipped due to xml schema validation failure

I think we should revisit our approach for this

PR #445, opened 2023-06-21
Populate "Dataset Details" tab in excel export

nhaydel:
This PR adds functionality to populate the dataset details tab in the excel export.

Steps to test:

1. Run the unit tests
2. Run a validation and verify the "Dataset Details" tab is populated with the information for all datasets
3. Run a test validation and verify the "Dataset Details" tab is populated with the appropriate details

### DIFF
--- a/cdisc_rules_engine/services/reporting/excel_report.py
+++ b/cdisc_rules_engine/services/reporting/excel_report.py
@@ -26,6 +26,7 @@ class ExcelReport(BaseReport):
 
     def __init__(
         self,
+        datasets: Iterable[dict],
         dataset_paths: Iterable[str],
         validation_results: List[RuleValidationResult],
         elapsed_time: float,
@@ -33,7 +34,7 @@ class ExcelReport(BaseReport):
         template: Optional[BinaryIO] = None,
     ):
         super().__init__(
-            dataset_paths, validation_results, elapsed_time, args, template
+            datasets, dataset_paths, validation_results, elapsed_time, args, template
         )
         self._item_type = "list"
 
@@ -59,6 +60,23 @@ class ExcelReport(BaseReport):
         )
         wb["Conformance Details"]["B3"] = f"{round(self._elapsed_time, 2)} seconds"
         wb["Conformance Details"]["B4"] = __version__
+
+        # write dataset metadata
+        datasets_data = [
+            [
+                dataset.get("filename"),
+                dataset.get("label"),
+                dataset.get("full_path"),
+                dataset.get("modification_date"),
+                dataset.get("size", 0) / 1000,
+                dataset.get("length"),
+            ]
+            for dataset in self._datasets
+        ]
+        excel_update_worksheet(
+            wb["Dataset Details"], datasets_data, dict(wrap_text=True)
+        )
+
         # write standards details
         wb["Conformance Details"]["B7"] = standard.upper()
         wb["Conformance Details"]["B8"] = f"V{version}"
--- a/tests/unit/test_services/test_reporting/test_excel_export.py
+++ b/tests/unit/test_services/test_reporting/test_excel_export.py
@@ -151,7 +151,7 @@ mock_validation_results = [
 def test_get_rules_report_data():
     with open(test_report_template, "rb") as f:
         report: ExcelReport = ExcelReport(
-            "test", mock_validation_results, 10.1, MagicMock(), f
+            [], "test", mock_validation_results, 10.1, MagicMock(), f
         )
         report_data = report.get_rules_report_data()
         expected_reports = []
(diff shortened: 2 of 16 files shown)


cdisc_rules_engine/services/reporting/excel_report.py
```diff
@@ -59,6 +60,23 @@ def get_export(
...
         wb["Conformance Details"]["B3"] = f"{round(self._elapsed_time, 2)} seconds"
         wb["Conformance Details"]["B4"] = __version__
+
+        # write dataset metadata
+        datasets_data = [
+            [
+                dataset.get("filename"),
+                dataset.get("label"),
+                dataset.get("full_path"),
+                dataset.get("modification_date"),
```
gerrycampion, line 70: I see there is an internal modification date and an external file system modification date. Can we display both of these?

cdisc_rules_engine/services/reporting/excel_report.py
```diff
@@ -59,6 +60,23 @@ def get_export(
...
+                dataset.get("size", 0) / 1000,
```
gerrycampion, line 71: should kb calculation be binary (1024) instead of decimal?

cdisc_rules_engine/services/reporting/excel_report.py
```diff
@@ -59,6 +60,23 @@ def get_export(
         )
         wb["Conformance Details"]["B3"] = f"{round(self._elapsed_time, 2)} seconds"
         wb["Conformance Details"]["B4"] = __version__
+
+        # write dataset metadata
+        datasets_data = [
+            [
+                dataset.get("filename"),
+                dataset.get("label"),
+                dataset.get("full_path"),
```
gerrycampion, line 69: this currently works on windows, but the full path name is not ideal. It looks like this:
`C:/Users/GerryCampion/Code/cdisc-rules-engine-tests/msg\ae.xpt`
If you change this line:
https://github.com/cdisc-org/cdisc-rules-engine/blob/f64c496a1f8e900fd0c274ad9effa295ce1ae427/core.py#LL197C12-L197C12
to this (using Pathlib):
`str(Path(data).joinpath(fn))` 
(or even maintain the Path type throughout and convert it to string later), then I get a path like this in the report on windows:
`C:\Users\GerryCampion\Code\cdisc-rules-engine-tests\msg\ae.xpt`

cdisc_rules_engine/services/reporting/excel_report.py
gerrycampion, line 1: I think we also need the details for the define.xml file in the dataset details tab.

gerrycampion (changes requested):

cdisc_rules_engine/services/reporting/excel_report.py
nhaydel: I will ask @nickdedonder about this tomorrow, since it was not in the AC for this ticket

cdisc_rules_engine/services/reporting/excel_report.py
nhaydel: The template only has one field for modification date, I think displaying both will require a new template, so it should be a new issue, but I will ask @nickdedonder about this tomorrow

cdisc_rules_engine/services/reporting/excel_report.py
nhaydel: I used decimal because I suspect our users will use decimal when converting the file size to kilobytes,  and I didn't want to cause any confusion.

cdisc_rules_engine/services/reporting/excel_report.py
nhaydel: I am curious about how the details for define.xml would fit into the template for dataset details since much of the information does not apply to a define.xml

cdisc_rules_engine/services/reporting/excel_report.py
nhaydel: updated

cdisc_rules_engine/services/reporting/excel_report.py
gerrycampion: i'm thinking the dataset, location, modified time stamp, and size would be populated. The rest would be blank unless Nick wants to overload them with something else.

cdisc_rules_engine/services/reporting/excel_report.py
nhaydel: Per @nickdedonder including internal modification date only.

tests/unit/test_services/test_reporting/test_excel_export.py
```diff
@@ -278,7 +278,7 @@ def test_get_export():
         # Check dataset details tab
         assert wb["Dataset Details"]["A2"].value == "test.xpt"  # filename
         assert wb["Dataset Details"]["B2"].value == "Test Data"  # label
-        assert wb["Dataset Details"]["C2"].value == "tests/unit/text.xpt"  # Location
+        assert wb["Dataset Details"]["C2"].value == "tests/unit"  # Location
```
gerrycampion, line 281: This is failing on windows:
'tests\\unit' == 'tests/unit'

gerrycampion (approved):

PR #446, opened 2023-06-23
Cg0014 fixes

nhaydel:
This PR fixes issues found when trying to write CG0014

Notable changes:

1. Cache feeder now caches variable metadata grouped by dataset/domain instead of having all variables in the same dictionary. This caused the metadata for common variables to be overwritten instead of being domain specific. (e.g EPOCH is required in TA but not in LB)
2. When duplicating conditions for rules, we now operate on a copy of the rule to avoid overwriting the conditions for other concurrent runs.
3. The permissibility operation is now the VariableLibraryMetadata operation. Example usage is included in the attached test files.

Steps to test:

1. Run unit tests
2. Run the rule test command with the attached rule and dataset.
3. Verify 3 errors are shown for the AE dataset where AETERM is missing, and 1 error is shown for the LB dataset where LBTESTCD is missing.
[cg0014-test-data.zip](https://github.com/cdisc-org/cdisc-rules-engine/files/11851378/cg0014-test-data.zip)


cdisc_rules_engine/rules_engine.py
gerrycampion, line 1: If I run a test without "Outcome.Output Variables", i get the following error:

```json
[
  {
    "executionStatus": "execution_error",
    "domain": "AE",
    "variables": [],
    "message": "rule execution error",
    "errors": [
      {
        "error": "An unknown exception has occurred",
        "message": "Output variables: [] not found in dataset"
      }
    ]
  }
]
```


For example:
```yaml
Core:
  Id: CDISC.SDTMIG.CG0014B
  Version: '1'
  Status: Draft
Description: ''
Outcome:
  Message: 'Violation'
  # Output Variables:
  #   - STUDYID
  #   - USUBJID
  #   - $var_perm
Rule Type: Record Data
Sensitivity: Record
Scope:
  Classes:
    Include:
      - ALL
  Domains:
    Include:
      - ALL
Check:
   any:
    - all:
      - operator: variable_metadata_equal_to
        value: Req
        metadata: $var_perm
      - operator: empty
Operations:
     - id: $var_perm
       operator: variable_library_metadata
       name: core
```

gerrycampion, line 1: When we leave the `name` (target) out of the operator check, it appears we don't have access to output the offending variable name and value?

I don't know if this something we can easily add here, a feature we can add in the future, or just a general comment that supports using the dataset builder methodology instead.

gerrycampion, line 1: Because this is a pretty critical update, we should probably run this with a full set of data and rules before and after the update and compare the reports.

gerrycampion (changes requested):

cdisc_rules_engine/rules_engine.py
nhaydel: I've attached the result of my run using main and using this branch. Note this branch has a few more rules on it than due to cache updates:

fix-branch: [CORE-Report-2023-06-26T14-59-56.xlsx](https://github.com/cdisc-org/cdisc-rules-engine/files/11873672/CORE-Report-2023-06-26T14-59-56.xlsx)
main: [CORE-Report-2023-06-26T15-01-34.xlsx](https://github.com/cdisc-org/cdisc-rules-engine/files/11873673/CORE-Report-2023-06-26T15-01-34.xlsx)

So far I have not seen any discrepancies

gerrycampion (approved):

PR #447, opened 2023-06-26
Schema documentation and intellisense

gerrycampion:
Breaks the schema into multiple files.
Add more documentation details to each of the schema components:
- Rule Types
- Check Operations
- Operators
- MetaVariables
Added markdown descriptions and examples (in md files).
Added required parameters.

Also includes small bugfixes for:
- operator metadata parameter
- TestRule API key name
- `test_get_name_referenced_variable_metadata` method name

To test, you can view the dev branch of the rule editor which points to this branch's schema:
https://icy-flower-095494b10-dev.centralus.azurestaticapps.net/

You can view the intellisense when selecting a Rule Type, Check Operation, or MetaVariables.
You can view the descriptions after an item is selected by hovering your cursor after making the selection.

This PR corresponds to this rule editor PR:


resources/schema/Operations.json
```diff
@@ -0,0 +1,338 @@
...
+      ]
+    },
+    "group": {
+      "items": {
+        "$ref": "CORE-base.json#/$defs/VariableName"
+      },
+      "type": "array"
+    },
+    "id": {
+      "pattern": "^\\$.+$",
```
nhaydel, line 309: The engine does not require the id of an operator start with "$". Do we want to require it from the editors?

resources/schema/Operations.md
```diff
@@ -0,0 +1,668 @@
...
+
+## variable_is_null
+
+True if variable is missing or if all values within a variable are null or empty string
+
+## variable_names
+
+Return the set of variable names from the library for the given standard
+
+## variable_permissibilities
```
nhaydel, line 643: This is now the VariableLibraryMetadata. It now requires a name specifying the value to get from the library metadata.

resources/schema/Rule_Type.md
```diff
@@ -0,0 +1,228 @@
...
+- `define_variable_format`
+- `define_variable_allowed_terms`
+- `define_variable_origin_type`
+- `define_variable_is_collected`
+- `define_variable_has_no_data`
+- `define_variable_order_number`
+- `define_variable_has_codelist`
+- `define_variable_codelist_coded_values`
+
+#### Rule Macro
```
nhaydel, line 79: What is the rule macro?

nhaydel (changes requested):
Overall a lot better than what we have. Just some minor updates and we should be good to go

resources/schema/Operations.json
gerrycampion: Updated

resources/schema/Operations.md
gerrycampion: Updated

resources/schema/Rule_Type.md
gerrycampion: This is referring to the "SPECIAL CASES FOR RULE TYPES". I'm not sure what to call this. Rule preprocessing? I'm also not sure if all of these are correct since this part is a bit unclear to me.
 https://github.com/cdisc-org/cdisc-rules-engine/blob/bca332136648d22917c34ec93051c4f731ac0c2c/cdisc_rules_engine/rules_engine.py#L194

nhaydel: I think we can call them rule macros, but I'm not sure if these need to be documented since the users cannot reference this data.

nhaydel (approved):

PR #457, opened 2023-07-08
Issue validation automated scripts

RamilCDISC:
This PR introduces the Automation scripts to handle issues that can be automated. The implementation covers most of the issues, except for those related to the rule editor. The rule editor, being an automation tool itself, falls into a different category and cannot be automated further.


tests/QARegressionTests/test_Issues/test_CoreIssue278.py
nhaydel, line 1: Why was this committed?

PR #458, opened 2023-07-10
ISS-427: Engine detect include special purpose

nhaydel:
This PR adds logic in the engine for handling when an included class is not one of the GENERAL OBSERVATIONS classes. It does this by checking library data and searching for the domain in each class.

Steps to test:

1. Write a rule that operates on a Special Purpose domain
2. Ensure that rule specifies that the SPECIAL PURPOSE class and at least one other GENERAL OBSERVATIONS class is included
3. Run the rule and verify that it runs on the appropriate datasets

### DIFF
--- a/cdisc_rules_engine/services/data_services/base_data_service.py
+++ b/cdisc_rules_engine/services/data_services/base_data_service.py
@@ -24,6 +24,7 @@ from cdisc_rules_engine.services import logger
 from cdisc_rules_engine.services.cdisc_library_service import CDISCLibraryService
 from cdisc_rules_engine.services.data_readers import DataReaderFactory
 from cdisc_rules_engine.utilities.utils import (
+    convert_library_class_name_to_ct_class,
     get_dataset_cache_key_from_path,
     get_directory_path,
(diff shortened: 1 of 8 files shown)


cdisc_rules_engine/services/data_services/base_data_service.py
```diff
@@ -148,7 +156,19 @@ def get_dataset_class(
...
+                cache_key = get_standard_details_cache_key(self.standard, self.version)
+                standard_data = self.cache_service.get(cache_key)
+                if not standard_data:
+                    standard_data = self.cdisc_library_service.get_standard_details(
+                        self.standard, self.version
+                    )
+                class_data, _ = get_class_and_domain_metadata(standard_data, domain)
+                name = class_data.get("name")
+                if name:
+                    return name.replace("-", " ")
```
gerrycampion, line 169: I think this should use `convert_library_class_name_to_ct_class`

gerrycampion (changes requested):

gerrycampion (approved):

PR #464, opened 2023-07-13
ISS-463: Allow dummy data service to get standard and standard_version

nhaydel:
[sd1033-test-data.zip](https://github.com/cdisc-org/cdisc-rules-engine/files/12042663/sd1033-test-data.zip)
Steps to test:

Run a test validation with the provided rule and dataset file. Verify that the provided domains are not skipped

### DIFF
--- a/cdisc_rule_tester/models/rule_tester.py
+++ b/cdisc_rule_tester/models/rule_tester.py
@@ -25,7 +25,12 @@ class RuleTester:
         self.datasets = [DummyDataset(dataset_data) for dataset_data in datasets]
         self.cache = cache or InMemoryCacheService()
         self.data_service = DummyDataService.get_instance(
-            self.cache, ConfigService(), data=self.datasets, define_xml=define_xml
+            self.cache,
+            ConfigService(),
+            standard=standard,
+            standard_version=standard_version,
+            data=self.datasets,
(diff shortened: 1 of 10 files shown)


gerrycampion (changes requested):
Good, except I think we need to return an error with a useful message instead of a scope skip if the library standard and version are missing, right?

Also, do you think it would be a good idea for the rule editor to use the latest standard/version provided in the Rule's Authorities Standards if one is not provided in the test data?

nhaydel:
> Good, except I think we need to return an error with a useful message instead of a scope skip if the library standard and version are missing, right?
> 
> Also, do you think it would be a good idea for the rule editor to use the latest standard/version provided in the Rule's Authorities Standards if one is not provided in the test data?

1. Do you think that message should be thrown when the test endpoint is called? Do we want to require a standard and version are provided
2. I think defaulting to the latest standard/version provided in the authorities section would be cool, but I would ask the rule authors if they would find it useful before implementing it

gerrycampion:
I'm not sure how errors currently work in the Engine, but I think it should be an `execution_error`. Lib standard/version should only be required when an operation/operator/scope etc needs the library.

gerrycampion (approved):

PR #470, opened 2023-07-17
ISS-459: Fix issues reading define 2.0 domain metadata

nhaydel:
Steps to test:

1. Run unit tests


tests/QARegressionTests/test_core/test_define_read.py
RamilCDISC, line 1: Can you please move this test file from test_core to a new folder 'test_issues'  in QAResgressionTests. I will add my other tests there too

tests/unit/test_define_xml_reader.py
```diff
@@ -106,11 +106,14 @@ def test_extract_domain_metadata():
         }
 
 
-def test_extract_variable_metadata():
+@pytest.mark.parametrize(
+    "filename", [(test_define_file_path), (test_define_2_0_file_path)]
```
gerrycampion, line 110: you should also add this parametrization to `test_extract_domain_metadata` to verify class and is_non_standard

gerrycampion (changes requested):
one item, otherwise looks good

tests/QARegressionTests/test_core/test_define_read.py
nhaydel: Actually, I think I am going to remove these tests because this and the unit tests are essentially doing the same thing. Are you ok with that?

PR #482, opened 2023-07-27
Update business rules, fix standard bug for DataServiceFactory

nhaydel:
Steps to test:

Run a test validation with the attached rule and dataset.
Verify that only 2 errors are raised for records 6 and 7
[cg0068.zip](https://github.com/cdisc-org/cdisc-rules-engine/files/12184430/cg0068.zip)


gerrycampion (changes requested):
Currently `present_on_multiple_row_within` in Operator.md says:
> A specific value can be indicated by value

This isn't very helpful, so can you copy the new description of value (comparator) that you added in the BRE

gerrycampion (approved):

PR #496, opened 2023-08-02
Add and update regression tests

RamilCDISC:
This PR is for adding new Regression Tests and updating old ones. I have added a regex match assertion for validate and test command regression tests to catch the soft error which result in rule skip but not in execution of the validation. I have added more regression tests in test_Issues folder and marked them as 'regression': so that they wont run in routine unit tests.

To run regression tests of test_issues you can use following command 
`python -m pytest -m "regression" tests` 
Note that this will run only tests marked as regression. 

To achieve the above functionality through pytest i have added two hooks in conftest.py file.


tests/QARegressionTests/test_Issues/test_CoreIssue145.py
```diff
@@ -0,0 +1,74 @@
+import pandas
```
nhaydel, line 1: I don't think we need to commit tests for the business operators. Those are handled in that repo

tests/QARegressionTests/test_Issues/test_CoreIssue208.py
```diff
@@ -0,0 +1,83 @@
+import json
+import os
+import subprocess
+import unittest
+import pytest
+
+"""This regression test is for automating the validation of acceptancce criteria
```
nhaydel, line 7: Can you include what the expected acceptance criteria is in these tests

tests/QARegressionTests/test_Issues/test_CoreIssue328.py
```diff
@@ -0,0 +1,49 @@
+from cdisc_rules_engine.config.config import ConfigService
+from cdisc_rules_engine.operations.valid_codelist_dates import ValidCodelistDates
+from cdisc_rules_engine.models.operation_params import OperationParams
+import pandas as pd
+from cdisc_rules_engine.services.cache.cache_service_factory import CacheServiceFactory
+from cdisc_rules_engine.constants.cache_constants import PUBLISHED_CT_PACKAGES
+import pytest
+
+"""These tests have been taken from unit tests written by nhyadel.
```
nhaydel, line 9: I don't think we need to duplicate unit tests in regression testing

tests/QARegressionTests/test_Issues/test_CoreIssue344.py
```diff
@@ -0,0 +1,52 @@
+import unittest
+import os
+from cdisc_rules_engine.services.define_xml.define_xml_reader_factory import (
+    DefineXMLReaderFactory,
+)
+import pytest
+
+"""Tests taken from unittests, written by @nhaydel.
```
nhaydel, line 8: same as above

tests/QARegressionTests/test_Issues/test_CoreIssue363.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@pytest.mark.regression
+class TestCoreIssue363(unittest.TestCase):
+    def test_post_request(self):
+        # URL to send the POST request to
+        url = (
+            "https://cdisc-library-conformance-rules-generator-dev.azurewebsites.net"
+            "/api/TestRule?"
+            "code=kWy0PLATQhG0w8hqOFZPGKJ2to-znk9NW7NK-UxgXWb6AzFuClLrmQ=="
```
nhaydel, line 15: This code should not be committed

tests/QARegressionTests/test_core/test_test_command.py
```diff
@@ -33,6 +42,7 @@ def test_test_command_with_all_options(self):
             ],
         )
         self.assertEqual(result.exit_code, 0)
+        self.assertTrue(not self.error_message.search(result.output))
```
nhaydel, line 45: I think we should check for the exact errors we are expecting instead of just checking for errors in general

nhaydel (changes requested):

tests/QARegressionTests/test_Issues/test_CoreIssue145.py
RamilCDISC: removed

tests/QARegressionTests/test_Issues/test_CoreIssue328.py
RamilCDISC: removed

tests/QARegressionTests/test_Issues/test_CoreIssue344.py
RamilCDISC: removed

tests/QARegressionTests/test_Issues/test_CoreIssue363.py
RamilCDISC: changed

tests/QARegressionTests/test_Issues/test_CoreIssue208.py
RamilCDISC: updated

tests/QARegressionTests/test_core/test_test_command.py
nhaydel: I think we want the regression tests to fail if an update causes the output to change. That would signal to the developer that the test needs to be updated

RamilCDISC: updated tests in test_validate_command and test_test_command to include assertions for checking output in the generated report. Also changed the code to use subprocess module for executing terminal commands because the previous implementation was not waiting for full execution of command.

tests/QARegressionTests/test_Issues/test_CoreIssue165.py
```diff
@@ -0,0 +1,75 @@
...
+import pandas as pd
+from business_rules.operators import DataframeType
+import pytest
+
+"""This automated test only validates the acceptance criteria for core issue 165
+for complete test cases coverage, refer to unit tests"""
+
+
+@pytest.mark.regression
+class TestCoreIssue165(unittest.TestCase):
```
nhaydel, line 11: These tests are covered in the unit tests

tests/QARegressionTests/test_Issues/test_CoreIssue119.py
```diff
@@ -0,0 +1,41 @@
...
+import pytest
+
+"""This regression test is for accpetance criteria of Core Issue 119.
+These tests only checks if the is_unique_set and is_not_unique_set functions now
+handle the missing columns in the value list properly by ignoring them or not.
+For other tests please efer to the unit tests"""
+
+
+@pytest.mark.regression
+class TestCoreIssue119(unittest.TestCase):
```
nhaydel, line 13: These tests are covered in unit tests and should be removed from regression tests

nhaydel (changes requested):

tests/QARegressionTests/test_Issues/test_CoreIssue165.py
RamilCDISC: removed

tests/QARegressionTests/test_Issues/test_CoreIssue119.py
RamilCDISC: removed

nhaydel (approved):

PR #500, opened 2023-08-03
ISS-448: Add new rule type to support CG0014

nhaydel:
Steps to test:

1. Run the unit test
2. Run a full validation
3. Run a test validation with the attached rule and dataset and ensure 1 error is reported.
[cg0014-test-data.zip](https://github.com/cdisc-org/cdisc-rules-engine/files/12254691/cg0014-test-data.zip)

### DIFF
--- a/cdisc_rules_engine/dataset_builders/variables_metadata_with_library_metadata.py
+++ b/cdisc_rules_engine/dataset_builders/variables_metadata_with_library_metadata.py
@@ -30,25 +30,20 @@ class VariablesMetadataWithLibraryMetadataDatasetBuilder(BaseDatasetBuilder):
             )
         )
         dataset_contents = self.get_dataset_contents()
-        content_variables_metadata[
-            "has_empty_values"
-        ] = content_variables_metadata.apply(
+        library_variables_metadata = self.get_variables_metadata()
+
+        data = content_variables_metadata.merge(
+            library_variables_metadata,
+            how="outer",
+            left_on="variable_name",
+            right_on="library_variable_name",
+        ).fillna("")
+
+        data["has_empty_values"] = data.apply(
             lambda row: self.variable_has_null_values(
                 row["variable_name"], dataset_contents
             ),
             axis=1,
-        )
-        library_variables_metadata = self.get_variables_metadata()
-
-        data = (
-            content_variables_metadata.merge(
-                library_variables_metadata,
-                how="outer",
-                left_on="variable_name",
-                right_on="library_variable_name",
-            )
-            .fillna("")
-            .drop_duplicates(subset=["variable_name"], keep="last")
         )
         return data
 
--- a/cdisc_rules_engine/dataset_builders/variables_metadata_with_library_metadata.py
+++ b/cdisc_rules_engine/dataset_builders/variables_metadata_with_library_metadata.py
@@ -1,0 +1,83 @@
+from cdisc_rules_engine.dataset_builders.base_dataset_builder import BaseDatasetBuilder
+import pandas as pd
+from typing import List
+from cdisc_rules_engine import config
+import cdisc_rules_engine.utilities.sdtm_utilities as sdtm_utilities
+
+
+class VariablesMetadataWithLibraryMetadataDatasetBuilder(BaseDatasetBuilder):
+    def build(self):
+        """
+        Returns the variable metadata from a given file.
+        Returns a dataframe with the following columns:
+        variable_name
+        variable_order_number
+        variable_label
+        variable_size
+        variable_data_type
+        has_empty_values
+        library_variable_name,
+        library_variable_label,
+        library_variable_data_type,
+        library_variable_role,
+        library_variable_core,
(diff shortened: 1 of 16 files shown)


cdisc_rules_engine/dataset_builders/variables_metadata_with_library_metadata.py
```diff
@@ -0,0 +1,78 @@
...
+        library_variables_metadata = self.get_variables_metadata()
+
+        data = content_variables_metadata.merge(
+            library_variables_metadata,
+            how="outer",
+            left_on="variable_name",
+            right_on="library_variable_name",
+        ).fillna("")
+
+        data["has_empty_values"] = data.apply(
```
gerrycampion, line 42: I think it would be better to call this `variable_has_empty_values` and add it to the data_service's get_variables_metadata so that the other variable metadata builders will have access to it.

cdisc_rules_engine/dataset_builders/variables_metadata_with_library_metadata.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def get_variables_metadata(self) -> pd.DataFrame:
+        # TODO: Update to support other standard types
+        variables: List[dict] = sdtm_utilities.get_variables_metadata_from_standard(
+            standard=self.standard,
+            standard_version=self.standard_version,
+            domain=self.domain,
+            cache=self.cache,
+            config=config,
+            include_model_variables=False,
```
gerrycampion, line 58: I think this should be true. Not required for CG0014, but probably useful for other rules

tests/unit/test_dataset_builders/test_variables_metadata_with_library_metadata_dataset_builder.py
```diff
@@ -0,0 +1,128 @@
...
+@patch(
+    "cdisc_rules_engine.services.data_services.LocalDataService.get_dataset",
+)
+@patch(
+    "cdisc_rules_engine.services.data_services.LocalDataService.get_variables_metadata",
+)
+def test_variable_metadata_with_library_metadata_dataset_builder(
+    mock_get_variables_metadata: MagicMock, mock_get_dataset: MagicMock
+):
+    mock_get_variables_metadata.return_value = pd.DataFrame.from_dict(
```
gerrycampion, line 22: Can you add tests for variables that are in the data, but not in the ig or model.
And if you make the other update, add tests for variables that are in the data and model, but not ig.

gerrycampion (changes requested):

cdisc_rules_engine/dataset_builders/variables_metadata_with_library_metadata.py
nhaydel: I think that kind of defeats the purpose of the metadata reader enhancements we made in the past. We have been trying to avoid reading the entire dataset just to get metadata, would it be possible to wait on adding it to all variable metadata datasets until we know we need it?

gerrycampion: After a searching, it looks like there is a pythonic solution to this using `@property` decorator:
https://stackoverflow.com/a/1598204/2418023

nhaydel: I'm not seeing how that would prevent us from having to read the entire dataset. Are you saying to add a property decorator to the dataset builder itself?

gerrycampion: Sorry, I was confusing the `data` dataframe with an object.
Maybe there's a way we could delay the computation of values within a dataframe column until it is accessed? But if it's not straightforward, then I guess it's best to keep this part in this builder as you've already done.

nhaydel: Since each instance of a dataset builder is targeted to a specific domain, the variable_has_no_data column can be a property on all dataset builders and added to all variable metadata datasets. Does that address your concern? I also think that we will want to abstract all use of dataframes into their own classes anyways so it may make more sense to do it when we perform that abstraction.

gerrycampion: I think so, and I think it makes sense to do it with the abstraction

nhaydel: I ended up not moving the "variable_has_empty_values" to the base_dataset_builder because, at least in the case it is being used for here, we need to perform the check after merging with the library metadata. This is because if a variable appears in the library metadata but not in the dataset we want to the value for "variable_has_empty_values" to be true. I believe this means it is better to calculate that column on the fly, depending on the dataset being evaluated. I realize we could just move the entire function to the base dataset builder, but I figured it would be better to do that if/when we need variable_has_empty_values in other datasets instead of making that change now.

gerrycampion (approved):

PR #508, opened 2023-08-08
ISS-261: Add valid_external_dictionary_value operation

nhaydel:
This PR adds the following:

1. New AbstractDictionaryValidator class this class is meant to be implemented for each supported dictionary type to encapsulate dictionary specific validation logic
2. MedDRADictionaryValidator - Meddra implementation of the AbstractDictionaryValidator
3. Add case_sensitive and external_dictionary_type to operation params
4. Add valid_external_dictionary_value operation

Steps to test:

1. Run the attached rule and dataset using the test meddra dictionary found in `tests/resources/dictionaries/meddra`:
    ex: `python core.py test -s sdtmig -v 3-4 -r meddra_check.json -dp meddra_check_dataset.json --meddra tests/resources/dictionaries/meddra`
[meddra_check.zip](https://github.com/cdisc-org/cdisc-rules-engine/files/12295288/meddra_check.zip)

### DIFF
--- a/cdisc_rules_engine/models/dictionaries/abstract_dictionary_validator.py
+++ b/cdisc_rules_engine/models/dictionaries/abstract_dictionary_validator.py
@@ -4,7 +4,12 @@ from abc import ABC, abstractmethod
 class AbstractDictionaryValidator(ABC):
     @abstractmethod
     def is_valid_term(
-        self, term_dictionary: dict, term: str, variable: str = "", **kwargs
+        self,
+        term_dictionary: dict,
+        term: str,
+        term_type: str = "",
+        variable: str = "",
+        **kwargs
     ) -> bool:
         """
         Method to identify whether a term is valid based on its term type.
@@ -12,9 +17,8 @@ class AbstractDictionaryValidator(ABC):
         Args:
             term_dictionary: The dictionary of available terms.
             term: The dictionary term used
-            variable: The variable corresponding with the component of the
-            dictionary to validate against.
-              - only necessary for hierarchical dictionaries i.e meddra)
+            term_type: The component of the dictionary to validate against.
+            variable: The source variable of the term value.
             kwargs: Additional validator specific variables
 
         Returns:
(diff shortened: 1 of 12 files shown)


cdisc_rules_engine/models/dictionaries/abstract_dictionary_validator.py
```diff
@@ -0,0 +1,39 @@
...
+    def is_valid_term(
+        self, term_dictionary: dict, term: str, variable: str = "", **kwargs
+    ) -> bool:
+        """
+        Method to identify whether a term is valid based on its term type.
+
+        Args:
+            term_dictionary: The dictionary of available terms.
+            term: The dictionary term used
+            variable: The variable corresponding with the component of the
```
gerrycampion, line 15: to make this more generic, it seems like this parameter should correspond to the dictionary's variable (like PT), not the source data's variable (like DECOD).  Both the source variable (--DECOD) and the target dictionary variable (PT) should be specified with the operator in the rule logic.

For example,

```yaml
- operation: valid_external_dictionary_values
  name: --DECOD
  id: $is_valid_decod_value
  external_dictionary_type: meddra
  case_sensitive: false
  dictionary_variable: PT
```

gerrycampion (changes requested):

cdisc_rules_engine/models/dictionaries/abstract_dictionary_validator.py
nhaydel: That makes sense, I will pass the source_variable and the target dictionary variable to the function though so we don't need any special mappings

nhaydel: Updated. I added a dictionary_term_type param to the operation and pass both the source variable ("--DECOD") in this case and the proved dictionary_term_type to the is_valid_term method

resources/schema/Operations.md
```diff
@@ -567,6 +567,29 @@ the operation will return
...
+## valid_external_dictionary_value
+
+Returns true if the target variable contains a valid external dictionary value, otherwise false
+
+Can be case insensitive by setting `case_sensitive` attribute to false. It is true by default.
+
+Input:
+
+```yaml
+- operation: valid_external_dictionary_values
```
gerrycampion, line 579: Missed this before.... should not have an "s" at the end.

gerrycampion (approved):
One small typo

PR #519, opened 2023-08-14
ISS-257: Add define_variable_has_comment attribute to support comment existence check

nhaydel:
Steps to test:

Run unit tests

### DIFF
--- a/cdisc_rules_engine/dataset_builders/define_variables_with_library_metadata.py
+++ b/cdisc_rules_engine/dataset_builders/define_variables_with_library_metadata.py
@@ -1,0 +1,46 @@
+from cdisc_rules_engine.dataset_builders.base_dataset_builder import BaseDatasetBuilder
+import pandas as pd
+
+
+class DefineVariablesWithLibraryMetadataDatasetBuilder(BaseDatasetBuilder):
+    def build(self):
+        """
+        Returns a dataset containing metadata for the variables
+        in the specified domain extracted from the define.xml.
+        Columns available in the dataset are:
+        "define_variable_name",
+        "define_variable_label",
+        "define_variable_data_type",
+        "define_variable_role",
+        "define_variable_size",
+        "define_variable_ccode",
+        "define_variable_format",
+        "define_variable_allowed_terms",
+        "define_variable_origin_type",
+        "define_variable_is_collected",
+        "define_variable_has_no_data",
+        "define_variable_order_number",
+        "define_variable_has_codelist",
+        "define_variable_codelist_coded_values",
+        "define_variable_mandatory",
+        "library_variable_name",
+        "library_variable_label",
+        "library_variable_data_type",
+        "library_variable_role",
+        "library_variable_core",
+        "library_variable_order_number"
+        """
+        # get Define XML metadata for domain and use it as a rule comparator
+        variable_metadata: pd.DataFrame = pd.DataFrame(
+            self.get_define_xml_variables_metadata()
+        )
+        library_variables_metadata = self.get_library_variables_metadata()
+
+        data = variable_metadata.merge(
+            library_variables_metadata,
+            how="outer",
+            left_on="define_variable_name",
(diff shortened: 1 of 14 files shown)


nhaydel:
@gerrycampion since this Issue can be solved multiple ways:
1. Using the expected_variables operation in conjunction with the new define_variable_has_comment attribute
2. Using the result of merging define variable metadata with library variable metadata

And I believe we agreed that we prefer option 2 since it is more robust. I will take this time to implement the new rule type for define variable metadata with library variable metadata. If you agree with this approach I will convert this PR to draft until I have implemented the additional rule type

gerrycampion:
> @gerrycampion since this Issue can be solved multiple ways:
> 
> 1. Using the expected_variables operation in conjunction with the new define_variable_has_comment attribute
> 2. Using the result of merging define variable metadata with library variable metadata
> 
> And I believe we agreed that we prefer option 2 since it is more robust. I will take this time to implement the new rule type for define variable metadata with library variable metadata. If you agree with this approach I will convert this PR to draft until I have implemented the additional rule type

Yes, agreed

nhaydel:
> > @gerrycampion since this Issue can be solved multiple ways:
> > 
> > 1. Using the expected_variables operation in conjunction with the new define_variable_has_comment attribute
> > 2. Using the result of merging define variable metadata with library variable metadata
> > 
> > And I believe we agreed that we prefer option 2 since it is more robust. I will take this time to implement the new rule type for define variable metadata with library variable metadata. If you agree with this approach I will convert this PR to draft until I have implemented the additional rule type
> 
> Yes, agreed

Added new rule type: `Define Item Metadata Check against Library Metadata`

cdisc_rules_engine/dataset_builders/define_variables_with_library_metadata.py
```diff
@@ -0,0 +1,46 @@
...
+        "define_variable_ccode",
+        "define_variable_format",
+        "define_variable_allowed_terms",
+        "define_variable_origin_type",
+        "define_variable_is_collected",
+        "define_variable_has_no_data",
+        "define_variable_order_number",
+        "define_variable_has_codelist",
+        "define_variable_codelist_coded_values",
+        "define_variable_mandatory",
```
gerrycampion, line 25: add "define_variable_has_comment",

resources/schema/MetaVariables.md
```diff
@@ -66,6 +66,10 @@ ItemGroupDef.ItemDef.DataType
 
 ItemGroupDef.ItemDef.CodeListRef exists
 
+## define_variable_has_comment
```
gerrycampion, line 69: Need to also create schema and documentation for `define_vlm_has_comment`

gerrycampion (changes requested):

resources/schema/MetaVariables.md
```diff
@@ -130,6 +134,10 @@ ValueListDef.ItemDef.DataType
 
 ValueListDef.ItemDef.CodeListRef exists
 
+## define_variable_has_comment
```
gerrycampion, line 137: should be define_vlm_has_comment

gerrycampion (approved):

gerrycampion:
For QA, I tried testing with the attached rule and data. I expect LB to return a issue for LBCAT, but instead I am receiving an error:
```json
[
  {
    "executionStatus": "execution_error",
    "domain": "LB",
    "variables": [],
    "message": "rule execution error",
    "errors": [
      {
        "error": "Column not found in data",
        "message": "ordinal"
      }
    ]
  }
]
```
[Rule (3).zip](https://github.com/cdisc-org/cdisc-rules-engine/files/12457691/Rule.3.zip)

The same data is also available in the prod rule editor and sharepoint under SD1149

nhaydel:
@gerrycampion with the latest update we are no longer seeing this issue. I believe it was an issue with how we were passing library metadata to the rule builders

PR #525, opened 2023-08-23
ISS-511: Update in memory cache to use LRU cache

nhaydel:
This PR makes the following notable changes:
1. The in memory cache now uses an LRU cache implementation from the cachetools library
2. Because of the eviction policy, library metadata is now stored outside of the cache for use by the rules engine, data services, rule processor, dataset builders and operations
3. Rules are no longer stored in the cache object

Steps to test:
1. Run a full validation and ensure the output is as expected
2. Run a test validation with a rule that requires library metadata and ensure the results are as expected.
3. Run `func start --python` from the root of the project to start the test endpoint
4. Make a request to the locally running test endpoint using a rule that requires library metadata and ensure the output is correct.

### DIFF
--- a/TestRule/__init__.py
+++ b/TestRule/__init__.py
@@ -79,7 +79,9 @@ def main(req: func.HttpRequest, context: func.Context) -> func.HttpResponse:
             raise KeyError("'datasets' required in request")
         validate_datasets_payload(datasets)
         define_xml = json_data.get("define_xml")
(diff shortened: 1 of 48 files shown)


gerrycampion (approved):

gerrycampion:
I've tested the update using the 3 different given options along with a fourth option of running with the Rule Editor. QA passed.

PR #531, opened 2023-09-04
Update test_rule.py

dostiep:
This PR is to fix an issue when using dataset path for the test fucntion in a Windows environment command line.

### DIFF
--- a/scripts/test_rule.py
+++ b/scripts/test_rule.py
@@ -93,7 +93,7 @@ def validate_single_rule(
                 validated_domains.add(dataset.domain)
                 validated_result = engine.test_validation(
                     rule,
-                    os.path.join(directory, dataset.filename),
+                    os.path.join(directory, dataset.filename).replace(os.sep, '/'),
                     datasets,
                     dataset.domain,
                 )


scripts/test_rule.py
```diff
@@ -93,7 +93,7 @@ def validate_single_rule(
                 validated_domains.add(dataset.domain)
                 validated_result = engine.test_validation(
                     rule,
-                    os.path.join(directory, dataset.filename),
+                    os.path.join(directory, dataset.filename).replace(os.sep, '/'),
```
gerrycampion, line 96: The linter is complaining about this line. See more info about formatting and pre-commits here: https://github.com/cdisc-org/cdisc-rules-engine#code-formatter

gerrycampion, line 96: This fix works, but it defers this OS file-separator issues. It looks like the actual cause of the issue is here:

https://github.com/cdisc-org/cdisc-rules-engine/blob/3c18a77ca164259d20d70f4710b2b094e4a85a76/cdisc_rules_engine/services/data_services/dummy_data_service.py#L51

Instead, it should be something like: 
```python
dataset_name = os.path.basename(dataset_name)
```

gerrycampion (changes requested):

dostiep:
dataset_name = os.path.split(path)[-1]  would probably work

gerrycampion:
> dataset_name = os.path.split(path)[-1] would probably work

Either should work fine. This is sort of what basename does, so basename is just a shortcut

dostiep:
Thanks for the tip :) 

I'll update the proper program instead and will create a new PR along with the updates for Dataset-JSON.

gerrycampion:
@dostiep In the future, you don't need to create a new PR to make updates for the same issue. You can just push the updates to the issue branch and rerequest review and they should show up in the current PR.  It should be a bit easier.

PR #535, opened 2023-09-08
Dataset json

dostiep:
This PR is for adding Dataset-JSON file format as permisible data format. 

XPT remains the default data format but users can use -df or --data-format option with value "json" in validate command call(i.e. core.py validate -s sdtmig -v 3-4 -d Examples/sdtm -dv 2.0 -df json)

Moreover, test command now accepts either Dataset-JSON or JSON dataset format that match the one provided by the rule editor. 

This PR also fixes a minor bug when the test command is called from a Windows environment.


dostiep:
Forgot to mention jsonschema module needs to be installed as well.

nhaydel:
> Forgot to mention jsonschema module needs to be installed as well.

It should be added to the requirements.txt file so the tests pass

dostiep:
I guess I still have some more work to do to pass the tests. Didn't know about those. Will get to it next week.

dostiep:
Closing for now to revisit unit testing

Issue #536, opened 2023-09-11
Abstract dataset representation


nhaydel:
In order to more easily support other dataframe libraries, we need to abstract the dataset representation into its own class.

AC:

* All dataframe operations are handled by a new dataset class
* Tests show the dataset class works with both pandas and dask
* Tests and functions are updated to use the new dataset class

PR #538, opened 2023-09-12
ISS-526: Add dask functionality for rule operators

nhaydel:
This PR:

* Moves rule operators implementation from business rules to the engine. This should allow for easier updates
* Creates a new type DaskDataframeType which utilizes much of the same functionality as the dataframe type.
* Adds tests for both dask and the pandas dataframe types.

Steps to test:
1. Regression test validation functionality both full validation and test validation


nhaydel:
@gerrycampion I don't know why I didn't think of this first, but do you think it would be better to implement the operators for the dataset abstraction instead of implementing them for each dataset library? This would mean we could easily add libraries without having to reimplement the operators. If you are ok with this I will move this work to the new dataset abstraction branch

nhaydel:
We are going to implement 1 DataframeType class that implements all operators on our dataset abstraction. Closing and moving implementation to #536

PR #543, opened 2023-09-14
Relrec merging

gerrycampion:
Allows Match Datasets to use match information provided in a relrec.

Example given in test data and in attached data. In attached data, you can see a sample rule, test data, and result (with one negative result).
[CG0174.zip](https://github.com/cdisc-org/cdisc-rules-engine/files/12610341/CG0174.zip)

- Library model meta is required in order to find variables that should have wildcards.
- This update includes 2 bugfixes for library and caching
  1. The testrule endpoint dataservice was being created without the library metadata object
  2. The model metadata cache key was being extracted from the ig incorrectly (wrong indices)

```yaml
# Variable: FAOBJ
# Condition: Related record present in parent domain dataset
# Rule: FAOBJ in (--TERM, --TRT, --DECOD)
Check:
  all:
    - name: FAOBJ
      operator: not_equal_to
      value: RELREC.__TERM
    - name: FAOBJ
      operator: not_equal_to
      value: RELREC.__TRT
    - name: FAOBJ
      operator: not_equal_to
      value: RELREC.__DECOD
Match Datasets:
  - Name: RELREC
    Wildcard: __
```

### DIFF
--- a/cdisc_rules_engine/utilities/dataset_preprocessor.py
+++ b/cdisc_rules_engine/utilities/dataset_preprocessor.py
@@ -84,6 +84,7 @@ class DatasetPreprocessor:
                 left_dataset_domain_name=self._dataset_domain,
                 right_dataset=other_dataset,
                 right_dataset_domain_details=domain_details,
+                datasets=datasets,
             )
         logger.info(f"Dataset after preprocessing = {result}")
         return result
@@ -102,6 +103,7 @@ class DatasetPreprocessor:
         left_dataset_domain_name: str,
         right_dataset: pd.DataFrame,
         right_dataset_domain_details: dict,
+        datasets: List[dict],
     ) -> pd.DataFrame:
         """
         Merges datasets on their match keys.
@@ -118,7 +120,16 @@ class DatasetPreprocessor:
         )
 
         # merge datasets based on their type
-        if self._rule_processor.is_relationship_dataset(right_dataset_domain_name):
+        if right_dataset_domain_name == "RELREC":
+            result: pd.DataFrame = DataProcessor.merge_relrec_datasets(
+                left_dataset=left_dataset,
+                left_dataset_domain_name=left_dataset_domain_name,
+                relrec_dataset=right_dataset,
+                datasets=datasets,
+                dataset_preprocessor=self,
+                wildcard=right_dataset_domain_details.get("wildcard"),
+            )
+        elif self._rule_processor.is_relationship_dataset(right_dataset_domain_name):
             result: pd.DataFrame = DataProcessor.merge_relationship_datasets(
                 left_dataset=left_dataset,
                 left_dataset_match_keys=left_dataset_match_keys,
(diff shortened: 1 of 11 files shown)


cdisc_rules_engine/utilities/dataset_preprocessor.py
```diff
@@ -118,7 +120,16 @@ def _merge_datasets(
         )
 
         # merge datasets based on their type
-        if self._rule_processor.is_relationship_dataset(right_dataset_domain_name):
+        if right_dataset_domain_name == "RELREC":
```
nhaydel, line 123: Are we properly detecting this domain?

cdisc_rules_engine/utilities/sdtm_utilities.py
```diff
@@ -374,3 +374,25 @@ def get_model_domain_metadata(model_details: dict, domain_name: str) -> dict:
...
 def replace_variable_wildcards(variables_metadata, domain):
     return [var["name"].replace("--", domain) for var in variables_metadata]
+
+
+def get_all_model_wildcard_variables(model_details: dict):
+    return {
+        classVariable["name"]
+        for cls in model_details.get("classes", [])
+        for classVariable in cls.get("classVariables", [])
+        if classVariable["name"].startswith("--")
```
nhaydel, line 384: Should the wildcards come from the rule? Applies to the next function as well

nhaydel (commented):
Overall looks good. I have a few questions though

cdisc_rules_engine/utilities/sdtm_utilities.py
gerrycampion: I don't think so. There are 2 different wildcards we are using:
- The wildcards used in the library model. This wildcard is always `--`, which is what we are using here and below. This means replace the wildcard with the current domain name.
- The wildcards specified by the rule and used by the `add_variables_wildcards` below. It will be the output wildcard in the dataset produced by the relrec matching. In the sample rule, I am using `__`. In the engine rule.py, I am defaulting this to `**` if it is not provided by the rule. This wildcard needs to be different from `--` because if we used the same wildcard, we would end up with something like `FATERM`, which is a variable in the left dataset, but not a variable in the right dataset.

cdisc_rules_engine/utilities/dataset_preprocessor.py
gerrycampion: This is comparing the value in:
```yaml
Match Datasets
- Name:
```
to "RELREC"

What did you have in mind?

I did notice that "domain_name" is not being added to the dataset metadata for datasets that do not have a "DOMAIN" column. Technically these aren't domains and they shouldn't have this populated, but a lot of the code depends on this metadata being populated for these datasets because the dataset name is not being used.  I updated this in the latest commit. In the future, we will want to fix this everywhere so that datasets and domains are distinguishes properly. It will become more apparent in other standards like adam.

nhaydel (approved):

PR #556, opened 2023-09-28
ISS-551: Update check operators to use new dataset abstraction and move operators to engine repo

nhaydel:
This PR:

    Moves rule operators implementation from business rules to the engine. This should allow for easier updates
    Creates a new type DaskDataframeType which utilizes much of the same functionality as the dataframe type.
    Adds tests for both dask and the pandas dataframe types.

Steps to test:

   * Run unit tests


cdisc_rules_engine/check_operators/dataframe_operators.py
nhaydel, line 1: This file was simply copied from the business_rules repo and modified to use the new DatasetInterface

PR #575, opened 2023-10-26
Schema updates for DDF

gerrycampion:
To test, you should be able to run a local instance of the core rule editor along with the following env variable:
"RULE_SCHEMA_URL": "https://raw.githubusercontent.com/cdisc-org/cdisc-rules-engine/1-update-rule-schema-so-it-can-process-ddf-rules/resources/schema/CORE-base.json"

After running, you can test the different rule formats for ddf in the rule editor and should not see yellow/red lines indicating issues with the schema.

### DIFF
--- a/resources/schema/CORE-base.json
+++ b/resources/schema/CORE-base.json
@@ -341,6 +341,7 @@
     "Executability": {
       "enum": [
         "Fully Executable",
+        "Not Executable",
         "Partially Executable",
         "Partially Executable - Possible Overreporting",
         "Partially Executable - Possible Underreporting"
--- a/resources/schema/CORE-base.json
+++ b/resources/schema/CORE-base.json
@@ -119,6 +119,14 @@
       },
       "type": "array"
     },
+    "Strings": {
+      "items": {
+        "minLength": 1,
+        "type": "string"
+      },
+      "minItems": 1,
+      "type": "array"
+    },
     "Subclasses": {
       "items": {
         "enum": [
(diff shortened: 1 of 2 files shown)


ASL-rmarshall (commented):
I have reviewed the changes and the resultant functionality in a local installation of the rules editor.  In general, I am happy that the changes are having the desired/expected effect, though I do have some additional comments:

Addition of Strings allows anything to be added under Classes. As discussed:
  - Ideally, there would be a restriction to only enter valid USDM class names. However, it wouldn't be great to have to maintain that manually. Would sort sort of look-up be possible, or would this introduce an unwanted dependency?
  - If we're not strictly restricting to USDM class names, should we add some sort of regex to restrict to things that look like USDM class names? E.g. no whitespace at a minimum and then there'd be a question of what text casing is acceptable.

Would the change at line 492 (change of `oneOf` to `anyOf` and requiring only `Classes`) have the unintended consequence of making checking of non-DDF rules less rigorous? For example, would this allow an SDTM rule to be specified without `Domains`

There's a question about whether ICH should be defined as a separate Organization for rule specifications.  However, it may be more appropriate to bundle the ICH M11 technical requirements under CDISC/DDF because the ICH M11 specifications don't apply directly to USDM data structures.  How the M11 requirements apply to USDM needs to be determined by the CDISC/DDF rules team, so this could be a justification for retaining all DDF rules under CDISC/DDF (using References/Citations to indicate rules based on ICH M11).

We may need to consider additional changes for `Match Datasets`, but this won't be possible until we have a better idea of how USDM classes might need to be joined for checking.

Would "3" instead of "2" be more appropriate as the required Version for the DDF Standard?  Although the USDM model is currently only at version 2.5, the rules POC is for DDF v3.  I don't know if any minor point version numbers would be required.

ASL-rmarshall (approved):
The changes to Scope and Executability are working as expected.

PR #577, opened 2023-10-28
Changes For Issue 536

RamilCDISC:
This PR add updates for Issue 536. Changes are as follows:
1. All operations have been integrated into one class
2. All operations handle both dask and pandas dataframe properly. 
3. All the unit tests have been updated to include tests for dask datafame too
4. All tests are updated to use new operations class
5. Changes have been made to main code to use services of new operations class 

To test run unit tests


gerrycampion (changes requested):
I haven't looked at the finer details of this PR yet. I've taken a high-level glance, and I think some issues need to be addressed first:
- Instead of the main branch, I believe the branch should be based off of and merged into: https://github.com/cdisc-org/cdisc-rules-engine/tree/ISS-536-datasets-abstraction
- Operations should not all be combined into a single file. They should remain as one file per operation (plus base classes, etc).
- I don't expect to see these "if" statements within the individual operations like this:

```python
if self.library == "pandas":
    data = self.params.dataframe[self.params.target].unique()
elif self.library == "dask":
    data = self.params.dataframe[self.params.target].compute().unique()
```

It should simply look something like:
`data = self.params.evaluation_dataset[self.params.target].unique()`

The implementation-specific details of `unique` across pandas vs dask (`unique()` vs `compute().unique()` should go in the dask and pandas implementations:
https://github.com/cdisc-org/cdisc-rules-engine/blob/ISS-536-datasets-abstraction/cdisc_rules_engine/models/dataset/dask_dataset.py
https://github.com/cdisc-org/cdisc-rules-engine/blob/ISS-536-datasets-abstraction/cdisc_rules_engine/models/dataset/pandas_dataset.py

gerrycampion:
Implemented by: https://github.com/cdisc-org/cdisc-rules-engine/pull/615

PR #608, opened 2023-12-22
Changes for Isssue 417

mdeforestbrown:
This PR uses the 1-based index value for OrderNumber if the OrderNumber does not exist. 

- I've added to an existing test but happy to create a new one if that's better practice here?


gerrycampion:
Closed because a newer PR includes these changes: https://github.com/cdisc-org/cdisc-rules-engine/pull/662

PR #622, opened 2024-02-21
Added datasetjson datatypes

gerrycampion:
decimal datatype was missing from dataset json datatypes

### DIFF
--- a/cdisc_rules_engine/services/datasetjson_metadata_reader.py
+++ b/cdisc_rules_engine/services/datasetjson_metadata_reader.py
@@ -131,9 +131,10 @@ class DatasetJSONMetadataReader:
         rule authors use.
         """
         rule_author_type_map: dict = {
-            "string": "Char",
-            "integer": "Num",
+            "decimal": "Num",
             "float": "Num",
+            "integer": "Num",
+            "string": "Char",
         }
         for key, value in self._metadata_container[
             "variable_name_to_data_type_map"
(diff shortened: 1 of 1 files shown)


cdisc_rules_engine/services/datasetjson_metadata_reader.py
```diff
@@ -131,9 +131,10 @@ def _convert_variable_types(self):
         rule authors use.
         """
         rule_author_type_map: dict = {
-            "string": "Char",
-            "integer": "Num",
+            "decimal": "Num",
             "float": "Num",
+            "integer": "Num",
+            "string": "Char",
         }
```
ASL-rmarshall, line 138: The Dataset-JSON specification describes `type` as:
> Type of the variable. Allowed values: "string", "integer", "decimal", "float", "double", "boolean". See ODM types for details.

If a mapping for "decimal" is being added, wouldn't it also make sense to add mappings for "double" and "boolean"?

ASL-rmarshall (commented):
I'd be happy to approve the addition of the single new mapping for "decimal", but want to confirm whether it would make sense to add mappings for "double" and "boolean" at the same time.

cdisc_rules_engine/services/datasetjson_metadata_reader.py
gerrycampion: Not sure what to map boolean to because I don't think it would exist in SDTM. I mapped it to num because this is what the sas json engine does according to Lex

jehugaleahsa: I had the same question. I'll try to remember to bring it up during a future dataset JSON meeting. I've already seen other people map it to CHAR with Yes/No instead. 😬

Issue #576, opened 2023-10-27
CORERULES-1334 / CG0202: scope skip for RELATIONSHIP class


marisawyckmans:
`rule_processor.is_relationship_dataset` needs to be revisited to ensure it covers special cases like SQ--
https://github.com/cdisc-org/cdisc-rules-engine/blob/3ed356af70051c917780ee54576e7b0e9df0f640/cdisc_rules_engine/utilities/rule_processor.py#L336

`base_data_service.get_dataset_class` needs to be updated to handle more classes, like `RELATIONSHIP`
https://github.com/cdisc-org/cdisc-rules-engine/blob/3ed356af70051c917780ee54576e7b0e9df0f640/cdisc_rules_engine/services/data_services/base_data_service.py#L143

[CORERULES-1334__CG0202_request.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/13189550/CORERULES-1334_CG0202_request.txt)

**Links to related JIRA Tickets**
- https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-1334
- https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-945

**Rule Information**
- **Standard**: SDTMIG
- **Rule ID**: CG0202
- **Rule Description**: When QORIG = ASSIGNED, QEVAL should not be null. This rule applies to SUPP- domains in SPECIAL PURPOSE class

**Describe the bug**
When Classes is specified as SPECIAL PURPOSE, the rule is scope skipping the test data, although test data is for SUPP- domains.
![image](https://github.com/cdisc-org/cdisc-rules-engine/assets/138012940/3159e866-b984-4b0d-b48f-498f7b17d72f)
![image](https://github.com/cdisc-org/cdisc-rules-engine/assets/138012940/d9a19325-00de-4e0e-bccd-e2e70a127b0f)

**Error returned from Rule Engine**
{
  "SUPPAE": [
    {
      "executionStatus": "skipped",
      "domain": "SUPPAE",
      "variables": [],
      "message": null,
      "errors": []
    }
  ],
  "SUPPEC": [
    {
      "executionStatus": "skipped",
      "domain": "SUPPEC",
      "variables": [],
      "message": null,
      "errors": []
    }
  ]
}

**Expected behavior**
The rule should not skip SPECIAL PURPOSE class. Perhaps related to #427 ?

Note: when I change Classes to include ALL, the rule works fine.

JozefAerts:
[...]

Issue #164, opened 2022-10-12
Rule blocked: CG0248, SENDIG-72 - Regular expression causes exception when variable is defined as "num"


JozefAerts:
Note that this issue also impacts sdtm rules like CG0248

Applies to e.g. rule SENDIG-72 "value of --DY must be an integer". 
When trying the regex "^-?[0-9]" on the DSDY variable, which is declared as being "num", an exception occurs ""Can only use .str accessor with string values!".

A solution could of course be that there are functions that test
-whether a value is a valid integer
-whether a value is a valid numeric (e.g. float) value.
[unit-test-coreid-SENDIG_72-negative.xlsx](https://github.com/cdisc-org/cdisc-rules-engine/files/9764048/unit-test-coreid-SENDIG_72-negative.xlsx)

JozefAerts:
Has this ever been fixed?

RamilCDISC:
QA validation part is complete. The Issue is succesfully resolved and rule is no more blocked. The chages were tested against postivie and negative datasets. The positive dataset had all non zero values for the variables DSDY and negative dataset had one string , one zero other than non zero values.
The validation were first conducted on rules editor. We got correct results for both postive and negative datasets. Following image shows output on run against postive dataset

<img width="1306" alt="2" src="https://github.com/cdisc-org/cdisc-rules-engine/assets/113539111/fd7bfe0f-022c-4fb7-ac28-f6c20de71df5">

I have also developed regression tests for the issue. The tests use 'test' command from core to validate the postive and negative dataset against the rule. Then it validates that the correct number of errors, in the correct rows and against correct values were recorded. The engine passes the tests. The following image shows the execution of those tests. 

<img width="1507" alt="1" src="https://github.com/cdisc-org/cdisc-rules-engine/assets/113539111/d130a229-ec77-49d3-9cdb-e1dde60b6929">


I will create a PR for these regression tests along with others once a number of tests are ready.

PR #626, opened 2024-02-27
regression tests for issues 164 and 576

RamilCDISC:
This PR is to push regression tests for issue #164 and #576 .


gerrycampion:
I am still reviewing this, but the first issue I am running into is that these tests have the `@pytest.mark.regression` annotation and these are not being automatically picked up by the test collector.

RamilCDISC:
> I am still reviewing this, but the first issue I am running into is that these tests have the `@pytest.mark.regression` annotation and these are not being automatically picked up by the test collector.

Please run tests marked as regression by passing "-m regression" to the pytest. If your tests are in folder named"tests" your command will be:
`python -m pytest -m regression "path to test folder"`
This will run only tests marked as regression

tests/QARegressionTests/test_Issues/test_CoreIssue164.py
```diff
@@ -0,0 +1,119 @@
+import subprocess
+import os
+import openpyxl
+import pytest
+
+
+@pytest.mark.regression
+def test_negative_dataset():
+    command = (
+        "python3 -m core test -s sdtmig -v 3.4 -r "
```
gerrycampion, line 10: This command breaks when running in a virtual env, because it is using the global instance of python instead of the venv instance of python that has the requirements installed.

Maybe you can do something like this:
https://stackoverflow.com/questions/749711/how-to-get-the-python-exe-location-programmatically

gerrycampion (changes requested):

tests/QARegressionTests/test_Issues/test_CoreIssue164.py
```diff
@@ -0,0 +1,119 @@
+import subprocess
+import os
+import openpyxl
+import pytest
+
+
+@pytest.mark.regression
```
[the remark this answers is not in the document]
RamilCDISC: updated

tests/QARegressionTests/test_Issues/test_CoreIssue164.py
RamilCDISC: updated

tests/QARegressionTests/test_Issues/test_CoreIssue576.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # Delete the file
+    os.remove(file_name)
+
+
+@pytest.mark.regression
+def test_positive_dataset():
+    command = (
+        f"{get_python_executable()} -m core test -s sdtmig -v 3.4 -r "
+        + os.path.join("tests", "resources", "CoreIssue164", "rule.json")
```
gerrycampion, line 79: Did you intend to use the rules and data from 164 to test 576?

tests/QARegressionTests/test_Issues/test_CoreIssue576.py
```diff
@@ -0,0 +1,124 @@
...
+from conftest import get_python_executable
+
+
+@pytest.mark.regression
+def test_negative_dataset():
+    command = (
+        f"{get_python_executable()} -m core test -s sdtmig -v 3.4 -r "
+        + os.path.join("tests", "resources", "CoreIssue576", "rule.json")
+        + " -dp "
+        + os.path.join("tests", "resources", "CoreIssue576", "Negative_Dataset.json")
```
gerrycampion, line 14: I'm not sure the 576 positive and negative test data is appropriate. It seems we should be testing that --SUPP datasets get included  and non-supps don't get included in the scope.

gerrycampion (changes requested):
The changes you made allow me to run successfully now. 
I have 2 new comments. I also created a new ticket for the other test that is failing:
https://github.com/cdisc-org/cdisc-rules-engine/issues/688

tests/QARegressionTests/test_Issues/test_CoreIssue576.py
RamilCDISC: updated

tests/QARegressionTests/test_Issues/test_CoreIssue576.py
RamilCDISC: I think right now we have only these datasets from the resources for unit tests. Maybe we can request a dataset that is according to what you have mentioned in your comment.

RamilCDISC: I have updated the test  and datasets both. The positive dataset have only supp datasets. The test checks if the rule was successfully executed or not along with some other information. The negative dataset has no supp dataset and it checks if in such case the rule is skipped.

tests/QARegressionTests/test_Issues/test_CoreIssue576.py
```diff
@@ -22,48 +31,36 @@ def test_negative_dataset():
...
     # Go to the "Issue Details" sheet
-    sheet = workbook["Issue Details"]
+    Issues_Sheet = workbook["Issue Details"]
+    Dataset_Sheet = workbook["Dataset Details"]
+    Rules_Sheet = workbook["Rules Report"]
 
-    record_column = sheet["F"]
-    variables_column = sheet["H"]
-    values_column = sheet["I"]
+    COREID_Column = Issues_Sheet["A"]
```
gerrycampion, line 42: We typically follow PEP 8 style guide, which says variable names should be lowercase, with words separated by underscores

https://peps.python.org/pep-0008/#function-and-variable-names

tests/QARegressionTests/test_Issues/test_CoreIssue576.py
```diff
@@ -22,48 +31,36 @@ def test_negative_dataset():
...
     file_name = stdout.decode().strip().split(": ")[1] + ".xlsx"
-    print(file_name)
     # Open the Excel file
     workbook = openpyxl.load_workbook(file_name)
 
     # Go to the "Issue Details" sheet
-    sheet = workbook["Issue Details"]
+    Issues_Sheet = workbook["Issue Details"]
+    Dataset_Sheet = workbook["Dataset Details"]
+    Rules_Sheet = workbook["Rules Report"]
```
gerrycampion, line 40: These workbook labels should come from globally-defined constants or enum within the reg tests. Otherwise, if we happen to change a label it will break all of the tests.

tests/QARegressionTests/test_Issues/test_CoreIssue576.py
gerrycampion, line 42: Similar to the comment about using globally-defined constants for the worksheet labels, we should be using a globally-defined enum for the Column indexes.

gerrycampion (changes requested):

tests/QARegressionTests/test_Issues/test_CoreIssue576.py
RamilCDISC: updated

tests/QARegressionTests/test_Issues/test_CoreIssue576.py
RamilCDISC: updated

tests/QARegressionTests/test_Issues/test_CoreIssue576.py
RamilCDISC: updated
now these constants come from globals.py file in the QARegressionTests

gerrycampion (approved):
Looks great. There is a minor merge conflict to resolve

PR #634, opened 2024-03-12
Update the target directory This is mandatory for function apps with python v3.7+ 

hannesSGS:
see: https://learn.microsoft.com/en-us/answers/questions/945144/requirements-file-not-installed-during-python-func

### DIFF
--- a/.github/workflows/rule-tester.yml
+++ b/.github/workflows/rule-tester.yml
@@ -24,7 +24,7 @@ jobs:
         run: |
           python -m venv venv
           source venv/bin/activate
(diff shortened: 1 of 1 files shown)


gerrycampion:
We are currently running on python 3.9, so I'm skeptical that the version is causing the issue.
Before deploying this update, I think we should figure out how to reproduce the issue so that we can find the cause and confirm the fix.

SFJohnson24:
Hi @hannesSGS, we were hoping to gain some more information regarding the issue you were having with the workflow ymls.  Do you have any logs or information for the issue you are fixing with this PR?

PR #639, opened 2024-03-19
Updates fro Issue 345

RamilCDISC:
I have updated the logic in the `concat_split_dataset `function of base data service class to check if split dataset is a supp dataset or not. 
If it is a supp dataset it merges the dataset based on the keys mentioned in the issue description.
if the dataset is not a supp dataset, then it merges dataset  in the same way as it was doing before.

All unit tests pass. This PR  is only for the 1/3 of the whole issue as mentioned in issue description


cdisc_rules_engine/services/data_services/base_data_service.py
```diff
@@ -132,15 +132,31 @@ def concat_split_datasets(
...
-        # concat datasets
-        full_dataset: pd.DataFrame = pd.concat(
-            [dataset for dataset in datasets],
-            ignore_index=True,
-        )
+        full_dataset = pd.DataFrame()
+        for dataset in datasets:
+            if "RDOMAIN" in dataset.columns:
+                supp_merged_dataset = self.merge_supp_dataset(full_dataset, dataset)
+                full_dataset = supp_merged_dataset
```
SFJohnson24, line 139: is supp_merged_dataset variable necessary?  Could we just make full_datase equal to the function call on line 138?

cdisc_rules_engine/services/data_services/base_data_service.py
```diff
@@ -132,15 +132,31 @@ def concat_split_datasets(
...
+        merge_keys = ["STUDYID", "USUBJID", "APID", "POOLID", "SPDEVID"]
+        merged_df = pd.merge(
+            full_dataset,
+            supp_dataset,
+            how="inner",
+            on=merge_keys,
+            left_on="IDVAR",
+            right_on="IDVARVAL",
+        )
+        print(merged_df.head())
```
SFJohnson24, line 157: did you mean to leave this print statement in or is this an artifact from debugging?

SFJohnson24 (changes requested):
CG0019 rule executes with provided dataset.  See comments for some slight refactors to code to be addressed.

cdisc_rules_engine/services/data_services/base_data_service.py
RamilCDISC: They were for debugging. Removed.

SFJohnson24 (approved):
looks great, nice housekeeping on the test suite print statements.  CG0019 still executing as expected with recent refactor.

PR #659, opened 2024-04-09
Filter fix

SFJohnson24:
fixed issue Els described where record count was giving 1 more record than expected.  This was due to the filtered dataframe being overwritten during iteration.


gerrycampion (changes requested):
Can you update the test so that it covers the issue

SFJohnson24:
I updated the multi-filter test case--it fails for the way filter was written as it would only filter on the second key:value in the filter and return 2; if it filters on both parts of the filter, it will return 1 which this fix does now.

gerrycampion (approved):

PR #660, opened 2024-04-10
FDA, Adam, USDM schema updates

gerrycampion:
Refer to connected issues

### DIFF
--- a/resources/schema/Organization_CDISC.json
+++ b/resources/schema/Organization_CDISC.json
@@ -38,7 +38,7 @@
                       "type": "object"
                     },
                     "Version": {
-                      "enum": ["4.0"]
+                      "enum": ["5.0"]
                     }
                   },
                   "type": "object"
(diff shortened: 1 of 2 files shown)


ASL-rmarshall (changes requested):
Apart from the ordering of the renamed "USDM" standard (see inline comment), schema updates have been applied as expected.

However, is it expected that "SEND-AR" is available as a standard for the FDA organization, but not for CDISC?

gerrycampion:
> Apart from the ordering of the renamed "USDM" standard (see inline comment), schema updates have been applied as expected.
> 
> However, is it expected that "SEND-AR" is available as a standard for the FDA organization, but not for CDISC?

CDISC hasn't published conformance rules for SEND-AR AFAIK. See versions tab: https://www.cdisc.org/standards/foundational/send

ASL-rmarshall (approved):
Changes applied as expected.

PR #671, opened 2024-04-12
Prerelease Cache and Version updates

gerrycampion:
Added new github actions to automate prerelease tasks:
- https://github.com/cdisc-org/cdisc-rules-engine/blob/main/.github/workflows/prerelease-update-cache.yml
- https://github.com/cdisc-org/cdisc-rules-engine/blob/main/.github/workflows/prerelease-update-version.yml
  - https://github.com/cdisc-org/cdisc-rules-engine/blob/main/update_version.py

These were already merged to main for testing purposes.
This branch has the results of running the pipelines. It can be rerun and merged before the release.


SFJohnson24 (approved):
Ran validation with report output on clean and issue datasets without issue.

PR #674, opened 2024-04-16
Iss 83 - Run CORE Validate on Local Unpublished rules

SFJohnson24:
added functionality to run unpublished rules with CORE validate CLI with unpublished rule input as yml or json formats.
tested with json, yml rules as well as multiple unpublished rules simultaneously.
updated readme, added unpublished rule validate test along with accompanying unpublished rule YML file in test/resources


README.md
```diff
@@ -103,7 +103,9 @@ Run `python core.py validate --help` to see the list of validation options.
                                   files
   --meddra TEXT                   Path to directory with MedDRA dictionary
                                   files
-  -r, --rules TEXT                Rule core id. ex: CORE-000001. Can be specified multiple times.
+  -r, --rules TEXT                Specify rule core ID ex. CORE-000001. Can be specified multiple times.
+  -u, --unpublished TEXT          Specify absolute path to unpublished rule yml and/or json rule file.
```
gerrycampion, line 107: Can we change this to `-l --local-rules` throughout?
(Ultimately we will want to support unpublished or 3rd party rules via local rules and loading from non-cdisc databases. This update does the first part.)

cdisc_rules_engine/models/rule.py
```diff
@@ -48,27 +48,36 @@ def from_cdisc_metadata(cls, rule_metadata: dict) -> dict:
...
+            ):
+                outcome = rule_metadata.get("Outcome", {})
+                output_variables = cls.get_key(outcome, "Output_Variables")
+                executable_rule["output_variables"] = output_variables
             return executable_rule
         else:
             return rule_metadata
 
+    @classmethod
+    def get_key(cls, data_dict, key):
```
gerrycampion, line 75: I think instead of this, it would be better to take the rule and replace all property name spaces with underscores *recursively* before it is provided to the model. I am worried that some non-root rule properties are accessed elsewhere throughout the engine.

core.py
```diff
@@ -134,7 +134,20 @@ def cli():
...
+    multiple=True,
+    help="specify rule core ID ex. CORE-000001. Can be specified multiple times",
+)
+@click.option(
+    "--unpublished",
+    "-u",
+    required=False,
+    multiple=True,
+    type=click.Path(exists=True, readable=True, resolve_path=True),
+    help="Absolute path to unpublished rule. Can be specified multiple times",
```
gerrycampion, line 149: Is this a directory or file? I think a directory would be more convenient than a file specified multiple times.
If necessary, it can still be used in combination with -r to filter out the specific rules to be targeted.

gerrycampion (changes requested):
Looks great. A few items to address.

Also, since this PR only address part of the ticket (local rules), can you create a new ticket that covers the other (less-important) piece of the ticket:
Ability to load unpublished rules from non-library databases

scripts/script_utils.py
```diff
@@ -113,6 +117,7 @@ def get_cache_service(manager):
         return manager.InMemoryCacheService()
 
 
+# flake8: noqa: C901
```
SFJohnson24, line 120: I added this decorator to ignore Flak8 to get my code commited.  Currently complexity is 13 for get_rules and it has to be below 10 for flake8 to not block commits.  I was interested in your thoughts @gerrycampion on how I could get it down under 10

SFJohnson24:
I made all changes from the prior review and made a ticket for the cache updater.  get_rule as it is currently written is too complex for the flake8 threshold of 10.  Any suggestions on clean ways of reducing this?  I was going to take the logic in the else portion and make it into it's own function to break things up.

scripts/script_utils.py
gerrycampion: Yes, this method is pretty complex and I think it also needs a little work because the args.rules filter should work regardless of the source.
I would reduce the complexity by creating separate methods for at least:
- getting rules from pickle cache
- getting rules from local files

README.md
```diff
@@ -103,7 +103,8 @@ Run `python core.py validate --help` to see the list of validation options.
                                   files
   --meddra TEXT                   Path to directory with MedDRA dictionary
                                   files
-  -r, --rules TEXT                Rule core id. ex: CORE-000001. Can be specified multiple times.
+  -r, --rules TEXT                Specify rule core ID ex. CORE-000001. Can be specified multiple times.
+  -lr, --local_rules TEXT          Specify absolute path to unpublished rule yml and/or json rule file.
```
gerrycampion, line 107: There are still references to "unpublished" here and 2 other places. Should be replaced with "local"

SFJohnson24:
I removed references to unpublished as well as splitting get_rules into helper functions behind filters that should work in cases where -r and -lr are both specified, when just one is specified, or neither is specified.

scripts/script_utils.py
```diff
@@ -114,30 +118,76 @@ def get_cache_service(manager):
 
 
 def get_rules(args) -> List[dict]:
```
gerrycampion, line 120: I think this new logic should be more like:
```python
def get_rules(args) -> List[dict]:
    return (
        load_rules_from_local(args) if args.local_rules else load_rules_from_cache(args)
    )
```

`load_rules_from_cache` will pretty much be the same as `get_rules` was before your updates
`load_rules_from_local` needs logic to handle args.rules (if it does/doesn't exists), args.standard, and args.version

gerrycampion (changes requested):

gerrycampion (approved):

PR #689, opened 2024-04-26
Refactor: Optimized rule applies to class functionality for large datasets

SFJohnson24:
-  added function get_contents_metadata when rule_applies_to_class is called to prevent whole dataset from being read versus just metadata
- tested validate with new functionality on SDTMIG 3.4
- implemented new unittest for added functionality


SFJohnson24:
[validate with update.xlsx](https://github.com/cdisc-org/cdisc-rules-engine/files/15134278/validate.with.update.xlsx)
[validate without update.xlsx](https://github.com/cdisc-org/cdisc-rules-engine/files/15134279/validate.without.update.xlsx)
I performed a validate with both the current changes and pre-changes and the report is identical

cdisc_rules_engine/services/data_services/local_data_service.py
```diff
@@ -94,6 +94,31 @@ def get_dataset_metadata(
         }
         return pandas.DataFrame.from_dict(metadata_to_return)
 
+    @cached_dataset(DatasetTypes.METADATA.value)
+    def get_dataset_contents_metadata(
```
gerrycampion, line 98: Why not use existing [get_variables_metadata](https://github.com/cdisc-org/cdisc-rules-engine/blob/2528e05eaa9ee2f734874d82312635189b2949c3/cdisc_rules_engine/services/data_services/local_data_service.py#L143) instead of creating this new method?

cdisc_rules_engine/utilities/rule_processor.py
```diff
@@ -180,17 +180,25 @@ def rule_applies_to_class(self, rule, file_path, datasets: List[dict], domain: s
...
 
         if excluded_classes:
-            dataset = self.data_service.get_dataset(dataset_name=file_path)
+            dataset = self.data_service.get_dataset_contents_metadata(
+                dataset_name=file_path
+            )
+            dataset_columns = dataset["variable_names"].iloc[0]
+            transformed_columns = pd.DataFrame(columns=dataset_columns)
             class_name = self.data_service.get_dataset_class(
                 dataset, file_path, datasets, domain
```
gerrycampion, line 203: I think you forgot to replace `dataset` with `transformed_columns` here

gerrycampion (changes requested):
I think I made a mistake when writing the ticket. I implied that we only need the metadata to determine the class. I assumed the domain was part of the metadata. We need the domain value for lines like this:
https://github.com/cdisc-org/cdisc-rules-engine/blob/2528e05eaa9ee2f734874d82312635189b2949c3/cdisc_rules_engine/services/data_services/base_data_service.py#L243

While we don't need to read in the entire dataset, we do need to read the first record of the dataset.
I think to fix this, you can pass the variables metadata (from `get_variables_metadata`) and the dataset metadata (from `get_raw_dataset_metadata` - this reads the first record) in `get_dataset_class`.
You will need to modify lines in `base_data_service` where it does `dataset["DOMAIN"].values[0]` to instead look at `dataset_metadata.domain_name`

cdisc_rules_engine/utilities/rule_processor.py
```diff
@@ -179,19 +181,29 @@ def rule_applies_to_class(self, rule, file_path, datasets: List[dict], domain: s
         if included_classes:
             if ALL_KEYWORD in included_classes:
                 return True
-            dataset = self.data_service.get_dataset(dataset_name=file_path)
+            variables = PandasDataset(
+                pd.DataFrame(
+                    columns=self.data_service.get_variables_metadata(
+                        dataset_name=file_path
+                    ).variable_name
```
gerrycampion, line 188: When I run a validation, I am getting an Error: 'PandasDataset' object has no attribute 'variable_name'. Error message: 'PandasDataset' object has no attribute 'variable_name'

I think it is caused by this line or 203

SFJohnson24: @gerrycampion I have corrected this and there is no error now, it is data.variable_names in the new dataset structure.  Sorry.

cdisc_rules_engine/utilities/rule_processor.py
```diff
@@ -179,19 +179,23 @@ def rule_applies_to_class(self, rule, file_path, datasets: List[dict], domain: s
         if included_classes:
             if ALL_KEYWORD in included_classes:
                 return True
-            dataset = self.data_service.get_dataset(dataset_name=file_path)
+            variables = self.data_service.get_variables_metadata(
```
gerrycampion, line 182: This seems to be working well with pandas (33% runtime improvement), but failing with dask. The reason it fails with dask is these lines: https://github.com/cdisc-org/cdisc-rules-engine/blob/2b4bcefaf38866fcce0954f8f4b5464c10da81b3/cdisc_rules_engine/services/data_services/local_data_service.py#L168-L174

This is because the file is originally xpt, but when using dask, we convert the xpt to a parquet file for faster processing. So now when the file is read in, it is .parquet, which does not have an associated metadata reader.

I don't think the parquet files contain the original metadata, so I think you need to somehow pass the original filename to get the metadata, not the parquet filename. (The parquet filename should only be used for fetching the content.)

SFJohnson24: when a dataset gets converted to parquet, the original file path is still in our dataservice.  I decided on just calling read_metadata again with the isolated original filepath instead of adding additional logic to process the filepath.  I dont believe the space or time complexity would be different and the readability is much nicer.  Let me know what you think. 
image

gerrycampion: I was hoping the file_path was already available, however, I don't think this code update is right.
For every call to read_metadata, it is now doing a read_metadata for every other file and it's not returning anything. We need to call read_metadat only for the current file_path's original file_path and we need to return the call to self.metadata.
We need a mapping or pointer from the current file's parquet file path to the xpt file path. Maybe this is already available somewhere, but if not, you might need to create it during the parquet file creation.

gerrycampion (changes requested):
Setting to "request changes" for easier tracking

cdisc_rules_engine/utilities/rule_processor.py
SFJohnson24: @gerrycampion I made some fairly large edits.  When we get to https://github.com/cdisc-org/cdisc-rules-engine/blob/5d4b57a1a08fda57b6d9292cf98507d1df1d3f7f/scripts/run_validation.py#L154 the threads of the multiprocessing Pool unfortunately cannot access the state prior to the pool's creation so a simple map or pointer created before the pool would unfortunately not work.  I tried to use Multiprocessing's Manager to give access to a Datamap class to all the pool threads but, because the class had to be instantiated before the parquet file conversion, it would need to be a dependency and injected down to our data_service (I also had difficulty getting it to function properly).  I also considered redis but configuring and creating it seemed overkill for the utility it needed to provided.  I opted to make a json file we manipulate with the validator.  When I was forcing it to use dask, It appeared to be pulling metadata properly from the datasets.

gerrycampion (commented):
Why can't we just add something like `original_path` as another dataset property here?
https://github.com/cdisc-org/cdisc-rules-engine/blob/d364ac102b8365dffde5c1551aedacdbd66cc89f/scripts/run_validation.py#L147-L150

SFJohnson24:
> Why can't we just add something like `original_path` as another dataset property here?
> 
> https://github.com/cdisc-org/cdisc-rules-engine/blob/d364ac102b8365dffde5c1551aedacdbd66cc89f/scripts/run_validation.py#L147-L150

this worked with a bit of decency injection to get it datasets to the read_metadata call.  I refactored to reflect this change and ran test validations which appear to be working.

gerrycampion (changes requested):
This approach is looking much better. But you are still going to get an error here:
https://github.com/cdisc-org/cdisc-rules-engine/blob/78e2bcaa8619e9bb814c203d046ae207fe9bb97d/cdisc_rules_engine/services/data_services/local_data_service.py#L185
Because it is using the parquet filename instead of the xpt filename to determine which MetadataReader to use.

gerrycampion (changes requested):
Still failing. New error. Here:
https://github.com/cdisc-org/cdisc-rules-engine/blob/90ea95d26caf2fcfee58ea9b6142e62b9d1cb106/cdisc_rules_engine/services/data_services/local_data_service.py#L187

Instead of `file_path`, I think it needs to be `file_metadata["path"]`. `file_path` is still pointing to the parquet path without metadata. `file_metadata["path"]` has the xpt path

SFJohnson24:
> Still failing. New error. Here:
> 
> https://github.com/cdisc-org/cdisc-rules-engine/blob/90ea95d26caf2fcfee58ea9b6142e62b9d1cb106/cdisc_rules_engine/services/data_services/local_data_service.py#L187
> 
> Instead of `file_path`, I think it needs to be `file_metadata["path"]`. `file_path` is still pointing to the parquet path without metadata. `file_metadata["path"]` has the xpt path

I looked to change this but it was extracting metadata properly without the change for some reason (I am using JSON) despite the file_path. pointing to the parquet.  First image is original, second is with file_metadata["path"]  Regardless, I changed it.
image
Screenshot 2024-05-17 104611

PR #693, opened 2024-05-04
Test for merge supp datasets

RamilCDISC:
I have added the test required to close the connected Issue as mentioned by @SFJohnson24

### DIFF
--- a/cdisc_rules_engine/services/data_services/base_data_service.py
+++ b/cdisc_rules_engine/services/data_services/base_data_service.py
@@ -154,9 +154,8 @@ class BaseDataService(DataServiceInterface, ABC):
         else:
             if "IDVAR" in full_dataset:
                 merge_keys.append("IDVAR")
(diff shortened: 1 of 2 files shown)


SFJohnson24:
@RamilCDISC the large dataset PR has caused some conflicts, can you resolve them?

SFJohnson24:
@RamilCDISC the function you created does seem to be performing the merge properly per the unit testing.  I was on your branch and tried to run a debugger validate using EC/SUPPEC as well as DM/SUPPDM datasets from sharepoint.   I put a breakpoint() inside your function (which never triggered) as well as inside some of the validate logic.  It appears engine is not entering the logic to invoke your function when a supp dataset is provided and thus the merge_supp_dataset never gets called.  We may need to check why this is the case.

gerrycampion:
Done by: https://github.com/cdisc-org/cdisc-rules-engine/pull/728

PR #697, opened 2024-05-08
Improved rule loading error logging

SFJohnson24:


scripts/script_utils.py
```diff
@@ -194,16 +207,16 @@ def load_rules_from_local(args) -> List[dict]:
             rule_data[rule_identifier] = rule
             rules.append(rule)
         else:
-            engine_logger.warning(
+            engine_logger.error(
                 f"Rule {rule.get('core_id')} not specified with "
-                "-r rule flag or duplicate rule. Skipping..."
+                "-r rule flag and in local directory or is a duplicate rule. Skipping..."
```
gerrycampion, line 212: I think skipping a rule in the local directory is somewhat typical and should actually be downgraded to `info`. Detecting a duplicate rule should be an error.  These 2 scenarios should be separate. Also, I don't see how this is detecting duplicate rules at all, but maybe I'm misunderstanding.

scripts/script_utils.py
```diff
@@ -136,6 +136,19 @@ def load_rules_from_cache(args) -> List[dict]:
         with open(rules_file, "rb") as f:
             rules_data = pickle.load(f)
             rules = [rules_data.get(key) for key in keys]
+
+        missing_rules = [rule for rule, data in zip(args.rules, rules) if data is None]
+        if missing_rules:
```
gerrycampion, line 141: I don't think this line is necessary

gerrycampion (changes requested):

scripts/script_utils.py
```diff
@@ -136,6 +136,19 @@ def load_rules_from_cache(args) -> List[dict]:
...
+                    f"The rule specified '{missing_rule}' is not"
+                    " in the standard {args.standard} and version {args.version}"
+                )
+        rules = [rule for rule in rules if rule is not None]
+        if not rules:
+            raise ValueError(
+                "All specified rules were excluded because they are not in the standard and version specified."
+            )
     else:
         engine_logger.warning(
```
SFJohnson24, line 153: I believe not specifying a rule and running all rules in the directory also should not be a warning.  I want to be consistent with the level of log.  I was going to change it to info.  Would you agree with the downgrade?

PR #700, opened 2024-05-12
fix bug split dataset-554

ab3263266:
fix bug split dataset-554
Add unit testing

### DIFF
--- a/cdisc_rules_engine/dataset_builders/base_dataset_builder.py
+++ b/cdisc_rules_engine/dataset_builders/base_dataset_builder.py
@@ -54,13 +54,20 @@ class BaseDatasetBuilder:
         """
         pass
 
+    @abstractmethod
+    def build_split_datasets(self) -> DatasetInterface:
+        """
+        Returns correct dataframe to operate on
+        """
+        pass
+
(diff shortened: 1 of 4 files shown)


SFJohnson24:
nice work, @arielBlum1 I noticed 1 things:
 if both of the split datasets have a name that is longer than 4 characters in violation of CG0017, the report only lists the first dataset in the report as an issue and not the second as well.  Currently, it works as expected when only one, either the first or the second, is too long, but not when both are.  And it is working for positive data when neither is too long.

ab3263266:
Hi @SFJohnson24 , nice catch 
Are you sure it should shoot twice, that's a good question, how do we define the split data as one or two data, in the seמetivity of this rule it is defined as Dataset that means it should fire once per dataset, the question is how do you define the split data, I think there is a bit of inconsistency in it, for example here:
image
 We defined as a dataset each file on its own, while under Issue summary and Issue details:
image

  we treated the entire domain as one, I would love to hear your input

SFJohnson24 (approved):
Happy to merge this.

- passes high level review
- demonstrates proper merge within concat_split_dataset
- works as expected for negative and positive test data

PR #701, opened 2024-05-12
Fix JSON reader to Parquet

nhaydel:
This PR fixes the to_parquet method in the json reader class.

It now writes the raw dataset to the tempfile instead of converting to dask first.

Steps to test:
1. Force the engine to use dask datasets. you can do this by setting the `DATASET_SIZE_THRESHOLD` environment variable to 0
2. Run a validation with dataset json. (I used this one for testing:
[TS_neg.json](https://github.com/cdisc-org/cdisc-rules-engine/files/15287474/TS_neg.json)
3. Verify the validation runs completely.

### DIFF
--- a/cdisc_rules_engine/config/config.py
+++ b/cdisc_rules_engine/config/config.py
@@ -13,7 +13,7 @@ class ConfigService(ConfigInterface):
     _config_keys = []
     _instance = None
     # TODO: Make this configurable via env variable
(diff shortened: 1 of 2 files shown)


SFJohnson24 (approved):
Happy to approve this PR.

- PR code passes high level overview
- able to force parquet using new environmental variable in config
- validation for dataset json source running properly

PR #703, opened 2024-05-17
Allow ordering in empty_within_except_last_row

gerrycampion:
Refer to the updated documentation and test cases.
`empty_within_except_last_row` and its complement now allow for an `ordering` property.

business rules were moved to rules engine, so this PR replaces this: https://github.com/cdisc-org/business-rules/pull/67


RamilCDISC (approved):
I reviewed the changes in context of AC of connected issue. The code changes and tests with update to operator.md files seems good.

Issue #499, opened 2023-08-03
Rule blocked: CORERULES-14


gerrycampion:
**Links to related JIRA Tickets**
- https://jira.cdisc.org/browse/CORERULES-14

**Rule Information**
- **Standard**: SDTMIG
- **Rule ID**: CG0028
- **Rule Description**:
Condition: Variable is not TS.TSSEQ
Rule: --SEQ is a unique number per USUBJID per domain, or a unique number per POOLID per domain, including when the domain is split into multiple files

**Describe the bug**
`is_not_unique_set` does not appear to be working as expected for split domains.
When testing CG0028 with positive data, we are getting unexpected negative results for the 2 split domains

[CG0028.zip](https://github.com/cdisc-org/cdisc-rules-engine/files/12252890/CG0028.zip)

ab3263266:
https://github.com/cdisc-org/cdisc-rules-engine/pull/708 PR Created

PR #708, opened 2024-05-26
rename parameters and add function

ab3263266:
fix bug 499 split datasets

 https://github.com/cdisc-org/cdisc-rules-engine/issues/499


RamilCDISC (changes requested):
Could you please add a unit test which checks the bug mentioned in issue 499

RamilCDISC:
Also when I ran a test using your branch,  I noticed some values (for debugging purpose i think so) were logged into the terminal.  Could you please remove it before we merge.

ab3263266:
Hi @RamilCDISC,
I wanted to let you know that the values weren't from my code; I found the prints and deleted them. Additionally, I have added the unit testing and am now waiting for your approval.

Thank you.

RamilCDISC (changes requested):
Thank you Ariel @ab3263266. Everything looks good. I did perform validation with negative and postive datasets. Just one small change before we can go for merging. Could you please move the dataset in the test file to a seprate json file. you can then just load the dataset from the file. This will make test more readable. You can keep the dataset file in resources folder under tests.

ab3263266:
@RamilCDISC Done

SFJohnson24 (approved):
dataset passes high level review, positive and negative data validation works as expected

PR #714, opened 2024-05-30
fix bug-168

ab3263266:
add condition to not return output variables if the operator is "not_exists"


SFJohnson24 (approved):
This code passes high level review.
Confirmed functioning as expected with SD1109, CG0368 for positive and negative data and SEND105, all no longer having column not found in dataset error

PR #716, opened 2024-06-01
Fix list-ct 

nhaydel:
The list ct command was not detecting define-xml ct packages. This PR updates the package file detection logic to detect those files.

Steps to test:

* run the command `python core.py list-ct`
* Verify that define xml ct packages are listed
* Verify that only ct packages are listed


lexjansen (approved):

SFJohnson24:
This code passes review, defineXML CT packages present, no non-CT packages listed

Issue #709, opened 2024-05-28
Rule blocked: CORERULES-CORE-000331


SFJohnson24:
JIRA Tickets - 
870   https://jira.cdisc.org/browse/CORERULES-870  
876  https://jira.cdisc.org/browse/CORERULES-876
 877 https://jira.cdisc.org/browse/CORERULES-877

**Rule Information**
- **Standard**: SDTMIG
- **Rule ID cg0118 / CG0125 / CG0124

**Describe the bug**
There is currently an error with the function  'is_not_contained_by.'  CG0118 uses this operator but CG0124/0125 are published and use this operator and used to execute properly but are now returning an error.

**Error returned from Rule Engine**
{2 items
"TA":[1 item
0:{5 items
"executionStatus":"skipped"
"domain":"TA"
"variables":[]0 items
"message":NULL
"errors":[]0 items
}
]
"DM":[1 item
0:{5 items
"executionStatus":"execution_error"
"domain":"DM"
"variables":[]0 items
"message":"rule execution error"
"errors":[1 item
0:{2 items
"error":"An unknown exception has occurred"
"message":"unhashable type: 'list'"
}
]
}
]
}

**Expected behavior**
Above error is from the positive dataset for CG0125.

**Screenshots**
![image](https://github.com/cdisc-org/cdisc-rules-engine/assets/96841389/42d712fd-061a-4c49-a588-5c85c794caff)
Report:
[CG0125.zip](https://github.com/cdisc-org/cdisc-rules-engine/files/15472674/CG0125.zip)

PR #717, opened 2024-06-04
ISS-709 - Fix contained by operator

nhaydel:
This PR fixes the `get` method implementation on the Dataset Interfaces to resolve the issues found in #709 

Steps to test:
* Test the rule and data found in the issue and verify the results are what you would expect
* Unit tests pass


SFJohnson24 (approved):
see comment above

PR #718, opened 2024-06-06
ISS-711: Sort values implementation

nhaydel:
* Adds a sort values method to existing dataset implementation
* Adds tests for operators that use this method (Note: most test cases were taken from the business rules repo)


SFJohnson24 (approved):
approving this PR--passes high level overview, provides correct results for CG0209 pos and negative data, and added tests run as expected

PR #719, opened 2024-06-07
Iss 690-estimate XML filesize without reading whole dataset

SFJohnson24:
changed XPT metadata reader to no longer read full dataset and instead read only header, 1 row, and end of file to estimate rows.

Works for edge cases --empty dataframe and small row size.  I can provide these datasets to the reviewer as github does not support XPT files and thus I cannot attach them to the PR.


RamilCDISC (changes requested):
The PR looks good. The code implements the acceptance criteria of the connected issue properly. Just one small change. Could you please add an assert check for dataset_lenght too in the test `test_read_metadata` This will ensure the proper integration of other units with the main functionality.

RamilCDISC (approved):
Everything looks good to me. The code integrates well. All testing pass

PR #722, opened 2024-06-11
Xml supp domains

SFJohnson24:
this PR
- addresses naming issues with XML
- refines is_split_dataset using its naming convention so that supp datasets do not trigger split_dataset control flow.  Addressing split merge logic will be addressed in ISS-345
- fixes merge logic using a merge_key that is added to the two dataframes and subsequently removed following merge.


RamilCDISC (changes requested):
I tried running a test using rule editor from the link that was in PR for testing the changes. I used the test data provided in the connected issue. In the test data we have a suppec.xpt which has rdomain and other columns. The rdomain is EC in the data. but if you see the result portion in the attached screenshot, the domain is SUPPEC. After new changes the domain should be reported as EC for this suppec.xpt, but it still shows SUPPEC


Screen Shot 2024-06-19 at 4 33 22 PM

RamilCDISC:
I ran test in rule engine using the test data from the connected issue. I noticed that,
If you see the attached screenshot of the generated report, the dataset name reported in the last column is XS for all, although it should be what is in the first column

RamilCDISC (approved):
The Review for this PR is complete. The updated code successfully implements the AC plus the rule CG0019 is also running successfully. The code was checked against cases such as:

1. Run with positive dataset 
2. Run with negative dataset
3. Run against a dataset having postive and negative both type of data
4. Cross check outcome values to be accurate

The incorrect dataset name being reported is also solved successfully. Refer previous comment for details.

PR #727, opened 2024-06-13
add new argument to get_variables_metadata calls in dataset builders

nhaydel:
Recently a new argument was added to the get_variables_metadata method of the LocalDataService, the dataset builders did not provide this argument. This PR adds the new argument to get_variables_metadata function calls.

Steps to test:

Verify rule CORE-000334 runs correctly on an xpt file

### DIFF
--- a/cdisc_rules_engine/dataset_builders/variables_metadata_dataset_builder.py
+++ b/cdisc_rules_engine/dataset_builders/variables_metadata_dataset_builder.py
@@ -14,5 +14,5 @@ class VariablesMetadataDatasetBuilder(BaseDatasetBuilder):
         variable_format
         """
         return self.data_service.get_variables_metadata(
-            self.dataset_path, drop_duplicates=True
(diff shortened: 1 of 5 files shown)


SFJohnson24:
I am not thrilled about how much we use dataset_name to mean dataset_path throughout the codebase but that convention seems to be fairly widespread.  the code runs as expected for CG0016 with builders arguments fix and previous_operations implementation passing review and is ready for merge.

SFJohnson24 (approved):
see above comment

nhaydel:
> I am not thrilled about how much we use dataset_name to mean dataset_path throughout the codebase but that convention seems to be fairly widespread.  the code runs as expected for CG0016 with builders arguments fix and previous_operations implementation passing review and is ready for merge.

I agree, this naming issue is due to the fact that the engine used to be a cloud based solution and each dataset had a "name" or "id" instead of a path. I think it's worth renaming these variables, but like you mentioned it would take time

PR #728, opened 2024-06-14
Iss 345

SFJohnson24:


cdisc_rules_engine/dataset_builders/base_dataset_builder.py
```diff
@@ -71,6 +72,17 @@ def get_dataset(self, **kwargs):
                 dataset_names=self.get_corresponding_datasets_names(),
                 **kwargs,
             )
+        elif (
```
RamilCDISC, line 75: Could you please mention the reason of specifically filtering on CG0019. Shouldn't we be using a general approach so in future even if ids change or some other rule of same nature comes it still works?

SFJohnson24: I was told we do not want to merge all supp datasets in the way this merge is performed.  I agree filtering on just the rule ID is narrow and not clean, reusable code but I am uncertain of what we do want to fitler into this merge logic on.  I believe @gerrycampion would be better able to advise on what we would want to filter on to control flow into this merge logic but he has been away.  I assumed we would revisit this filter logic later when he had returned.

cdisc_rules_engine/models/dataset/pandas_dataset.py
```diff
@@ -38,6 +38,14 @@ def groups(self):
     def empty(self):
         return self._data.empty
 
+    @property
+    def at(self):
```
RamilCDISC, line 42: I think we should have some logic that will accomodate this in the dask_dataset.py too. In case dask is used to handle the dataset, it may give an error.

cdisc_rules_engine/dataset_builders/base_dataset_builder.py
RamilCDISC: yes, we can revisit it later then

RamilCDISC (approved):
The PR now looks good to me. The code successfully implements the AC. We may have to revisit the logic of using the rule id directly in the code in future. For now as discussed in above comments in PR it is good to go. New code handles the merging of supp datasets using both pandas and dask dataframes.

PR #729, opened 2024-06-18
added None set to empty check and updated tests

SFJohnson24:


RamilCDISC (approved):
The PR successfully implements the first solution sugested in description of connected issue. The addition of considering {None}  also as empty in `empty_operator` integrates well with code. It does not affect previous code and new cases for the connected issue are successfully covered.

PR #730, opened 2024-06-19
Updated target_is_not_sorted_by operator

SFJohnson24:
This update no longer expects the seq to start at 1 and allows for sequences to ascend/descend by values other than 1.

Of note, the testing suite has been updated as this implementation does not have the granularity that the past implementation does--if this is an issue, I can edit again.  In this implementation if the entire sequence is not in the selected ascend/descending order requested in the parameters, all values of the sequence are deemed false.  It appears the past implementation did not do this.


RamilCDISC (approved):
The PR looks good to me. The changes successfully implements the AC of the connected issue. The updatesd unit test also confirms the proper implementation and successful running. I manually ran test too with cases such as

- Chronological order starting with 1
- Chronological order but not starting at 1
- Non Chronological order

The engine passes the tests and PR is good to be merged

PR #732, opened 2024-06-20
Fixed logger for validation command

SFJohnson24:


RamilCDISC (approved):
The pull request successfully solves the issue mentioned in the connected issue. I have validated the chages by running engine with:

1. No log level mentioned explicitly
2. log level as warn
3. log level as error
4. log level as critical
5. log level as info

for each log level only those logs were displayed which were in scipe of specified log level.
Also checked if log level as error is directly set and works fine for 'test' command.

PR #734, opened 2024-06-25
FDA schema fix to sections and cited guidance

SFJohnson24:


RamilCDISC (approved):
The PR successfully implements the AC of connected issue. Schema is updated to accommodate FB prefix and also pattern check for cited guidance is removed for FDA. This completes the AC

PR #737, opened 2024-06-27
further updates to FDAB schema

SFJohnson24:
Rule: FB4001-4005
image
- added 'Section' for FDA documents to resolve schema error in above rules

If you sort in rule editor under FDA, there are many rules that begin with TRC - technical rejection criteria.  
- added 'TRC' for Section-- these IDs will be either TRC with 4 digits or TRC with 4 digits then a letter
- also added TRC into Rule ID regex-- allowing patterns for rule ID that start with TRC- prefix

### DIFF
--- a/resources/schema/Organization_FDA.json
+++ b/resources/schema/Organization_FDA.json
@@ -33,6 +33,10 @@
                           },
                           "Document": {
                             "const": "FDA"
+                          },
+                          "Section": {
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
Implements the AC successfully. The FDA schema is updated to include section key as desired

PR #743, opened 2024-07-11
Split dataset fix

SFJohnson24:


SFJohnson24:
[CORE-Report-2024-07-12T18-35-51.xlsx](https://github.com/user-attachments/files/16233945/CORE-Report-2024-07-12T18-35-51.xlsx)
[supp_and_split.json](https://github.com/user-attachments/files/16233946/supp_and_split.json)
This rule likely need to return to editing.  The current code resolves the issue with is_split_dataset functionality being too specific, the rule clearly has a clause to exclude SUPP-- domain.  Engine and rule editor now are correctly getting the domain off supplementary datasets so SUPP-- Domains no longer exist within engine.  We likely need to add a clause that targets the supp domain's name similar to what was in the old Jira comments https://jira.cdisc.org/browse/CORERULES-9

RamilCDISC:
The code looks great for now and implements the AC perfectly. From the previous comment, we now have two options:

1. We can approve this pull request and merge. While the rule can be updated later.
2. We wait for the rule to be updated and perform some more validation using the updated rule. Then we can approve and merge accordingly

My recommendation would be to go with second option. It will help ensure that we close the issue for once.

RamilCDISC (approved):
The PR implements AC successfully. After the rerolling of function for getting proper domain of SUPP datasets back to reading full domain name, the engine now does not confuse the SUPP datasets as split datasets and successfully excludes them. Also the case for less than 3 characters dataset name which caused rule skipping, is fixed and works fine. The code passes the edge cases too.

PR #745, opened 2024-07-12
update defineXML metadata reader

SFJohnson24:
- added dataset_domain and define_XML_domain to metadata for 'Dataset Metadata Check against Define XML'
- updated defineXMLReader tests to reflect new metadata


RamilCDISC (approved):
The acceptance Criteria for this was reduced to only second point in the AC of connected issue after discussion with Marcelina, as per Sam. The code changes properly implements it. The Define Datast domain is being correctly read for the concerned cases. The unit tests are also updated and supports that the values are correctly being populated and utulised

PR #751, opened 2024-07-17
FileNames operator

SFJohnson24:


RamilCDISC (approved):
The code successfully adds the new operator for dataset names. Also the related documentation file is updated. The unit tests properly checks the working of the operator against both Pandas and Dask dataset types. Also it verifies the edge cases below :

1. missing domains.
2. duplicate domains.

PR #753, opened 2024-07-19
added astype()

SFJohnson24:
As Richard described in the issue, casting the datatype as a bool resolves the issue with empy equal_to series using any or all. 
- For non-empty Series: The apply function works as before, and the result is cast to boolean.
- For empty Series: Instead of returning an empty float64 Series, it returns an empty boolean Series.

I believe this is the most minimal change that doesn't disrupt the operator or involve restructure the existing code.


RamilCDISC (approved):
The changes solves the issue and implements the AC. The issue of appearing of same error when running a test with empty dataset from beginning is resolved too. All test cases pass.

PR #756, opened 2024-07-23
Return error when incorrect rule sensitivity

SFJohnson24:
Rules were published with invalid values for sensitivity.  Currently engine assumes record sensitivity if it is not dataset sensitivity.  This change will create an error fed into the report when there is incorrect sensitivity--of note, this will only work when there are errors for the report being generated.  Presuming a rule will have negative and positive data, if it has incorrect sensitivity, when negative data is tested, the issues will state the rule sensitivity is incorrect.

### DIFF
--- a/cdisc_rules_engine/models/actions.py
+++ b/cdisc_rules_engine/models/actions.py
@@ -111,7 +111,9 @@ class COREActions(BaseActions):
                 lambda df_row: self._create_error_object(df_row, data), axis=1
             )
             errors_list: List[ValidationErrorEntity] = errors_series.tolist()
-        else:  # rule sensitivity is not defined
+        elif (
+            self.rule.get("sensitivity") is not None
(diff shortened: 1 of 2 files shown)


RamilCDISC (approved):
The updated code properly implements the Acceptance Criteria. The engine now only allows for two values of sensitivity i.e Dataset and Record. Any other value specified results in an error mention in the generated report. 

One Remaining part is that the report still shows Successful execution of rule in rule reports tab. We need to make the rule skip. As mentioned by Sam during the discussion on this, the engine mostly checks for sensitivity when it is near the report making part. Also making the rule skip breaks alot of tests. TThe feasible solution for now would be to create a new ticket for solving the rule skipping part. 

One suggestion from Sam is to have a schema check and allow only Dataset and Record as values of Sensitivity

PR #757, opened 2024-07-23
Local Rules Cache Updater

SFJohnson24:
this PR allows the addition of local rules to the cache --they can be yaml or json but most have a key 'custom_id'
list-rules, update-cache, and the validate command for local rules cache have been added (note: a validation with these rules cannot be run [yet--)](https://github.com/cdisc-org/cdisc-rules-engine/issues/761)


RamilCDISC (changes requested):
The changes succsefully implements the AC in the connected issue. Just  a small change before we can close 
1. We need to add a check for custom id while adding them in local rules. All rules should have different custom id
2. We will also have to add a check for when we execute the addition to local cache command. Here we have to must specify tule ids. So if a directory contains more than one rule it should take corresponding number of rule ids. Right now only one works for 2  or more rules too

SFJohnson24:
@RamilCDISC I have added both a duplicate rule check and a duplicate local rule ID check when updating the cache per your review.

RamilCDISC (approved):
The code successfully implements the flags and functionalities for the addition, deletion or lisiting of custom rules in the local cache. The addition deletion of local rules does not affect in any way the cdisc rule cache.  Associated tests have also been updated. After the first review area around custom id has also been tightened so ids can not be repeated.

PR #770, opened 2024-08-01
CG0138: contains_special_topics fix

SFJohnson24:
https://jira.cdisc.org/browse/CORERULES-888?jql=project%20%3D%20CORERULES%20AND%20resolution%20%3D%20Unresolved%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC

There is an error in engine causing the rule to be skipped as the dataset object was not parsed correctly in contains_special_topics.  This was leading engine to think returning None instead of True, leading to domain skipping.  I created a helper function that preserves the way it was checking for the column topics but also checks in dataset.values as these columns may not be contained as keys in the dataframe object.

To test: run CG0138 against test data, the rule should no longer be skipped

### DIFF
--- a/cdisc_rules_engine/services/data_services/base_data_service.py
+++ b/cdisc_rules_engine/services/data_services/base_data_service.py
@@ -295,12 +295,18 @@ class BaseDataService(DataServiceInterface, ABC):
         Checks if the given dataset-class string ends with a particular variable string.
         Returns True/False
         """
-        if "DOMAIN" not in dataset.values and "RDOMAIN" not in dataset.values:
+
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
The changes sucessully solves the bug in engine causing the rule to be skipped as the dataset object was not parsed correctly in contains_special_topics. Valdiation woth the mentoned rule in AC shows succesful running through engine. The rule was also validated with negative cases to cross check if the functionality of engine is not affected for opposite case. The unit test passig shows proper integration of changes in the system

PR #771, opened 2024-08-02
updated JSON and excel reports

SFJohnson24:
This PR brings the JSON report more in line with the excel report.
It also adds the domain of the datasets from issues that arise to give more informative readouts.
new report readouts are below:
[CORE-Report-2024-08-02T15-32-03.xlsx](https://github.com/user-attachments/files/16474628/CORE-Report-2024-08-02T15-32-03.xlsx)
[CORE-Report-2024-08-02T15-34-10.json](https://github.com/user-attachments/files/16474629/CORE-Report-2024-08-02T15-34-10.json)


RamilCDISC (approved):
The JSON report is now fully updated and in coherence with the excel report. Both have same keys in the template. The bug which was found during initial review causing whodrug and meddra version values not being saved in the  json report is also fixed. The variables in code for json and excel report create have been also updated to be coherent. This all fulfills the AC

PR #772, opened 2024-08-06
added defineXML version check for CLI

SFJohnson24:


RamilCDISC (approved):
The restriction of define version using click library ensures that only valid version are mentioned for define. This fulfills the AC. The PR passes the validation.

PR #774, opened 2024-08-07
Update README.md

SFJohnson24:


RamilCDISC (changes requested):
Looking at the AC, i think we should also add in the readme that you have to also provide filename in the directory without file extension. This is what Jozef found out and did not understand from readme

SFJohnson24:
good point, I changed it

RamilCDISC (approved):
Readme is properly updated to provide full understanding of use of flag -o/--output. This fulfills the AC.

PR #775, opened 2024-08-07
added invalid_duration ISO8601 operator

SFJohnson24:
this PR adds the testing, documentation, and functionality for an invalid_duration check that returns a bool based on whether or not a given duration matches ISO 8601 formatting.


SFJohnson24:
I believe the test check is failing as the business rules must be updated first before this PR as the code in this PR makes use of the functionality in that pull request

PR #782, opened 2024-08-12
update exist function

SFJohnson24:
[Rule (8).zip](https://github.com/user-attachments/files/16589915/Rule.8.zip)
this is from before this code was implemented but it contains the define/dataset/rule for this branch.  I do want to make sure this preserves others rules that use exist/not_exist so it would be worth checking another rule or two from this list and ensuring exists functionality is preserved with the changes: [rules_20240806_135751.xlsx](https://github.com/user-attachments/files/16589936/rules_20240806_135751.xlsx)

### DIFF
--- a/cdisc_rules_engine/check_operators/dataframe_operators.py
+++ b/cdisc_rules_engine/check_operators/dataframe_operators.py
@@ -79,19 +79,13 @@ class DataframeType(BaseType):
         target_column = self.replace_prefix(other_value.get("target"))
 
         def check_row(row):
-            for item in row:
-                if isinstance(item, list):
-                    return target_column in item
-                elif isinstance(item, str):
-                    try:
-                        item_list = eval(item)
-                        if isinstance(item_list, list):
-                            return target_column in item_list
-                    except (SyntaxError, ValueError, NameError):
-                        pass
-            return False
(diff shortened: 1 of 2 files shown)


RamilCDISC (changes requested):
The new changes preserves the old logic while adding new functionality of checking within the lists that may appear. This is great. If you would please add a test case in unit tests that verifies the new functionality of checking in the nested lists, we can approve and close.

SFJohnson24:
@RamilCDISC I updated the test suite.  In doing so, I noticed the nested logic was behaving differently than the original function which would return all False or True depending on presence of the target.  I changed the nesting logic to reflect this so we were not getting mixed returns of True/False bools (I believe the presence of True would override the other Falses though in processing which is why I believe this still behaves the same as before.  I tested on positive and negative data for CG0320.)  Would love for you to give it a second glance though but I believe the one line did not change anything except for giving a cleaner result consistent with the way the function worked previously.

RamilCDISC (approved):
The logic is successfully updated to handle the cases, where a column may exist with in the nested lists. Also the test suite is updated to include cases for the new functionality, as requested in the first review. The PR is good to be approved

PR #783, opened 2024-08-13
added regex to schema and properly parse it from rules

SFJohnson24:
This PR has changed to also include updates to the rule parsing logic to allow CG0334 to run.
[Rule_underscores.json](https://github.com/user-attachments/files/16601510/Rule_underscores.json)
This is the regex rule and there are added changes  to engine to correctly parse the new regex argument off the rule and use it for the rule.


RamilCDISC (approved):
The changes  fulfills the AC. The addition in operator.json, allows addition of key regex to be added, making engine not to see it as a violation. Also appropriate code has been added to read the regex when present. Considering these we can approve the PR

PR #785, opened 2024-08-13
Update release.yml

gerrycampion:


RamilCDISC (approved):
The getting of username for PYPI is changed from secrets to vars so that it is easier for devs to know that we are using a token now.

PR #786, opened 2024-08-14
Add loinc support

nhaydel:
[loinc_check.json](https://github.com/user-attachments/files/16606537/loinc_check.json)
[Datasets.json](https://github.com/user-attachments/files/16606538/Datasets.json)
This PR adds loinc support for the rules engine. It also updates the terms factories to return a new ExternalDictionary type. This type contains the version of the dictionary as well as the terms. 

Steps to test:
- [ ] Run the test command with the attached rule and dataset.
- [ ] Verify records 1 and 2 are flagged as errors


RamilCDISC (changes requested):
The code is all good, It works with both positive and negative datasets and oproperly checks for availability of LONIC codes in the specified LOINC dictionaries. One thing could you please update the readme too and mention about the lonic flag. Also I think we need to see if it works with getting version from define.xml as mentioned in the AC. Subsequnetluy we can confirm proper running of CG0400

nhaydel:
> The code is all good, It works with both positive and negative datasets and oproperly checks for availability of LONIC codes in the specified LOINC dictionaries. One thing could you please update the readme too and mention about the lonic flag. Also I think we need to see if it works with getting version from define.xml as mentioned in the AC. Subsequnetluy we can confirm proper running of CG0400

Hey, I can make those updates. What should happen if the version supplied in the define.xml does not match the version supplied when initiating validation? Should the rule be skipped or should all rows be errors?

nhaydel:
> The code is all good, It works with both positive and negative datasets and oproperly checks for availability of LONIC codes in the specified LOINC dictionaries. One thing could you please update the readme too and mention about the lonic flag. Also I think we need to see if it works with getting version from define.xml as mentioned in the AC. Subsequnetluy we can confirm proper running of CG0400

I also don't think the existing operator should pull the version from the define xml, I think that should be handled by a separate operator. Let me know what you think

RamilCDISC (approved):
The support for loinc dictionary was successfully added. This fulfills the half of the AC. The rest part will be covered in a separate ticket (for more detail please refer to comments in the PR). All related files were updated successfully and flags were added to take in lonic dictionary. The readme is also updated for new users. Proper unit tests has been added and execute successfully.

PR #814, opened 2024-08-28
fixed keyword arguments

SFJohnson24:
the decorator for base data service https://github.com/cdisc-org/cdisc-rules-engine/blob/ac7a9f9a6580cc3aeb7c919c38f053e530939daa/cdisc_rules_engine/services/data_services/base_data_service.py#L54  
which wraps the get_dataset_metadata function https://github.com/cdisc-org/cdisc-rules-engine/blob/ac7a9f9a6580cc3aeb7c919c38f053e530939daa/cdisc_rules_engine/services/data_services/local_data_service.py#L83 that is called by this rule expects kwargs.  The named arguments were changed at some point resulting in this to cause an error and skip this rule


RamilCDISC (approved):
The engine skipping CG0017 issue has been solved. The new changes has been thoroughly valdiated to see if any bug remains. Following steps were taken to ensure the validation:
1. Test proper running of rule with 'test' command
2. Test running of the rule with 'validate' command
3. Test with split datasets only
4. Test with SUPP Datasets only (The rule states that these should be excluded, the enigne succesfully excludes them)
5. Test with Dataset having splits datasets including SUPP datasets
6. Test with dataset other than split Datasets 
7. 
For all the above cases positive and negative test data was used. I will write regression test for this and other rules which are getting skipped and upload them in a Single PR

PR #818, opened 2024-08-30
updated has_next_corresponding_record

SFJohnson24:
This rule skips for 2 reasons--one: I believe the rule needs to be edited.  
[Rule_underscores.json](https://github.com/user-attachments/files/16820378/Rule_underscores.json)
"Check": {
    "all": [
      {
        "name": "SEENDTC",
        "operator": "does_not_have_next_corresponding_record",
        "ordering": "SESEQ",
        "value": "SESTDTC",
        "within": "USUBJID"
      }
    ]
we want to order within seq, and not SESTDTC as the rule is currently written.  This results in an error that I describe on the jira ticket.  (see: https://jira.cdisc.org/browse/CORERULES-179)
the second issue is the addition of NON in the old code.  because we cannot compare the last row to the next column, we add NaN to match the number of rows.  When the code tries to perform the not operator on NAN
   def does_not_have_next_corresponding_record(self, other_value: dict):
        return ~self.has_next_corresponding_record(other_value)
we get an error.  I changed the current operator to add True for the last value as, although we do not have a next row to compare for the last row, we can assume it is true.


RamilCDISC (approved):
The code changes successfully solves the bug. The idea of returning true instead of False or Nan for last rows in the compare_target_with_comaprator_next_row operator, ensures that the engine does not raise error due to Nans or False for the last rows, as It is last and would not have a next comparator. This ensure the rule does not skip

PR #820, opened 2024-08-30
cg0105

SFJohnson24:
- this PR allows cg0105 to run both in validation and in test CLI commands.
- this was again, an issues related to differences in test and validation commands.  I have corrected all calls to get_datasets in this PR to use `dataset_name=` for the function arguments as this leads to a bug in the decorator in base data services


RamilCDISC (approved):
The  changes successfully makes the test and validation commands code coherent. The updates ensures that the rule is not skipped because of missing dataset_name= in function argument. All test and validations pass

PR #833, opened 2024-09-16
Actions dependancies

SFJohnson24:
we are getting failed builds during our 'deploy azure rule tester' pipeline.  V2 of the upload artifact has deprecated.
This PR updates the checkout versioning and upload artifact version as well as setup-python versioning for this pipeline


RamilCDISC (approved):
updated accordingly

PR #836, opened 2024-09-19
updated invalid duration logic

SFJohnson24:
I have updated the testsuite in this PR along with the business rules to update the regex and logic for invalid_duration.  
https://regex101.com/r/rnLW9r/1  here is the new regex pattern along with the test cases.  It may need to be updated for DDF if they desire negative durations.
see: https://en.wikipedia.org/wiki/ISO_8601#Durations for a description of duration ISO8601.  Week dates also describes the week logic as weeks cannot be intermingled with other duration indicators (years, months, days).  , or . can be used for decimals but it must be the smallest unit of the duration.  This should fix the errors in the durations @ASL-rmarshall described


RamilCDISC (approved):
The PR successfully implements the required AC. All the positive and negative cases mentioned in the  AC are validated and the engine handles them properly. The case of  allowing negative also works fine. Multiple runs have been made with positive and negative datasets to confirm the proper functioning.

PR #838, opened 2024-09-21
Fix MEDRT install

nhaydel:
Steps to test:
unit tests pass


RamilCDISC (approved):
updated as required. Fixes the minor bug of installing medrt dictionary

PR #843, opened 2024-09-24
Add fillna() method to custom datasets, fix get_model_filtered_variables operator, add shares elements operator suite

SFJohnson24:
this PR fixes bugs with get_model_filtered_variables operator call--mainly the lack of a fillna() method in the custom Dask and Pandas implementation as well as a bug related to library data dictionaries having either ordinal, order_number, or both as properties when the keys were hardcoded into the sort methodology.

shares_exactly_one_element_with, shares_at_least_one_element_with, and shares_no_elements_with were also all missing from the check operators despite being in the reference guide.  This PR adds those operators as well as an associated testsuite.

To test: I used CG0219 to test get_model_filtered_variables as well as shares_no_elements_with.  The other two shares operators do not have any rules to test but could be used in this variants of this rule to test them as well.


RamilCDISC (approved):
The code is successfully updated to support the breaking rules. The engine handles CG0219 and SEND-65 properly. New required operators are being properly tested through unit test. Validation testing has also proved proper functionality of these operators.  The updates carefully implements the changes for all type of datasets (pandas, dask)

PR #849, opened 2024-09-30
Create dependabot.yml

SFJohnson24:
this yml creates the schedule for dependabot and sets the package ecosystem as well as the directory to look for the requirements.txt package manifest.  It also sets out labels for the pull requests it will make when updating packages


RamilCDISC (approved):
Properly Implemented for dependabot.

PR #856, opened 2024-10-02
Tests: added comprehensive rule test suite workflow

SFJohnson24:
I have commented out the automatic aspects of this workflow given the issues with Dask presently and for testing.
This code will likely need to be merged for test and the workflow will need to be manually triggered on a PR to test.  We will also need to set the secrets for a library key, access key for the test_suite before testing.

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -1,20 +1,25 @@
 name: CORE Test Suite Validation
-# will need to set up the secrets for testsuite access token, API_KEY and API_ID for cdisc library
 on:
-  push:
-    branches: [main]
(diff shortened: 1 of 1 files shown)


SFJohnson24:
to test--you will need to make a fork of engine with this test suite.  You will need to set ACCESS_TOKEN and CDISC_LIBRARY_API_KEY as secrets for that repo.  You then go into the actions for the repo and can manually trigger the testsuite job to test it.

RamilCDISC (approved):
The test suite implements the AC and runs successfully. To test I set up fork of the rules engine and ran the workflow manually. The following screenshot shows the successful running of the test suite

PR #863, opened 2024-10-07
Update dependabot.yml

SFJohnson24:
Unfortunately, biweekly is not a frequency for depandabot.  I converted it to monthly so that every other sprint, it will add dependancy tickets and will not be adding more tickets mid-sprint


RamilCDISC (approved):
updated as required

PR #866, opened 2024-10-07
bug: updated local rule parser to fix missing output variables

SFJohnson24:
this PR makes the adjustments recommended by @ASL-rmarshall to confirm the fix.

To test: run a validation with -lr flag using a yaml rule that uses output variables/match datasets.  I used CG0222
[CORE-Report-2024-10-07T16-45-19.xlsx](https://github.com/user-attachments/files/17284349/CORE-Report-2024-10-07T16-45-19.xlsx)


RamilCDISC (approved):
The update has fixed the issue of not mentioning the variables specified in rule in yaml format. The attached report in PR description also shows proper functionality. The tests assert that the updates integrates well

PR #869, opened 2024-10-08
invalid duration updates

SFJohnson24:
this PR addresses the issue from @ASL-rmarshall 
- negative argument has been converted to a bool input
- the schema has been updated
- the operator.md reference has been updated to reflect this change
a seperate PR will be made to address the rule editor .md file

to test, use DDF60

### DIFF
--- a/resources/schema/Operator.json
+++ b/resources/schema/Operator.json
@@ -304,6 +304,9 @@
       "properties": {
         "operator": {
           "const": "invalid_duration"
+        },
+        "negative": {
+          "type": "boolean"
         }
       },
       "required": ["operator", "negative"],
--- a/resources/schema/Operator.json
+++ b/resources/schema/Operator.json
@@ -304,9 +304,6 @@
       "properties": {
         "operator": {
           "const": "invalid_duration"
-        },
-        "negative": {
-          "type": "boolean"
         }
       },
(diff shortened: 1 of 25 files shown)


resources/schema/Operator.json
```diff
@@ -304,6 +304,9 @@
       "properties": {
         "operator": {
           "const": "invalid_duration"
+        },
+        "negative": {
```
ASL-rmarshall, line 308: I think `type` specifications for operator parameters are usually defined in the `properties` section, not within the `anyOf` section that specifies the allowable operations.  Within the `properties` section, the definitions are usually ordered alphabetically, so I'd recommend moving the `type` specification for `negative` to be between the specifications for `name` and `operator` (i.e., somewhere near line 906).

SFJohnson24: Thank you for pointing this out @ASL-rmarshall  I made the changes to the schema and pushed them to remote branch

ASL-rmarshall: The change looks good.

RamilCDISC (approved):
The PR has updated the code with all required changes in the connected issue AC. The problem of internal server error is solved . All related readme files are also updated. The changes integrate well with the system. DDF00060 has been used to validate the proper implementation

PR #873, opened 2024-10-10
added null filter

SFJohnson24:
added a filter that processes errors before the report to replace blanks and None with null.
to test: you can run the certification datasets from Nick.  Here is an example of the before and after.  It also does this for json output.
[CORE-Report-2024-10-10T12-28-09.xlsx](https://github.com/user-attachments/files/17332605/CORE-Report-2024-10-10T12-28-09.xlsx)
[CORE-Report-2024-10-10T13-35-03.xlsx](https://github.com/user-attachments/files/17332606/CORE-Report-2024-10-10T13-35-03.xlsx)

### DIFF
--- a/cdisc_rules_engine/services/reporting/base_report.py
+++ b/cdisc_rules_engine/services/reporting/base_report.py
@@ -123,24 +123,32 @@ class BaseReport(ABC):
                         "row": error.get("row", ""),
                         "SEQ": error.get("SEQ", ""),
                     }
-
+                    values = [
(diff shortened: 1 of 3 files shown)


RamilCDISC (approved):
The code changes implement the AC properly. The negone puts a null for the missing values in the report. Any variable mentioned  in the rule, that is to be reported, and its value is missing i the dataset is represented by null now. Please see attached screenshot for cross reference

PR #874, opened 2024-10-18
updated dy function

SFJohnson24:
this PR addresses the issue with the val operator incorrectly calculating the date difference when datetime is given by simply using just date in the operator.  
To test: use CG006, SEND73 or any other published rule using the val operator
[neg.xlsx](https://github.com/user-attachments/files/17439434/neg.xlsx)
[pos.xlsx](https://github.com/user-attachments/files/17439435/pos.xlsx)
this is using the test data--the results are preserved.  Change the dates to be mixes of datetimes to test


RamilCDISC (approved):
The code changes have successfully implemented the AC. The problem with val operator incorrectly calculating the date difference when datetime is given by simply using just date in the operator, is now resolved. The engine works fine with positive and negative datasets

PR #876, opened 2024-10-22
added Domain to prefix check

SFJohnson24:
The issue stems from domain not being part of dataset metadata but needing to do a dataset name check with the domain.  the check operators not having access to self.params.domain but they do have access to the column_prefix which is set previously in execute_rule() in RulesEngine line 351.  I simply added a filter to grab this prefix, when DOMAIN is specified, which is equal to the domain, thus allowing the check to be performed as desired.

to test: run the rule with the test data


RamilCDISC (approved):
The issue of editor giving wrong result on positive data is soled. The engine properly classifies positive and negative datasets. tests wee performed against positive and negative datasets using both rules and and rules editor. The issue is clear from validation.

PR #878, opened 2024-10-29
Add UNII Support

nhaydel:
This PR does the following:

* Refactors the way external dictionaries are supplied to the engine by introducing a container class. This should help reduce all of the places that need changes when adding a new dictionary.
* Adds UNII models, terms_factory, and validator.
* Ensures reports properly include dictionary versions

Steps to test:
1. Unit tests pass
2. Running a rule that requires dictionary support is successful when the correct dictionary is supplied. Also the provided dictionary version is included in the resulting report.


RamilCDISC (approved):
The PR successfully update the engine to add support for external dictionary UNII and also update the code architecture to provide an easy way of integrating new dictionaries. The changes has been verified by running validation using UNII and previous dictionaries. The updates integrates well and do not cause any bugs. UNII support was validated by writing a rule that verified the UNII code values in a column in the dataset. Following screenshot shows successful runs and no issue found

Issue #837, opened 2024-09-19
CORERULES-9552 - checking values against Library CT for SDTM


SFJohnson24:
This is the first part of the generalized library CT check operator that will crosscheck values against codelist values

1.  The operator will check the rule type, if it is a 'value check againbst define XML' we are SDTM operator,
if it is a 'value check - metadata' we are DDF as this will give us access to the values to compare to codelists.

2. the version of the CT package will be specified in CLI command or found in the defineXML - Codelist - Standard - publishingSet/Version: 
<def:Standards>
      <def:Standard OID="STD.1" Name="STDTMIG" Type="IG" Version="3.3" Status="Final" def:CommentOID="COM.ST1"/>
      <def:Standard OID="STD.2_1" Name="SDTMIG-MD" Type="IG" Version="1.1" Status="Final" def:CommentOID="COM.ST2"/>
      <def:Standard OID="STD.4" Name="CDISC/NCI" Type="CT" PublishingSet="DEFINE-XML" Version="2020-12-18" Status="Final" def:CommentOID="COM.ST4"/>

3. sdtm gets from library differently - uses define and library metadata; ddf uses schema for library.   with SDTM, will need to check if it is non-extensible. if it is extensible, not in list of CT but in defineXML:
<CodeList OID="CL.NVTEST" Name="Nervous System Test" DataType="text" def:StandardOID="STD.3">
        <EnumeratedItem CodedValue="Interpretation" OrderNumber="1" def:ExtendedValue="Yes"/>
        <Alias Context="nci:ExtCodeID" Name="C116103"/>
      </CodeList>

4. we will want to cache codelist values, if there are mutiple codelists will need the union of them

5.   this operator should be able to check both for valid "submission values" and for valid C-codes for the terms in the specified codelist.

see #794 for the DDF implementation and notes from initial ticket

SFJohnson24:
this description for #2 is define 2.1, add that it is only applicable to define 2.1 in operator description

SFJohnson24:
per @eljanssens
As discussed during the Office Hours on Tuesday July 29th, the operation 'get_codelist_attributes' needs to be revised.
<img width="505" height="353" alt="Image" src="https://github.com/user-attachments/assets/f1047760-65de-431c-b081-6628068828b2" />

[...]

PR #879, opened 2024-10-29
New Ruletype: Variable Metadata with Define metadata and library

SFJohnson24:
This PR creates a new Ruletype "Variable Metadata Check against Define XML and Library Metadata" and updates associated documents as well as dataset builder test coverage.


SFJohnson24:
note - cdisc_rules_engine/dataset_builders/variables_metadata_with_define_and_library_dataset_builder.py is a placeholder and will be altered/addressed in ticket #837

RamilCDISC (approved):
The new ruletype is successfully added and implements the requirements as mentioned in the AC. The rule that required comparison for variables from library, define and dataset now runs successfully and engine performs the validation properly. The values for these variables were validated using breakpoints in the code at the point of creation of final dataframe containing all variables. See attached screenshots for reference.

PR #882, opened 2024-10-30
removed PMDA from schema, updated reporting services

SFJohnson24:
I have removed PMDA logic from the schema and reporting as well as added the updated templated from @nickdedonder 
This is what a new validation report looks like: 
[CORE-Report-2024-10-30T14-25-50.xlsx](https://github.com/user-attachments/files/17577404/CORE-Report-2024-10-30T14-25-50.xlsx)


RamilCDISC (approved):
PMDA is successfully removed from engine and report template is updated. Implementing the connected issue AC as required

PR #892, opened 2024-11-06
removed parquet/dask logic with usdm

SFJohnson24:
this PR implements some filters to prevent USDM datasets from entering Dask/Parquet logic per the discussions in CORE Scrum and with Richard regarding technical hurdles--mainly related to the processing/caching overhead that takes place prior to to_parquet() being invoked and the size of USDM datasets anticipated.

To test--run a USDM validation with DATASET_SIZE_THRESHOLD in .env set to 0 to force daskdataset implementation with logger enabled.


RamilCDISC (approved):
The PR successfully implements the AC decided in the scrum meetings. The engine now prevents USDM datasets from entering Dask/Parquet logic. The following screensot shows the before mentioned when a usdm rule is executed

PR #893, opened 2024-11-07
updated codelist_term_map logic

SFJohnson24:
This PR updates the get_codelist_terms_map() logic in the library service to get rid of the allowed terms structure and keep synonyms, submission values and preferred terms as their own separate entities.  
I have run update-cache with this new logic and the new pickle files represent the updated structure.

to test: you can put a breakpoint in the get_codelist_terms_map function and run the update-cache command in the python debugger.
you can also compare the SDTMCT 2020-03-27 picklefile to this https://evs.nci.nih.gov/ftp1/CDISC/SDTM/Archive/SDTM%20Terminology%202020-03-27.pdf to ensure the structure is preserved.

The codelist level data is not returned from library--just the term level so the structure outlined in the ticket was not feasible.


RamilCDISC (approved):
This PR successfully get rid of the allowed terms structure and keep synonyms, submission values and preferred terms as their own separate entities. Also the cache files are updated with new structure. Test has been run to validate things using breakpoints in code logic and python pdb. See attached screenshot for execution

PR #896, opened 2024-11-08
updated pickle cache

SFJohnson24:
after further consideration, I altered the CT codelist cache processing further.
Initially, I had 4 maps -- two at the codelist level and two at the term level with keys of the ID and submission values at each level.  This made the pickle files too large, especially the term level maps.  This was pivoted to a map that goes from either codelist or term level submission value to the codelist and term ID (if appicable).  This will facilitate checks/lookups of both submission values and Id values for both terms and codelists while using far less memory.


RamilCDISC (approved):
updated as required

PR #899, opened 2024-11-14
fix print statement

SFJohnson24:
a print statement was merged into main, leading to many <cdisc_rules_engine.models.rule_validation_result.RuleValidationResult object at 0x00000262E2840EB0> being printed with each validation

### DIFF
--- a/cdisc_rules_engine/services/reporting/base_report.py
+++ b/cdisc_rules_engine/services/reporting/base_report.py
@@ -166,7 +166,6 @@ class BaseReport(ABC):
         """
         rules_report = []
         for validation_result in self._results:
-            print(validation_result)
             rules_item = {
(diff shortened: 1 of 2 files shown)


RamilCDISC:
The issue mentioned in description is solved successfully. The validation has been done using validate command and test command both. But it is observed that for test command the engine prints out the Report Name but does not for validate function. I think it would be better to have it consistent for both

RamilCDISC (approved):
The PR successfully removes the unwanted print statements form the engine. The updated code has been validated using validate command run against xpt and json both datasets. Test command was also run to confirm no unwanted print statements. It was noticed that test command prints the name of the report generated but validate command does not. The code was updated to add report name printing for validation also

Screenshot 2024-11-15 at 5 48 17 PM

PR #918, opened 2024-11-25
Add Support for SNOMED Dictionaries

nhaydel:
Steps to test:
Unit tests
Run the command:
```
python core.py test -s sdtmig -v 3.4 -r rule.json -dp datasets.json --snomed-version 2024-09-01 --snomed-edition SNOMEDCT-US
```
With the rule and datasets provided in the attached zip file. Only one error should be flagged.
[snomed-package.zip](https://github.com/user-attachments/files/17912174/snomed-package.zip)

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,10 @@ Run `python core.py validate --help` to see the list of validation options.
                                   files
   --medrt TEXT                  Path to directory with MEDRT dictionary
                                   files
+  --unii TEXT                  Path to directory with UNII dictionary
+                                  files
+  --snomed-version TEXT        Version of snomed to use. (ex. 2024-09-01)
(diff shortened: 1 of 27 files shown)


nhaydel:
@RamilCDISC lately the snomed api has been down. This may cause issues when testing the rule

RamilCDISC (approved):
The SNOMED dictionary is successfully integrated in the engine. The APi syystem for retrieving values for SNOWMEd works fine. The engine successfully executes the rules against values such as conceptID from snowmed when necessary flags for snowmed are filled. The unit tests passing asserts proper integration. Regression tests passing shows engine functionality is not affected. See the following screenshot which depicts corrct identifying of negative cases by the engine.

PR #919, opened 2024-11-25
updated schema for TIG 1.0

SFJohnson24:
This PR updates the standards and rules schema for the TIG conformance rules per the specs laid out by Els.  I have added the layout of rules she provided me and what these updates are based on
[TIG.authorities(1).xlsx](https://github.com/user-attachments/files/17903802/TIG.authorities.1.xlsx)

### DIFF
--- a/resources/schema/CORE-base.json
+++ b/resources/schema/CORE-base.json
@@ -119,6 +119,21 @@
       },
       "type": "array"
     },
+    "DomainStructure": {
+      "additionalProperties": false,
+      "properties": {
+        "Exclude": {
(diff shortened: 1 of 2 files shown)


RamilCDISC (approved):
The schema is successfully updated to compensate TIG rules. The changes have been validated by running rule editor locally and using multiple TIG rules available. THe following screenshot shows TIG rules being according to standard. There are no red or yellow undermarkings observed. Validation of YAML against schema also passes.   Other standards rules are unaffected by the schema changes.

PR #924, opened 2024-11-27
Enhancement: Controlled terminology and defineXML

SFJohnson24:
This PR does several things;
- control flow for various inputs using -ct and -dxp.  Engine will now extract the controlled terminology from the defineXML if it is 2.1, otherwise it expects the -ct flag to be used if the define is 2.0
- handle edge case (for SDTM) where multiple CT packages are used in a defineXML and compose and cache a merged CT package for use with the codelist operators to use
- define_extensible_codelists operator which returns the extensible values for a list of codelists
- reporting now gets the controlled terminology from the define if it is 2.1

for testing: this rule triggers the new operator  NOTE: you may need to adjust the logic based on the define you are using as they are all different
[extensible_rule.json](https://github.com/user-attachments/files/17983523/extensible_rule.json)
[Datasets.json](https://github.com/user-attachments/files/17983534/Datasets.json)

these are the different defines to check 2.1/2.0 and 2.1 with multiple CT files:
[Definefiles.zip](https://github.com/user-attachments/files/17983530/Definefiles.zip)

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,9 @@ Run `python core.py validate --help` to see the list of validation options.
   -ct, --controlled-terminology-package TEXT
                                   Controlled terminology package to validate
                                   against, can provide more than one
+                                  NOTE: if a defineXML is provided, if it is version 2.1
+                                  engine will use the CT laid out in the define.  If it is
+                                  version 2.0, -ct is expected to specify the CT package
   -o, --output TEXT               Report output file destination and name. Path will be
                                   relative to the validation execution directory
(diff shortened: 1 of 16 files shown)


RamilCDISC (approved):
This PR is a part of a bigger ticket. Current changes being pushed layout base for further changes and completing the AC for the bigger ticker. The code has been carefully validated and all unit + regression tests pass. The changes are in accordance with the AC. For full valdiation a rule, dataset and all version of define xml's will be required which will be done when final PR will be created.

PR #929, opened 2024-12-02
added optional letter in ruleID

SFJohnson24:
from an email from Els:  For rule ID a tweak is needed because the TIG v1.0 has a conformance rule with ID TIG0016a and this one doesn't get recognized by the scheme.

this schema update allows for the addition lowerecase letter after the 4 digits


RamilCDISC (approved):
updated as required

PR #937, opened 2024-12-09
TIG 1.0 Caching, Rule Editor

SFJohnson24:
This PR does several things:
- updates the caching methods to not add the TIG for all substandards (CDASH/SDTM/ADAM/SEND) to the local cache under standard details--this includes the codelists/variable metadata/standard details for each of these API pathways.
- allows for rule editor functionality to test TIG rules.  A schema update is also included in this PR to add a substandard property under Authorities.  Engine will now use this when populating the cache in editor to allow for rule execution.  RELREF domain has also been added to the schema.
- adds a substandard command to both the test and validate CLI commands which allows for the selection of the TIG substandards within the cache

here are the rules to test TIG
[Rule_underscores.json](https://github.com/user-attachments/files/18068433/Rule_underscores.json)
[Datasets.json](https://github.com/user-attachments/files/18068434/Datasets.json)

- rule editor with engine run locally should be done to test/review RuleTester functionality
- update cache methods should be tested/reviewed as they now grab TIG from library endpoints and store the standards in standard_details.pkl


RamilCDISC (approved):
The PR successfully updates the support for tig rules. The caching for engine is updated and successfully gets the TIG standard also form the endpoint. This has been validated after updating the engine cache and checking the availability of the TIG standard in standards_details.pkl file. Also the rule editor works fine and does not run into any server error for TIG. The changes has been validated running the engine locally.

PR #942, opened 2024-12-10
tig schema update

SFJohnson24:
this ticket adds the substandard for TIG rules as well as adding RELREF as a domain


RamilCDISC (approved):
updated as required

PR #944, opened 2024-12-10
is_consistent_across_study

SFJohnson24:
this PR adds is_consistent_across_study operator that takes a target and makes sure it is consistent across values

to test:
run test using operator:
[neg.json](https://github.com/user-attachments/files/18085567/neg.json)
[Rule_underscores.json](https://github.com/user-attachments/files/18085569/Rule_underscores.json)
[send pos.xlsx](https://github.com/user-attachments/files/18085570/send.pos.xlsx)
[send neg.xlsx](https://github.com/user-attachments/files/18085572/send.neg.xlsx)
[pos.json](https://github.com/user-attachments/files/18085573/Datasets.json)

### DIFF
--- a/cdisc_rules_engine/check_operators/dataframe_operators.py
+++ b/cdisc_rules_engine/check_operators/dataframe_operators.py
@@ -949,6 +949,29 @@ class DataframeType(BaseType):
 
     @log_operator_execution
     @type_operator(FIELD_DATAFRAME)
+    def is_consistent_across_study(self, other_value):
+        target = self.replace_prefix(other_value.get("target"))
+        comparator = other_value.get("comparator")
+        grouping_cols = []
+        if isinstance(comparator, str):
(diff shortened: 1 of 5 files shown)


RamilCDISC:
I ran the test for the operator. the operator accurately points out any inconsistencies if found in the dataset. But what I have noticed is that even if I run the test function in the engine using standard as sdtmig (-s sdtmig) the engine still runs for this sending rule. I think this needs to be resolved on priority

sdtmig

RamilCDISC (approved):
The new operator is successfully added and works fine. The operator is integrated properly and does not break any other part of the engine. Unit and regression test passing vouch for that. It was observed that the engine was executing sendig rule even though in the executing command standard was specified as sdtmig . Sam explained that in test command until the rule specifically check the standard and version, the execution of other standards is normal.

PR #946, opened 2024-12-16
updated rule name

SFJohnson24:
updated operator name per recap


RamilCDISC (approved):
Updated accordingly, operator name updated as required

PR #947, opened 2024-12-17
updated cache and QA regression

SFJohnson24:
In order to release--the cache will be updated causing the QA regression to fail.  This PR removes the opinions of the regression test (the checks for a blank issues tab in the report)


RamilCDISC (approved):
Updated as required.

PR #955, opened 2024-12-18
added custom skip logging

SFJohnson24:


RamilCDISC (approved):
The PR updates the logging to not interfere with validation when there is a column missing in the dataset The issue is now resolved., The engine now does not put out any errors when a variable is missing but skips validation as required in AC.

PR #956, opened 2024-12-18
business rule version

SFJohnson24:
incremented version to match updated business rules.

to test these PRs:
take any of the rules from invalid_duration operator DDF0061/DDF0062 and make one of the values being checked into (i.e., NULL, None, NaN, etc.) to test


RamilCDISC (approved):
updated as required. The tests are also updated with required test cases and all tests pass

PR #963, opened 2025-01-03
added test_release action

SFJohnson24:
created test workflow for release that will build binary
tests the --help command on the build
make the releases available as downloadable artifacts in the actions tab
won't create actual GitHub releases or deploy to PyPI

this PR also has a potential hotfit ```--hidden-import=pyreadstat._readstat_writer``` for the missing module which we can test once it is merged and manually triggered

### DIFF
--- a/.github/workflows/build-binary.yml
+++ b/.github/workflows/build-binary.yml
@@ -1,0 +1,50 @@
+name: Build Binary
+
+on:
+  workflow_call:
+    inputs:
+      os:
+        required: true
+        type: string
+      name:
+        required: true
+        type: string
+jobs:
+  build:
+    runs-on: ${{ inputs.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        id: pysetup
+        with:
+          python-version: "3.9"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel twine
+          pip install -r requirements.txt
+      - name: Build Binary
+        run: |
+          if [ "${{ runner.os }}" = "Windows" ]; then
+            pyinstaller core.py --dist ./dist/output/${{ inputs.name }} --hidden-import=pyreadstat._readstat_writer --add-data="$env:pythonLocation\Lib\site-packages\xmlschema\schemas;xmlschema/schemas" --add-data="resources/cache;resources/cache" --add-data="resources/templates;resources/templates" --add-data="resources/schema;resources/schema"
+          else
+            pyinstaller core.py --dist ./dist/output/${{ inputs.name }} --hidden-import=pyreadstat._readstat_writer --add-data=$pythonLocation/lib/python3.9/site-packages/xmlschema/schemas:xmlschema/schemas --add-data=resources/cache:resources/cache --add-data=resources/templates:resources/templates --add-data=resources/schema:resources/schema
+          fi
+        shell: bash
+      - name: Test Binary
+        run: |
+          cd dist/output/${{ inputs.name }}/core
+          if [ "${{ runner.os }}" = "Windows" ]; then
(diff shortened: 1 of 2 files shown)


gerrycampion (changes requested):
Using a [matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow) would be nice here, specifically the code sample that has: `includes_only:`. For example:

```yaml
jobs:
  includes_only:
    strategy:
      matrix:
        include:
          - os: "ubuntu-latest"
            name: "core-ubuntu-latest"
          - os: "ubuntu-20.04"
            name: "core-ubuntu-20-04"
          ...
```
The release workflow would also get updated to use this matrix strategy.

If you plan to reuse the build code in the release, it will also be good to put the reusable steps in a [reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#creating-a-reusable-workflow)

There is an example of a matrix like this and reusable workflows here: https://github.com/cdisc-org/cdisc-library-src-files/tree/master/.github/workflows

SFJohnson24:
I made a matrix implementation that I believe can be reused for both the test_release and the release.yml (assuming we add some additional logic for the release and pypi deployment in the latter's yml file)

.github/workflows/build-binary.yml
```diff
@@ -0,0 +1,50 @@
...
+            pyinstaller core.py --dist ./dist/output/${{ inputs.name }} --hidden-import=pyreadstat._readstat_writer --add-data="$env:pythonLocation\Lib\site-packages\xmlschema\schemas;xmlschema/schemas" --add-data="resources/cache;resources/cache" --add-data="resources/templates;resources/templates" --add-data="resources/schema;resources/schema"
+          else
+            pyinstaller core.py --dist ./dist/output/${{ inputs.name }} --hidden-import=pyreadstat._readstat_writer --add-data=$pythonLocation/lib/python3.9/site-packages/xmlschema/schemas:xmlschema/schemas --add-data=resources/cache:resources/cache --add-data=resources/templates:resources/templates --add-data=resources/schema:resources/schema
+          fi
+        shell: bash
+      - name: Test Binary
+        run: |
+          cd dist/output/${{ inputs.name }}/core
+          if [ "${{ runner.os }}" = "Windows" ]; then
+            if ./core.exe --help; then echo "test passed"; else echo "test failed"; exit 1; fi
```
gerrycampion, line 39: I just want to confirm - If the test fails, the workflow fails, right?

.github/workflows/build-binary.yml
```diff
@@ -0,0 +1,50 @@
...
+          else
+            chmod +x core
+            if ./core --help; then echo "test passed"; else echo "test failed"; exit 1; fi
+          fi
+        shell: bash
+      - name: Archive Binary
```
gerrycampion, line 45: The archive upload step should go before the test step, that way it will be available for troubleshooting even if the test fails. See next comment

.github/workflows/build-binary.yml
```diff
@@ -0,0 +1,50 @@
...
+        id: archive
+        uses: vimtor/action-zip@v1
```
gerrycampion, line 47: I think you should use `actions/upload-artifact@v4` here instead of this.
`upload-artifact` zips and uploads to make it available on the actions screen. I think `action-zip` only zips, which means this zip file will never be used.

gerrycampion (changes requested):
looks great now. a few smaller updates...

.github/workflows/build-binary.yml
SFJohnson24: yes, if the test fails the workflow fails

.github/workflows/build-binary.yml
SFJohnson24: I reordered the steps and changed to actions/upload-artifact from action-zip

gerrycampion (approved):

PR #969, opened 2025-01-06
Release unittest

SFJohnson24:
this PR does 3 things:
- adds a click command to use the pyreadstat module, write a .sav file, then read it.  
Additionally it adds logic within the build-binary action to use this function to test each OS release's pyreadstat module functionality using this command
see: https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/12755487773

-adds a click command to run a test validation which is then run by the executable in the gitaction container

- updates the release.yaml to use the matrix build_binary.yaml see: https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/12755487773  The release and PyPi deploy is preserved, the run posted above of the test release which is identical to the release will produce the same executables.


SFJohnson24:
here is ubuntu with the release: 
image

RamilCDISC (approved):
The PR updates the release automation on GitHub. The issue with pyreadstat library is solved. During the initial validation of build it was found that the created builds don't work other than the execution of --help command.

The PR was then updated to fix the issue. New builds created by the updated automations were again validated. The builds worked for windows and linux but for mac it showed errors and did not execute. After looking into the issue, it was discovered that the build is not working because of apple's quarantine. The build was tested again after removing the quarantine but still it was discovered that there was discrepancy between test and validate command. The test command read the dataset properly and showed accurate number of records in the dataset sheet but validate command failed to do so.

After further changes and new tests added for the release the mac build was validated again.  All builds work fine now. The cases for incorrect path or missing files are also handled gracefully with error message available on the terminal

Following screenshots are from the successful running from MAC, WINDOW and UBUNTU builds plus reports are attached from mac, windows and ubuntu.  The validation on windows and ubuntu were done with help of @SFJohnson24 

MAC:
Screenshot 2025-01-15 at 2 37 24 PM

WINDOWS 
image (1)

UBUNTU
Screenshot 2025-01-15 at 2 45 46 PM


[mac_report.xlsx](https://github.com/user-attachments/files/18430153/mac_report.xlsx)
[windows_report.xlsx](https://github.com/user-attachments/files/18430154/windows_report.xlsx)
[ubuntu_report.xlsx](https://github.com/user-attachments/files/18430249/ubuntu_report.xlsx)

PR #970, opened 2025-01-08
Logging

SFJohnson24:
This PR fixes issues with errors in operations when a domain is not present.  This should result in a scope skip.  There was logic in the check and aggregate operations when executed to catch these missing variable/dataset issues but the domain check fell outside of this.  I have created a custom error with status skip for when it looks for a domain in the operation and it is not present.

This ticket also brings all custom exceptions in line under the EngineError class for consistency as well eliminating un-needed code in the rule engine exceptions

to test: run editor locally with this branch of engine attach to python functions, test TIG0005 with positive data 2 on sharepoint.  TIG0008/9 also experience this
I ran CG0007 to ensure past skips were preserved

### DIFF
--- a/cdisc_rules_engine/exceptions/custom_exceptions.py
+++ b/cdisc_rules_engine/exceptions/custom_exceptions.py
@@ -45,29 +45,36 @@ class VariableMetadataNotFoundError(EngineError):
     )
 
 
-class DomainNotFoundInDefineXMLError(Exception):
+class DomainNotFoundError(EngineError):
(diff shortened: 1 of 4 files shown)


RamilCDISC (approved):
The issue is being validated running the engine locally with @SFJohnson24 branch as backed. TIG0005 is being used are rule and passed in local editor. Run test using positive and negative both datasets. The engine works fine and shows scope skips and it must be.

PR #979, opened 2025-01-19
735 rule blocked corerules cg0019

gerrycampion:
- logging fix
  https://github.com/cdisc-org/cdisc-rules-engine/pull/979/files#diff-33a60fe7b374d92b667a12922ca45cd9092a95e183e429941b1490ec4d3f0543
- better domain and dataset metadata handling
  https://github.com/cdisc-org/cdisc-rules-engine/pull/979/files#diff-a587bf003c12391dbabcf84a6f4ad13d5da388cc57cf65701f55817c3cbfceb2
  https://github.com/cdisc-org/cdisc-rules-engine/pull/979/files#diff-85625e39b07f9388ff317713f2747de66c9433e69ff93aa0dd72a6fe9ec7e446
- new builder
  https://github.com/cdisc-org/cdisc-rules-engine/pull/979/files#diff-d5ba947905c3e5b29054393aeec9088e2ad461790058c0ab74de3cebf3223ccb
- merge pivot supp update - moved from base_data_service to dataset processor
  https://github.com/cdisc-org/cdisc-rules-engine/pull/979/files#diff-1dffb7e663f4da3b3f4ee46f399bc0514405e6d21479dda7f83c6c2e5264f8b1
- display column values when column names are part of output
  image
  https://github.com/cdisc-org/cdisc-rules-engine/pull/979/files#diff-d3ea7b7bbd3500e3df66c640297815867ed8b2caf435c52eca2b5c9fb6626889
- Example
```yaml
Core:
  Id: CDISC.SDTMIG.CG0019
Description: Trigger error if records are not unique as per sponsor defined key
  variables as documented in the define.xml
Rule Type: Dataset Contents Check against Define XML
Match Datasets:
  - Is Relationship: true
    Keys:
      - USUBJID
    Name: SUPP--
Check:
  all:
    - name: define_dataset_key_sequence
      operator: is_not_unique_set
Sensitivity: Record
```

Sample test data:
[CG0019_split_and_supp.xlsx](https://github.com/user-attachments/files/18564627/CG0019_split_and_supp.xlsx)
Sample define.xml:
[define_CG0019_split_and_supp.xml.txt](https://github.com/user-attachments/files/18647778/define_CG0019_split_and_supp.xml.txt)

Sample report:
[CORE-Report-2025-01-27T16-18-32.xlsx](https://github.com/user-attachments/files/18564613/CORE-Report-2025-01-27T16-18-32.xlsx)


RamilCDISC (approved):
The PR implements the logic so the engine can work with cases where both split and supp datasets are available for a domain. Initially,  to accomodate CG0019, an if statement was used checking if the rule running is CG0019 and handle the case required. Now the engine is updated to handle all such rules. 

1.  Other than these a lot of enhancements have been made to the engine code. The code is now more cleaner and readable. 
2.  With the changes for cleaning the code, the tests have also been updated ensuring that the changes does not break any part of the system. 
3.  Regression test passing shows that the overall functionality is preserved. 

To verify this PR the CG00019 rue as run against a dataset having split and supp dataset. The engine succesfully executed the rule and correctly identified the non unique records from the split and supp datasets. Also it acurately identifies the sponsor variables from the define xml. to cross valdiate the domain of the split dataset was changes and the engine ignored the non unique record which was not uniqe because of split dataset. Attached is the screnshot of the successfull execution.

PR #981, opened 2025-01-20
removed pyreadstat-test artifact

SFJohnson24:
this was an artifact that I meant to remove.  pyreadstat is used within the XPTreader module so if the XPT validation within the test_validate runs, the pyreadstat module works appropriately.  It was determined this was a good enough test of its functionality and a unittest was unnecessary and clogs up our core.py file.  
this is the test build run without this test to show it has been removed with no impact to release creation:  https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/12871561191

### DIFF
--- a/.github/workflows/build-binary.yml
+++ b/.github/workflows/build-binary.yml
@@ -64,13 +64,3 @@ jobs:
             echo "Test validate failed with exit code $exit_code"
           fi
         shell: bash
-      - name: Test pyreadstat
-        run: |
-          cd dist/output/${{ inputs.name }}/core
-          if [ "${{ runner.os }}" = "Windows" ]; then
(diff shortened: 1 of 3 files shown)


RamilCDISC (approved):
The PR removes pyreadstat test from the core, which was added as a unittest for release. It was decided that successful running of validate command is in itself enough testing for pyreadstat module to work fine. To validate this the release was used to run validation against multiple xpt datasets. The engine ran validation successfully and executed all  rules. To cross validate a mix of json and xpt dataset folder was also used for which engine gracefully puts error and  exits. Attached screenshot shows the successfully execution and a report of validation against xpt datasets is also attached. 


[CORE-Report-2025-01-20T13-26-49.xlsx](https://github.com/user-attachments/files/18481867/CORE-Report-2025-01-20T13-26-49.xlsx)

PR #986, opened 2025-01-23
enhancement: pull custom domains from lib model

SFJohnson24:
[Rule_underscores.zip](https://github.com/user-attachments/files/18526225/Rule_underscores.zip)
this PR corrects how our SDTM utility gets metadata for custom datasets, using just the model for custom datasets.  I have included the rule, a test dataset and the define above.  There is also an excel sheet of the custom dataframes included in the sample data--of note, XS does not contain any of the variables to indicate a class so its class cannot be determined so it has a None-class and thus produces an error.   The results of the validation are here: [CORE-Report-2025-01-23T14-31-18.xlsx](https://github.com/user-attachments/files/18526364/CORE-Report-2025-01-23T14-31-18.xlsx)

I also converted outer joins into left joins on variable names as this was creating incorrectly shaped dataframes as seen here: [left_outer.xlsx](https://github.com/user-attachments/files/18527411/left_outer.xlsx) where there was library data added to the frame without any variable data

to validate:
- test send9.1 to insure changes in this branch do not change the rule's results as the operator that calls the sdtm_util was updated to reflect the changes made when that same function was called in data_service
- perform test validation with data/define/rule
- to fully validate this we will need a dataset + define that has a custom domain and correlated variables.  I tagged Nick and Gerry in the zenhub ticket regarding this.


SFJohnson24:
[xd.zip](https://github.com/user-attachments/files/18715268/xd.zip)
here is the data for test validation

SFJohnson24:
I have adapted the code to fit the new PRs using SDTMDatasetMetadata.
This PR fixes several bugs including:
- lack of a get attribute for SDTMDatasetMetadata for instantiating dummy dataset for test command
- some filter logic in dummy dataset when first_record exists and the record parsing logic was not working, resulting in no first record for the SDTMDatasetMetadata and thus no domain to check during rule execution.
- resolved bug where XPT files run in test did not have number_of_records

RamilCDISC (approved):
The PR adds the support for proper execution of CG00010 that requires cross check of define xml against the library metadata. The update to the code successfully executes CG0010. For validation following steps were performed.

1. Code Review: The PR was reviewed closely to ensure that the code is according to the coding standards followed with in the project level. 
2. Tests: The unit and regression tests passing shows that the changes do not break any other part of the code or overall execution of the engine.
3. Manual validation: The rule CG00010 was ran against multiple datasets with define.xml. The result generated by the engine were validated to be true. 

   - The rule was changed to include library variable name in the report for cross check.
   - The variable that had different value for role in define and IG were confirmed manullay to be different by converting the IG (standard_details.pkl) from cache to json.
   - The values for role were found different for the variables pointed by the engine.
   
The generated report form the engine is attached below. Rule, dataset and define file are attached in the PR description and Issue comments. To reproduce same result run the test command with the test artifacts

[CORE-Report-2025-02-13T15-36-19.xlsx](https://github.com/user-attachments/files/18790975/CORE-Report-2025-02-13T15-36-19.xlsx)

PR #987, opened 2025-01-23
add sorting by ordinal

RamilCDISC:
The PR adds sorting to the variables metadata based on ordinal values. To test you can use the following rule and dataset. 
[dataset.json](https://github.com/user-attachments/files/18527488/dataset.json)
[rule.json](https://github.com/user-attachments/files/18527489/rule.json)


SFJohnson24 (changes requested):
this code does fit the theory of what is being requested but there is some type of bug in engine is executing library metadata.  I ran the rule with and without the changes and the resultant list for $ig_variables was identical so the sorting is failing.  I believe this is due to the following: If you put a breakpoint at line 24 when you execute your rule--you can see there are several examples of metadata that do not have 'ordinal' keys but instead have 'order_number'.  
example:
{'_links': {'parentClass': {...}, 'parentProduct': {...}, 'priorVersion': {...}, 'rootItem': {...}, 'self': {...}}, 'definition': 'The date or date and time of the assessment or the specimen or data collection from the subject, represented in a standardized character format.', 'describedValueDomain': 'ISO 8601 datetime or interval', 'label': 'Date/Time of Collection', 'name': '--DTC', 'role': 'Timing', 'simpleDatatype': 'Char', 'variableCcode': 'C82515', 'order_number': '26'}
if you look library-- https://github.com/cdisc-org/cdisc-library-src-files/blob/a337bb5ec454b3c773e932e96d5cb41d6b732971/cdisc-json/products/data-tabulation/sdtm-2-0.json#L995 you can see this was 'ordinal: 26' before engine processed it.

I believe these keys formerly had ordinal but were changed by this map:   https://github.com/cdisc-org/cdisc-rules-engine/blob/42e56aaf506daa39b7b40c6fa1916a241c176569/cdisc_rules_engine/dataset_builders/base_dataset_builder.py#L174 in order to become "library_variable_order_number" for some of the dataset builders.  I am not certain where this method is being called along the thread of execution but we need to remedy this for these library operators as this may also be impacting get_column_model_order() and get_model_filter_variables() as well

PR #990, opened 2025-01-27
Performance testing

### LINKED ISSUE #881: Benchmark baseline testing - Reports and test data
### Reporting
For each of the following, we want number of calls, cumulative time, mean time, median time
- Total run time
- Time taken for each rule and each dataset
- Time taken for each rule
- Time taken for each dataset
- Time taken to preprocess each dataset
- Time taken for each Operator
- Time taken for each Operation
- Time taken for each Dataset Builder
- Any remaining time unaccounted for
- Telemetry reports (OpenTelemetry)
  - Need to ensure that these reports do not have any data or proprietary information

### Test Data
Test data should include:
- A set of small datasets and define-xml (Nick's certification data)
- A set containing at least one large (~1GB) dataset
- All currently published rules

RamilCDISC:
The pull request adds performance Test to the tests folder. The performance test can be run by following command 
`python tests/PerformanceTest.py -dd path/to/folder/with/datasets -rd path/to/folder/with/rule/files -total_calls 1`
Total calls define how many times do you want to execute a rule. The report will be saved in the directory from whic the test file was executed


.flake8
```diff
@@ -2,10 +2,14 @@
...
+count = True
+show-source = True
 exclude = .github,
     .pytest_cache,
     cdisc_rules_engine/resources,
+    tests/PerformanceTest.py,
     venv,
     build,
-    dist
+    dist
```
gerrycampion, line 15: No newline at end of file. See the "insert final newline" setting in vscode

tests/PerformanceTest.py
gerrycampion, line 1: Can you add the new cli instructions to the readme

gerrycampion (changes requested):

README.md
```diff
@@ -73,6 +73,24 @@ From the root of the project run the following command (this will run both the u
...
+### **Performance Testing**
+
+This repository includes a performance testing script located in the `tests` folder under the filename `PerformanceTest.py`. The script is designed to evaluate the execution time of rules against datasets by running multiple test iterations.
+
+### Running the Performance Test
+
+To execute the performance test, navigate to the root directory of the project and run the following command:
+
+```sh
+python tests/PerformanceTest.py -dd <DATASET_DIRECTORY> -rd <RULES_DIRECTORY> -total_calls <NUMBER_OF_CALLS> -od <OUTPUT_DIRECTORY>
```
gerrycampion, line 85: If this is the same as the `-d, --data TEXT` and `-lr, --local_rules TEXT` args in the `validate` command, it would be good to use the same arg names.

README.md
```diff
@@ -73,6 +73,24 @@ From the root of the project run the following command (this will run both the u
 
 `python -m pytest tests`
 
+### **Performance Testing**
+
+This repository includes a performance testing script located in the `tests` folder under the filename `PerformanceTest.py`. The script is designed to evaluate the execution time of rules against datasets by running multiple test iterations.
+
+### Running the Performance Test
```
gerrycampion, line 80: This and the next header should be nested under the Performance Testing header. IOW, use 4 `#`

README.md
```diff
@@ -73,6 +73,24 @@ From the root of the project run the following command (this will run both the u
...
+
+To execute the performance test, navigate to the root directory of the project and run the following command:
+
+```sh
+python tests/PerformanceTest.py -dd <DATASET_DIRECTORY> -rd <RULES_DIRECTORY> -total_calls <NUMBER_OF_CALLS> -od <OUTPUT_DIRECTORY>
+```
+### Performance Test Command-Line Flags
+
+- **`-dd` (Dataset Directory)**: The directory containing the dataset files in `.json` or `.xpt` format.
```
gerrycampion, line 89: Format the documentation similar to the existing documentation. For example, use a code block for the args

gerrycampion (changes requested):

RamilCDISC:
Since engine is updated to remove test column, I will re-go over my code and update it to be according to updated engine. Putting back to in progress for now. @SFJohnson24 @gerrycampion

tests/PerformanceTest.py
```diff
@@ -0,0 +1,509 @@
...
+        rule_name = os.path.basename(rule)
+
+        # Initialize variables to collect times for the dataset across all rules
+        all_time_taken = []
+        all_preprocessing_times = []
+        all_operator_times = {}
+        all_operation_times = []
+
+        rule_names = []
+        for dataset_path in dataset_files:
```
gerrycampion, line 263: It looks like this method is essentially running the same set of commands as the previous function, just in a different order. I think it is redundant.
I don't think it supports rules where the rule might require multiple datasets that are joined together since only a single dataset is being passed at a time. I think the timing feedback needs to given from within the engine. Maybe an additional logging mechanism can be used.

gerrycampion (changes requested):

PR #994, opened 2025-01-30
increment black version in git action/precommit to match requirement.txt

SFJohnson24:
this PR increments the gitaction CI linting and pre-commit config to match the version that was incremented in requirements.txt


RamilCDISC (approved):
updated accordingly

PR #997, opened 2025-01-31
Fix get column order from library

RamilCDISC:


cdisc_rules_engine/dataset_builders/define_variables_with_library_metadata.py
SFJohnson24, line 1: I think what the goal should be is to preserve the prior function calls, remove the map involving ordinal to order_number, and have the name change from library_variable_ordinal to library_variable_order_number in the builders to prevent this mixing of ordinal and order_numbers in the library metadata which seems to be causing issues.

cdisc_rules_engine/dataset_builders/variables_metadata_with_define_and_library_dataset_builder.py
```diff
@@ -42,7 +44,26 @@ def build(self):
         define_metadata: DatasetInterface = self.dataset_implementation.from_records(
             variable_metadata
         )
-        library_metadata: DatasetInterface = self.get_library_variables_metadata()
+        variables: List[dict] = sdtm_utilities.get_variables_metadata_from_standard(
```
SFJohnson24, line 47: again, I would leave 
```library_metadata: DatasetInterface = self.get_library_variables_metadata() ```
instead of copying its code into each builder.

File: cdisc_rules_engine/dataset_builders/variables_metadata_with_library_metadata.py [hunk 2]
```diff
@@ -30,6 +33,28 @@ def build(self):
         )
         dataset_contents = self.get_dataset_contents()
         library_variables_metadata = self.get_library_variables_metadata()
+        variables: List[dict] = sdtm_utilities.get_variables_metadata_from_standard(
```
SFJohnson24, line 36: same as above-- we can call ```library_variables_metadata = self.get_library_variables_metadata()``` no need to copy its code into each builder

cdisc_rules_engine/operations/library_column_order.py
```diff
@@ -21,6 +21,10 @@ def _execute_operation(self):
         # get variables metadata from the standard model
         variables_metadata: List[dict] = self._get_variables_metadata_from_standard()
 
+        # sort variables_metadata by the "ordinal" key
```
SFJohnson24, line 24: this looks good.

cdisc_rules_engine/utilities/sdtm_utilities.py
```diff
@@ -229,11 +229,11 @@ def standardize_order_number(var):
             var["order_number"] = var.pop("ordinal")
         return var
 
-    identifiers_metadata = list(map(standardize_order_number, identifiers_metadata))
-    timing_metadata = list(map(standardize_order_number, timing_metadata))
+    # identifiers_metadata = list(map(standardize_order_number, identifiers_metadata))
```
SFJohnson24, line 232: remove code you are commenting out in a pull request to keep things clean.
I believe we can also remove the standardize_order_number() function entirely as it is no longer called.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-black==24.10.0
+black==22.6.0
```
SFJohnson24, line 1: see above re:deprecating the linter

tests/unit/test_dataset_builders/test_define_variables_with_library_metadata.py
```diff
@@ -116,11 +116,11 @@ def test_define_variables_metadata_with_library_metadata_dataset_builder(
         "define_variable_mandatory",
         "define_variable_has_comment",
         "library_variable_name",
+        "library_variable_ordinal",  # Added this line
```
SFJohnson24, line 119: - try to check for comments being pushed with PRs.

- We should move to preserve ```library_variable_order_number``` as it may be used in rules that have already been created.  see comment above re:moving the renaming logic into the builders.

SFJohnson24 (changes requested):
I appreciate the effort you put into this. There are some areas that need improvement--see my comments and try to refactor it.  Let me know if you’d like to go over it together

cdisc_rules_engine/dataset_builders/define_variables_with_library_metadata.py
```diff
@@ -35,7 +35,6 @@ def build(self):
             self.get_define_xml_variables_metadata()
         )
         library_variables_metadata = self.get_library_variables_metadata()
-
```
SFJohnson24, line 38: We should try to preserve the variables names in the reference guide, especially since library_variable_order_number is more descriptive than library_variable_ordinal.  Can you add map logic for renaming library_variable_ordinal to library_variable_order_number in the dataset builders that have the variables?  The variables in the corresponding tests would have to be restored as well

SFJohnson24 (changes requested):

SFJohnson24:
after reviewing with Marcelina-- this appears to have been fixed and the operator is functioning as expected

Issue #287, opened 2023-02-28
As VALIDATOR, I want to cross-check my study data against the study define.xml for inconsistent use of the HasNoData attribute


chowsanthony:
This check only applies to study define that uses Define-XML v2.1.

When a variable has `ItemDef[@HasNoData]` set to "Yes", then check if the variable in the study data should always be null.

gcampion-cdisc:
Do we also need the crosschecks in the other directions?

- ItemDef[@HasNoData] eq "Yes" -> Data Variable all null  (^^specified above)
- ItemDef[@HasNoData] ne "Yes" -> Data Variable has a non null
- Data Variable all null -> ItemDef[@HasNoData] eq "Yes"
- Data Variable has a non null -> ItemDef[@HasNoData] ne "Yes"

nickdedonder:
@RamilCDISC can you confirm the estimate from QA point of view?

gerrycampion:
I believe this impacts rule CG0015: https://jira.cdisc.org/browse/CORERULES-45
Condition: Variable Core Status = Permissible and data is collected
Rule: Variable present in dataset and ^= null

RamilCDISC:
### Validation Summary
### Validation Methodology:
The validation process incorporated rigorous testing methodologies to ensure the accuracy and reliability of the system. This included:
**Manual Testing:**
Manual Testing was done. The test functionality of rules engine was used to run validations and verify the output, to check compliance with the Acceptance criteria.
**Unit Testing:**
Unit Testing was done for validating the issue. The unit tests were performed covering edge test cases. The unit test were run and their results were assert tested to verify the compliance with the acceptance criteria.
**Code Review:**
The code changes in the code files were reviewed and were found to be in accordance with the Acceptance criteria. Also, the code was reviewed for following: 

- Code Quality Evaluation
- Logic and Functionality Review
- Error Handling and Exception Handling Assessment
- Security Considerations


**Regression Testing:**
Regression tests were run to validate that the changes did not affect any functionality of the overall system.

### Test Execution:

> The Tests were run by executing validation against different datasets. Following command was used
> python -m pytest tests/unit/test_operations/test_variable_is_null.py
> Python -m pytest tests/unit/test_define_xml_reader.py
> 
> The regression tests were executed using the following command:
> Python -m pytest test_cli.py


### Results:
After successful running of unit and regression tests and careful validation, the issue was found to be good to close as the Acceptance criteria was fulfilled and all validations passed.
[...]

PR #1006, opened 2025-02-05
CG0019 regression test + updates to previous tests

RamilCDISC:
This PR adds a regression test for CG0019 and updates the old regression tests to be coherent with the current state of the engine.


tests/QARegressionTests/test_Issues/test_CoreIssue747.py
```diff
@@ -0,0 +1,79 @@
...
+    rule_status_column_values = [cell.value for cell in rule_status_column[1:]]
+
+    # Remove None values using list comprehension
+    coreid_values = [value for value in coreid_values if value is not None]
+    dataset_values = [value for value in dataset_values if value is not None]
+    rule_status_column_values = [
+        value for value in rule_status_column_values if value is not None
+    ]
+
+    # Perform the assertion
```
SFJohnson24, line 65: I think it would be good to add a line catching the stderr
```assert process.returncode == 0, f"Process failed with error: {stderr.decode()}"``` something like this so if the command fails, we can catch and display the error to the dev

File: tests/resources/CoreIssue287/is_null_rule.json [hunk 2]
```diff
@@ -1,41 +1,109 @@
...
+      ]
+    }
+  ],
   "Check": {
     "all": [
       {
-        "name": "define_variable_has_no_data",
-        "operator": "equal_to",
-        "value": "Yes"
+        "name": "variable_name",
```
SFJohnson24, line 75: I may be out of the loop here--the name of this rule seems to indicate is a null check, and it is using operators that check for blank define/variables.  Why are we changing it versus retaining it in the testsuite and adding a new rule that tests the operators you added?

SFJohnson24 (changes requested):
updates look good--just a small change and the clarification on that rule that is being eliminated (we should change the rule's name if it is to be eliminate it).

tests/resources/CoreIssue287/is_null_rule.json
RamilCDISC: I went over the issue 287 again and saw there that the issue is related to CG0015. I updated the rule file to be CG0015. [See here](https://github.com/cdisc-org/cdisc-rules-engine/issues/287#issuecomment-1501997654)

SFJohnson24 (approved):
PR looks good with requested changes.

PR #1010, opened 2025-02-07
Bypass InMemoryCacheService

gerrycampion:
This update has the `update-cache` command bypass the inmemorycache. See the connected issue for the reasons.

To test, simply run the `update-cache` command and then run a validation. The update-cache command should take significantly less time than main. The resulting cache files should be the same size (and in most cases where ordering is preserved, the same content). The exception would be the rules cache which has updated rules files.

### DIFF
--- a/cdisc_rules_engine/services/cache/cache_populator_service.py
+++ b/cdisc_rules_engine/services/cache/cache_populator_service.py
@@ -7,7 +7,7 @@ from cdisc_library_client.custom_exceptions import (
     ResourceNotFoundException as LibraryResourceNotFoundException,
 )
 
-from cdisc_rules_engine.enums.library_endpoints import LibraryEndpoints
+from cdisc_rules_engine.enums.default_file_paths import DefaultFilePaths
 from cdisc_rules_engine.interfaces import (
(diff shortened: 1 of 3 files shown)


RamilCDISC (approved):
The PR updates the update_cache functionality to make the cache update more robust and faster. The validation was done in the following way:


- [ ] Code Quality Assessment: the quality of code was manually validated and is up to mark with the standards followed in our software code.
- [ ] Functional Validation: The changes were tested to confirm that everything works as intended. To perform this validation update cache command was run using the updated code. The check passed.
- [ ] Compare with Current update cache function. Current version of update cache function was ran to get results. These results were then compared to the results from the updated function. The new function got all the cache files successfully. The number of cache files was same.




- [ ] Compare content cache files: for deeper comparisons, a python script (attached) was developed to  cross check the content between cache files from updated and older update cache function. As expected, It was found that the updated function rules.pkl file has more rules then the old update cache function. 



[test_same_file_content_for_cache.py.zip](https://github.com/user-attachments/files/18713585/test_same_file_content_for_cache.py.zip)

PR #1013, opened 2025-02-07
Fix missing standard_substandard arg

gerrycampion:
Fixes a large bug where standard_substandard was not being assigned to one of the DataServiceFactory calls, causing a parameter mismatch and library_metadata was getting the max_dataset_size instead of the library_metadata. Any rules involving library_metadata would be skipped.

To show that the new code works, you can run a validation before and after the branch change on this sharepoint data:
unitTesting\Samples and Templates\CDISCPILOT01\Issues\xpt

You will see a lot less rules skipped after the update. See the connected issue for a screenshot comparison. See the following old and new reports:
[Broken report](https://github.com/user-attachments/files/18710645/CORE-Report-2025-02-07T10-59-25.xlsx)
[Fixed report](https://github.com/user-attachments/files/18710646/CORE-Report-2025-02-07T12-10-29.xlsx)

### DIFF
--- a/cdisc_rules_engine/rules_engine.py
+++ b/cdisc_rules_engine/rules_engine.py
@@ -67,12 +67,13 @@ class RulesEngine:
         self.dataset_paths = kwargs.get("dataset_paths")
         self.cache = cache or CacheServiceFactory(self.config).get_cache_service()
         data_service_factory = DataServiceFactory(
-            self.config,
-            self.cache,
-            self.standard,
-            self.standard_version,
-            self.library_metadata,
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
The PR fixes the bug which was cause of alot of rules being skipped by the engine. This is one part of the bigger ticket (1007). For initial valdiaiton the changes show that the engine now is skipping way less rules. All unit test passing validates further. 

For Now we are approving this PR based on half validation. For full validation rule regression tests will be required, which are under development. Once the tests are ready the PR will be re-validated

PR #1028, opened 2025-02-14
Schema and docs update for is_unique_set supporting CG0019

gerrycampion:
`is_unique_set`
`name` can be a variable containing a list of columns and `value` does not need to be present

### DIFF
--- a/resources/schema/Operator.md
+++ b/resources/schema/Operator.md
@@ -727,6 +727,8 @@ Multiple grouping variables:
 
 Relationship Integrity Check
 
+`name` can be a variable containing a list of columns and `value` does not need to be present
+
 > --SEQ is unique within DOMAIN, USUBJID, and --TESTCD
 
 ```yaml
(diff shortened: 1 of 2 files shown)


resources/schema/Operator.md
```diff
@@ -727,6 +727,8 @@ Multiple grouping variables:
 
 Relationship Integrity Check
 
+`name` can be a variable containing a list of columns and `value` does not need to be present
```
SFJohnson24, line 730: I think it might be good to change one of the yaml examples for either is_unique_set or is_not_unique_set to match the case you are outlining where Name is a list of columns and no value is given

SFJohnson24 (changes requested):

SFJohnson24 (approved):
This PR updates the schema requirements for is_not_unique_set / is_unique_set operators appropriately as well as providing examples within the docset.

RamilCDISC (approved):
Updated as required.

PR #1029, opened 2025-02-17
Codelist schema

SFJohnson24:
This PR fixes the schema bug that Jozef describes in the ticket.  It also adds codelists and extensible which are properties of other operations that were not added to the schema accidentally.


RamilCDISC (approved):
The PR fixes the bug where the editor required the presence of codelist property when operator is set to codelist_extensible, but when codelist is added the schema did not recognize it. The schema was missing those properties. The PR has added them. To test the rule editor was run locally with schema set from the updated branch. The bug is now fixed. See attached screenshot for reference

PR #1036, opened 2025-02-18
Fixed row number json serialization

gerrycampion:


SFJohnson24 (approved):
code correctly remedies issue with errors containing n64 Numpy characters.  
restores editor functionality when run in tandem with engine locally on both positive and negative data.
image

validate command with cert datasets also runs well.  [CORE-Report-2025-02-18T11-08-54.xlsx](https://github.com/user-attachments/files/18848170/CORE-Report-2025-02-18T11-08-54.xlsx)

PR #1038, opened 2025-02-18
Update rule-blocking-bug.md

gerrycampion:
make issues template use `Types` instead of labels


RamilCDISC (approved):
The PR updates the issue template to use type keyword instead of label to represent type of issue. The update is done as required.

PR #1051, opened 2025-02-26
Fix permissible/required/expected sort order

SFJohnson24:
- This PR updates the docs to include a definition of what perm/req/exp are in the library data as well as explain where library variable metadata is drawn from for custom and non-custom domains.  
- it removes the problematic set logic from these 3 operators as this was removing the ordinal sort order from these operators.  The new logic uses a dictionary comprehension which preserves insertion order.  We then take the keys and convert that to a list, thus preserving the order while also preserving the uniqueness set() was implemented to do (I assume)
- corrects the filtering in get_variables_metadata_from_standard so that custom domains come from model and non-custom come from model and IG.

This PR also includes the bug fix which removes the problematic call in the base_operation to the dataservice that was causing issues and retrieves the metadata which has already been acquired at this stage of a validation.

to test:
run cg0219
test a rule with the permissible, expected, required_variable operations and compare the report with the previous report


RamilCDISC (approved):
The validation for the pr is complete.  This PR merges the initially two different PR's.
 
1. Bug fix which removes the problematic call in the base_operation to the dataservice that was causing issues. The bug us fixed and there is no more issue with the test function..the validation was done using manual testing where it was confirmed that the engine does not raise and error anymore. The rule ran successfully and did the checks properly. 
2. The other fix is for the ordinal (order number sorting). The bug was identified and fixed. The problematic code causing the sorting to be destroyed is changed so the functionality remains same but the sorting is not affected. 

The following screenshot shows the successful runs

PR #1056, opened 2025-02-27
removed xml code

SFJohnson24:
This PR removes the logic that would go into the specified directory and look for a .xml file and assign that to the report, even if that define.xml was not specified in the validation.  Now unless the define_path or define_version are specified, no define_version will show up in report.

### DIFF
--- a/cdisc_rules_engine/services/reporting/excel_report.py
+++ b/cdisc_rules_engine/services/reporting/excel_report.py
@@ -120,9 +120,7 @@ class ExcelReport(BaseReport):
             if define_xml_path:
                 define_version = get_define_version([define_xml_path])
             else:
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
The PR fixes the xml version bug. Now the version is only mentioned in the report when there is a define path or define version given explicitly. Following steps were taken for validation: 
1. Run test command with rule and dataset without a define file (No define version in report found)
2. Run test command with rule and dataset with a define file path. (Correct version of define was put in the report)
3. Run test command with -dv flag and no define file path. (The -dv flag version was added in the report).
4. Run test command with define xml path and -dv both flags. The -dv flag was given 2.0 as version where as define xml had version 2.1 (The define xml version was put in the report).
5. All the above test with validate command. ( Same outputs)

Issue #1064, opened 2025-03-06
TIG0037


SFJohnson24:
SEND1/TIG0037 -- TIG0037 returns a 500 error, Send1 does not in editor

[unit-test-coreid-TIG0037-negative1.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK1FtQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e5480bc265302c28ea38f3f05af23360638058a9/unit-test-coreid-TIG0037-negative1.xlsx)
[unit-test-coreid-SEND1-negative1.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK1VtQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4cdc31e9f78dfe51bad10acf491be6ffe1ae7303/unit-test-coreid-SEND1-negative1.xlsx)

PR #1065, opened 2025-03-06
bug fix

SFJohnson24:
the issue was that in the test rule it was grabbing the first standard under authorities instead of iterating through them and looking for tig

to test: run editor/engine locally and test tig0037 with data in #1064

### DIFF
--- a/TestRule/__init__.py
+++ b/TestRule/__init__.py
@@ -55,7 +55,7 @@ def handle_exception(e: Exception):
         )
 
 
-def main(req: func.HttpRequest, context: func.Context) -> func.HttpResponse:
+def main(req: func.HttpRequest, context: func.Context) -> func.HttpResponse:  # noqa
     try:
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
The bug that was causing the TIG0037 rule ending in 500 error is fixed. The issue was in the TestRule file, where appropriate logic was missing. Following steps were done to validate:
1. Run the editor and engine locally (works as expected)
2. Run a test with TIG0037 rule using TIG negative dataset. The editor no longer gives server error.
3. run a test with TIG0037 rule using TIG positive dataset. The specified operator executes successfully.
4. Run a test with TOG0037 rule using positive SEND1 dataset. The editor performs the validation successfully. 
5. Run a test with TOG0037 rule using negative SEND1 dataset. The editor performs the validation successfully. 
6. Run tests to check if other functionalities of editor are preserved

PR #1069, opened 2025-03-09
Custom schema

SFJohnson24:
this PR does a few things:
- updates the gitignore for the custom_schema.json, This will allow prevention of change tracking to that file but most importantly--it won't commit those changes, and importantly, won't update that file during operations like pulls or merges.
- readme update explaining the custom schema
- Organization_Custom schema with some premade properties
- updated the base schema to select the Organization_Custom subschema as well as defining `Category` where custom properties are held


RamilCDISC (approved):
The validation for PR is complete. The PR address the guide to use custom columns and custom rules. The validation was done by running the editor and engine locally. following steps were followed:

1. Run the rules against datasets and confirm the working of the execution.(works as expected)
2. Run the sorting of the rules by searching (works as expected)
3. Run the sorting by introducing the custom column (works as expected)
4. Add custom elements to the rule in the editor (works as expected)
5. Add custom column based on the custom element in the rule (works as expected)
6. Check for integration. (unit test and other test show integration. (works as expected)

PR #1071, opened 2025-03-11
Custom bug

SFJohnson24:
image
a bug was pushed to main with the custom schema when you hover over authorities in any rule.
this is resolved with this PR.
- category was moved under authorities in base schema to reflect the way Organization_CDISC/FDA/Custom subschemas are imported
- category was made a requirement for the custom schema to prevent the error shown above where multiple schemas can be applicable
- updated readme for new pathing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -147,4 +147,4 @@ resources/cache/local_rules.pkl
 #ignore vim .swp files
 *.swp
 
-# Organization_Custom.json
+Organization_Custom.json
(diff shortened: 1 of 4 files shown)


RamilCDISC (approved):
The PR fixes the bug which was causing the issue in validation of schema on the rules. The bug was fixed by updating the schema to incorporate category with in the Authority class. The related documentation is also updated. The custom column adding and filtering works fine. The updated schema lets the user to define their own properties in the rules.

PR #1081, opened 2025-03-13
usdm to use datasemetada class

gerrycampion:
problem is connected issue is resolved. Can be tested using the data and command connected to the ticket.


RamilCDISC (approved):
The bug for USDM validation in CLI is fixed. The engine is updated to handle usdm datasets proeprly. The validation was done with following steps:
1. Run the rules and dataset attached in the AC, using the command given there. ( works as expected)
2. Verify the rules against the dataset to validate that the engine executes the rules properly. (works as expected)
3. Run another dataset and rule from usdm standard to verify proper functioning (works as expected)
4. Run the test command for usdm dataset and rule file, to validate the bug ifix across both validate and test commands (works as expected)
5. Run a usdm dataset and rule on rule editor to confirm successful execution in the editor (works as expected)

PR #1082, opened 2025-03-13
fix for expanding list of columns when column doesn't exist

gerrycampion:
Minor fix for reporting bug with missing variables.

See the attached ticket for test data. I tested with 78 and 98 and got the following results.
[DDF00098 - neg01 result.json.txt](https://github.com/user-attachments/files/19236791/DDF00098.-.neg01.result.json.txt)
[DDF00078 - neg01 result.json.txt](https://github.com/user-attachments/files/19236792/DDF00078.-.neg01.result.json.txt)

Note that the 98 results don't exactly match the original results, but I'm not sure if the original results were correct.

### DIFF
--- a/cdisc_rules_engine/models/actions.py
+++ b/cdisc_rules_engine/models/actions.py
@@ -71,11 +71,12 @@ class COREActions(BaseActions):
     def _get_target_names_from_list_values(self, target_names, rows_with_error):
         expanded_target_names = set(target_names)
         for target in target_names:
-            for candidate_list in rows_with_error[target]:
-                if isinstance(candidate_list, list):
-                    for value in candidate_list:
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
The PR fixes the bug which was causing the rules to skip when a variable is missing that  is in the check or the outcome section of the rule. The validation for the PR was done with following steps: 

1. Run a test where all variables in the check and outcome section are available in the dataset.
2. Run a test where one variable of the outcome is not in the dataset but all check variables are present.
3. Run a test where one variable of check section is not on the dataset but all outcome variables are available in the dataset.
4. Run a test where variables form both check and outcome sections are missing in the dataset.
5. Run validate command for above test cases

PR #1088, opened 2025-03-15
updated version to string

SFJohnson24:
the version on the deploy action was recently incremented to 3.10.  The gitaction is currently treating it as a decimal and install python 3.1, this PR converts that to a string to ensure it gets 3.10.  
see [here](https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/13872707280/job/38821410497) for bug


RamilCDISC (approved):
Updated as required

PR #1094, opened 2025-03-19
Test merged into validate

gerrycampion:
Removes a bunch of redundant, unsynchronized code.
- Removed `test` cli command and moved any test-specific logic into the `validate` command
- Removed entire `cdisc_rule_tester` package for the rule tester endpoint and moved required code to `run_validation.run_single_rule_validation` to parallel the similar cli functionality
- Removed `rule_engine.validate` because it was not being used
- Renamed `rule_engine.validate_single_rule` to `validate_single_dataset` to disambiguate functionality. Created and merged a new `rule_engine.validated_single_rule` from the other logic branches that validates all datasets against a single rule.
- Replaced many instances of `dataset_path` arguments when the `dataset_metadata.full_path` argument can be used instead


gerrycampion:
Currently test suite is failing because of template update on another branch. Requires update to comparison report template and passing the test suite before merging this.

RamilCDISC (changes requested):
I ran the editor locally using this branch's engine as the backend and tested it with DDF00078. The editor returned a server error, as shown in the attached screenshot.

To cross-check, I ran the same test using the main branch as the backend, and the error did not occur.

In previous PR validations, a similar issue was observed when the API key was missing from the environment. However, in this case, I have confirmed that the API key is correctly set in the .env file

gerrycampion:
@RamilCDISC I'm not able to reproduce the error you're seeing

RamilCDISC (approved):
The PR removes the test function from the engine. The test function is merged into the validate function. The validation for the changes was done by:

1. Going through the code to confirm there are no inconsistencies.
2. Checking if the associated documentation (read me) files are updated.
3. Checking if the code related to the test function is removed fully.
4. Running validation locally in terminal using validate command
5. Running validation with different set of flags
6. Running the editor locally and testing all types of standards for proper execution
7. Passing of unit tests + regression tests + test suite

PR #1098, opened 2025-03-21
update message length

SFJohnson24:
- updates the is_suitable_for_validation to return a tuple, the reason is then parsed into the error message for rule skipping.


RamilCDISC (approved):
The PR updates the error messages to include the reason. The code is accurately updated. The engine was validated by running test which results in skipping of rule. With debug mode on, the error messages can be observed to see the reason included as shown in the screenshot.

PR #1095, opened 2025-03-20
Get library class and domain only if standard and version provided

gerrycampion:

PR #1099, opened 2025-03-21
Standard ruletester

SFJohnson24:
- this PR removes the portion of RuleTester that grabs the standard/rule/substandard from the rule.  This will be provided by the library tab of datasets if needed.
- The new logic https://github.com/cdisc-org/cdisc-rules-engine/pull/1095 from this PR will account for the classes that can be IDed without the library tab.

To test: run TIG0006 without library tab; test some other tig rules with new logic


RamilCDISC (approved):
updated as required

PR #1110, opened 2025-03-25
missing all outcome vars logic changed

SFJohnson24:
this adds logic when all outcome variables are not present.  Instead of throwing an error, this code creates a single validation error record.  It will detect that none of the target variables exist in the dataset.  It then creates a single ValidationErrorEntity stating that none of the output variables are present.

to test--take any rule and add several outcome variables that do not exist in the dataset to trigger this logic


SFJohnson24:
this executes fine for me in both editor and the CLI.  Did you properly set the standard/version/substandard when using engine?  This would result in a class skip

RamilCDISC (changes requested):
I have attached a rule and dataset below to reproduce this bug. 

1. If there is a check that triggers the rule and there are outcome variables which is mix of varaibles that are mix of present and missing in the dataset, the engine outputs the values successfully in the issues page. Which is correct.

> "Check": {
        "all": [
            {
                "name": "--VALTRG",
                "operator": "matches_regex",
                "value": "^-?([1-9]\\d*|0)(\\.\\d+)?$"
            }
        ]
    }
    
  

>  "Output_Variables": [
            "--VALTRG",
            "--VALMAX",
            "--VALU"
        ]


2. But if the check triggers and there is only one variable in the outcome, which is missing in the dataset then the engine does not put any errors in the issues sheet. 

> "Check": {
        "all": [
            {
                "name": "--VALTRG",
                "operator": "matches_regex",
                "value": "^-?([1-9]\\d*|0)(\\.\\d+)?$"
            }
        ]
    }
    
  

>  "Output_Variables": [
            "--VALU"
        ]


The engine does not fills the issue sheet at all with the second case. Where as it should still fill the sheet and report the missing variable as not in dataset. Command used to execute 
`python core.py validate -s tig -ss sdtm  -v 1.0 -lr localrule/Rule_underscores.json -dp Rule/Datasets.json`

[Rule_underscores.json](https://github.com/user-attachments/files/19527408/Rule_underscores.json)
[Datasets.json](https://github.com/user-attachments/files/19527411/Datasets.json)

SFJohnson24:
@RamilCDISC nice catch, this is fixed

PR #1112, opened 2025-03-26
restore substandard

SFJohnson24:
restores deleted logic--this gets substandard for library tab.  test tig0016a (read description of rule as there are 2, look for part B) and test against Part B negative data


RamilCDISC (approved):
This PR solves the bug causing the tig rule to fail, which requires substandard from the library tab of datasets. The rule runs successfully now.

PR #1113, opened 2025-03-26
update error logic

SFJohnson24:
I was unable to reproduce this error with Marcelina's original code but I did find a rule where this issue was occurring.
CG0433 - negative dataset.  

The bug was in our create error object--it was converting a pd.series directly to a dictionary without processing any potential NaN's, resulting in incorrect json in editor and the NaN that I suspect Marcelina was encountering.  This PR resolves the issue with CG0433 and would likely resolve the issue Marcelina was having

### DIFF
--- a/cdisc_rules_engine/models/actions.py
+++ b/cdisc_rules_engine/models/actions.py
@@ -206,6 +206,11 @@ class COREActions(BaseActions):
         )
         source_row_number: Optional[pd.Series] = data.get(SOURCE_ROW_NUMBER)
         source_filename: Optional[pd.Series] = data.get(SOURCE_FILENAME)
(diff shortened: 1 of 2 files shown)


SFJohnson24:
per @eljanssens This NaN issue is reproduced with CG0114/TIG0377 - running the CG0114 test data.

SFJohnson24:
FB0911 was also encounter this issue, negative data

RamilCDISC (approved):
The PR updates the error logic that was causing a few rules to return invalid json due to presence of nan values. The PR has fixed the issue. All the rules that were known to have the bug have been validated to work fine now. The engine executes the rules successfully and accurately. All known rules that had bugs were validated for their results manually too.

PR #1115, opened 2025-03-28
update readme

SFJohnson24:
I updated the readme as I received some question regarding configuriation.  Added some clarifying details, a table of contents, and the issue submission portion at the bottom

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,32 @@
 
 [![Python 3.10](https://img.shields.io/badge/python-3.10-blue.svg)](https://www.python.org/downloads/release/python-3100)
 
+## Table of Contents
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2,6 +2,32 @@
 
 [![Python 3.10](https://img.shields.io/badge/python-3.10-blue.svg)](https://www.python.org/downloads/release/python-3100)
 
+## Table of Contents
```
gerrycampion, line 5: Is there a precommit hook to generate this automatically? Worried it will get out of sync quickly. Also, I'm not sure how necessary this is since gh already generates a TOC here: 

image

gerrycampion (changes requested):

gerrycampion (approved):

PR #1123, opened 2025-04-03
simplify and fix lib metadata

gerrycampion:
Fixes many issues with inconsistency when using the LibraryClient. See the attached issue for the command that was previously broken but now works.

### DIFF
--- a/cdisc_rules_engine/dataset_builders/base_dataset_builder.py
+++ b/cdisc_rules_engine/dataset_builders/base_dataset_builder.py
@@ -10,7 +10,6 @@ from cdisc_rules_engine.utilities.utils import (
     tag_source,
 )
 from typing import List, Iterable
-from cdisc_rules_engine import config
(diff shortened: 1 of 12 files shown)


RamilCDISC (approved):
The PR fixes the library metadata bug, which was causing sometimes the validations to fail for some rules. The updated code was validated by:

1. Running validations with SDTMIG datasets in terminal. 
2. Running validations with SDTMIG datasets in rule editor run locally. 
3.  Running validations with SENDIG datasets in terminal. 
3. Running validations with SENDIG datasets in rule editor run locally. 
4.  Running validations with TIG datasets in terminal. 
5. Running validations with TIG datasets in rule editor run locally. 
6.  Running validations with USDM datasets in terminal. 
7. Running validations with USDM datasets in rule editor run locally. 
8. Check rule CG0014 for proper execution (it gave library error before).

PR #1124, opened 2025-04-07
replaced list with pandas series

SFJohnson24:
the issue arises when a RDOMAIN is used to link the CO dataset to LB in positive 2 but an IDVAR is not present.  We were adding a blank list but it needed to be a pandas series as the .unique() method was called on the return value.

no longer getting errors for positive 2--correctly gives skips
image

test cg0370 / 371 locally with the changes in editor


RamilCDISC (approved):
The PR fixes the bug causing an error when executing the rule CG0370 and CG0371. 

The change was validated by:

1. Running the validation for both rules locally in engine
2. Running validation for both rule on editor ran locally
3. Validation of successful execution of unit test and regression tests

PR #1114, opened 2025-03-28
Direct excel datasets added

gerrycampion:
- Allows you to pass in an excel dataset file using the `--dataset-path` flag. The excel file format is the same as the format used for the rule editor. Multiple datasets can be included in a single excel file
- Unifies the separate data service `get_dataset_metadata` methods into a single method in the base class.

You can test with a rule and test data from the rule editor unit testing.
Note that library and ct need to be specified in the command line. They are not read from excel.

PR #1127, opened 2025-04-08
update null flavors

SFJohnson24:
This code adds the fixes from #1113 but ensures that cg0019 continues to work with the fixes.  (This was removed in https://github.com/cdisc-org/cdisc-rules-engine/pull/1114 this PR).  The old implementation used isna() but this assumed the value in the error was a scalar.  This adds logic for scalar values but also lists which was the issue.  I removed pd.NaN from null flavors as pd.NaN == pd.NaN comes out to false so having it in the array for an equality operator check will not work as intended.

adds logic for NaN scalar and list values.
to test: fb0907, cg0433 negative data; ensure that cg0019 validation function is retained


cdisc_rules_engine/models/actions.py
gerrycampion, line 1: considering we were back and forth on this, i think we should have add unit or regression tests to cover the 2 scenarios here.

gerrycampion (changes requested):

RamilCDISC (changes requested):
The rule CG0374 is affected by the changes to logic. The rule checks for null vlaue and is currently getting skipped because of an error. The error can be seen by setting the log level to debug mode. The command to reproduce it:

`python core.py validate -s sdtmig -v 3.4 -r CORE-000356 -dp unitTesting/SDTMIG/CG0374/negative/data/unit-test-coreid-CG0374-negative.xlsx -l debug`

The dataset is the first negative dataset in sharepoint for this rule

SFJohnson24:
I cant reproduce this error @RamilCDISC.
this is the validation with your command on main branch: 
[CORE-Report-2025-04-10T13-10-21.xlsx](https://github.com/user-attachments/files/19689159/CORE-Report-2025-04-10T13-10-21.xlsx)
this is the validation on NaN branch with your command:
[CORE-Report-2025-04-10T13-13-16.xlsx](https://github.com/user-attachments/files/19689163/CORE-Report-2025-04-10T13-13-16.xlsx)

SFJohnson24:
added unittest with scalar and lists containing NaN

gerrycampion (approved):

PR #1131, opened 2025-04-11
docker image gitaction

SFJohnson24:
This creates a gitaction that updates the image on dockerhub.
local dockerfile always grabs latest but if they grab the image from dockerhub, latest is not being updated.  This adds a automated step when a release is published that checks if the current release tag is also the tag version on latest.  If it is, it builds and pushes to dockerhub both as latest, and as the specific version of the release.

testing this locally might be a challenge.  What I would do is merge it, run the action (it will check the current release and notice it is not latest but it should go through the steps of getting the release tags.  except the image build and upload which wont be used until we do a latest release).   The build is taken exactly from the dockerhub docs (links to this url: https://github.com/marketplace/actions/build-and-push-docker-images)

### DIFF
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -1,0 +1,61 @@
+name: Build and Push Docker Image
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+jobs:
+  check-if-latest:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


RamilCDISC:
I ran the action in github, the action failed. @SFJohnson24 could you please have a look.

SFJohnson24:
see #1132

PR #1132, opened 2025-04-11
Docker

SFJohnson24:
there was a syntax error, causing the last run to fail @RamilCDISC 
see: https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/14410095271

note: GITHUB_REF value is refs/heads/docker which is the branch name, when this is triggered on a release, it will be the version and trigger the build


RamilCDISC (approved):
Code duplicate removed. Updated as required

PR #1137, opened 2025-04-15
end2end testing for engine

RamilCDISC:


SFJohnson24 (changes requested):
See my comment in the editor ticket as well @RamilCDISC
Both scripts need a few more additional steps but they look great so far

.github/test/selenium_test_editor.py
```diff
@@ -0,0 +1,94 @@
...
+try:
+    print("Opening Rule Editor site...")
+    driver.get(RULE_EDITOR_URL)
+
+    print("Clicking the search field for rule search...")
+    rule_search_field = wait.until(
+        EC.element_to_be_clickable((By.XPATH, '//*[@id="mui-10"]'))
+    )
+    rule_search_field.click()
+    rule_search_field.send_keys("CG0002")
```
SFJohnson24, line 40: CG0002 is not published--can you pick a published rule for the test as this rule and its logic will likely change

.github/test/selenium_test_editor.py
```diff
@@ -0,0 +1,94 @@
...
+                By.XPATH,
+                '//*[@id="tabpanel-1"]/div[5]/div[2]/div/div/div/div/label/input',
+            )
+        )
+    )
+
+    file_path = os.path.abspath(".github/test/unit-test-coreid-CG0002-negative.xlsx")
+    print(f"Uploading file: {file_path}")
+    file_input.send_keys(file_path)
+    print("Error result shows up")
```
SFJohnson24, line 80: we want to be looking for positive and negative rule results showing up in the DOM--not an error

SFJohnson24 (changes requested):

.github/test/selenium_test_editor.py
```diff
@@ -0,0 +1,187 @@
...
+                        "dataset": "ie.xpt",
+                        "error": "Rule format error",
+                        "message": "Rule contains invalid operator",
+                    }
+                ],
+            }
+        ],
+        "LB": [
+            {
+                "executionStatus": "execution_error",
```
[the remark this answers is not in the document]
RamilCDISC: updated

SFJohnson24 (approved):
reviewed code and action run.
looks good @RamilCDISC

PR #1139, opened 2025-04-16
updated validate to run tig standard

SFJohnson24:
This PR allows you to run TIG in the list rules command as well as running validations with just tig 1-0 sdtm specified without any specific rules

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,6 +264,10 @@ To obtain an api key, please follow the instructions found here: <https://wiki.c
 
       `python core.py list-rules -s sdtmig -v 3-4`
 
+- list rules for integrated standard (substandard: "SDTM", "SEND", "ADaM", "CDASH"):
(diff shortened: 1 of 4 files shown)


RamilCDISC (approved):
The list rules function is updated to display the TIG rules. The rules can be filtered out based on their substandard too. The validation was done by:

1. Run the list-rules command with TIG standard and different combination of version and substandard
2. Run the list-rules command with SDTMIG and different combination of versions
3. Run the list-rules command with SENDIG and different combination of versions
4. Run the list-rules command with SENDIG-AR and different combination of versions
5. Run the list-rules command with SENDIG-DART and different combination of versions
6. Run the list-rules command with SENDIG-GENETOX and different combination of versions
7. Run the list-rules command with USDM and different combination of versions

PR #1140, opened 2025-04-17
update logic

SFJohnson24:
this PR updates the include_split_dataset logic.  Currently if this flag is set to true, we do not get to the logic in the rule processor to use it to determine if split datasets apply--we skip the rule before https://github.com/cdisc-org/cdisc-rules-engine/blob/5258f7858f903fd600e28acaa8ccb7d4e8c2160e/cdisc_rules_engine/rules_engine.py#L103 due to this logic.  This PR adds a check for this flag in that rule to allow for split datasets to also be validated in addition to their parent when this flag is present in the rule.  

to validate:
[Datasets2.json](https://github.com/user-attachments/files/19799241/Datasets2.json)
[Rule_underscores2.json](https://github.com/user-attachments/files/19799242/Rule_underscores2.json)
This rule currently produces a positive but is negative dataset 2 for cg0333.

### DIFF
--- a/PYPI.md
+++ b/PYPI.md
@@ -43,7 +43,7 @@ def load_rules_cache(path_to_rules_cache):
     return cache
 ```
 
-Rules in this cache can be accessed by standard and version using the get_rules_cache_key function.
+Rules in this cache can also be accessed by standard and version using the get_rules_cache_key function.
 
 ```python
 from cdisc_rules_engine.utilities.utils import get_rules_cache_key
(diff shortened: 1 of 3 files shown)


RamilCDISC (approved):
The PR adds the support in rule to pass a check for split dataset. If the check "include_split_dataset "is passed the engine will process the split datasets even if the parents dataset is missing. By default the behavior of engine would be to skip such a dataset. The validation for the issue was done by:

1. Validating the use of "include_split_datasets" with a dataset where parent is MISSING, in the editor
2. Validating the use of "include_split_datasets" with a dataset where parent is PRESENT, in the editor
3. Validating the use of "include_split_datasets" with a dataset where parent is MISSING, in the engine (terminal)
4. Validating the use of "include_split_datasets" with a dataset where parent is PRESENT, in the engine (terminal)
5. Validate against unspilt datasets

The engine only supports TRUE as value for the "include_split_datasets". Even a FALSE would be wrong value for this and would cause an error.

PR #1142, opened 2025-04-18
Use case

SFJohnson24:


RamilCDISC (approved):
The PR adds the feature to the engine to account for use cases in the rules. A dataset is only processed if the use case covers the dataset. The validation was done by the following test scenarios:

1. Validating all use case with SDTM as substandard for TIG standard.
2. Validating all use case with send as substandard for TIG standard.
3. Validating when the use case have an INCORRECT VALUE. 
4. Validating the proper running of SDTMIG standard for rules.
5. Validating the proper running of SEND standard for rules.
6. Validating when the use case is MISSING from the rule.

PR #1145, opened 2025-04-21
Debugging for cache keyerror

gerrycampion:
This simply adds a lock around writes to the in memory cache. This resolves the problem where multiple processes are writing to the same cache at the same time, causing some rules to get skipped randomly.
Evidence that this is working can be viewed by seeing the test suite validation successes from today (4/23): https://github.com/cdisc-org/cdisc-rules-engine/actions/workflows/test_suite.yml

### DIFF
--- a/cdisc_rules_engine/services/cache/in_memory_cache_service.py
+++ b/cdisc_rules_engine/services/cache/in_memory_cache_service.py
@@ -7,7 +7,6 @@ from cdisc_rules_engine.interfaces import (
 from cdisc_rules_engine.models.dataset import DatasetInterface
 from cachetools import LRUCache
 import psutil
-from cdisc_rules_engine.services import logger
 from multiprocessing import Lock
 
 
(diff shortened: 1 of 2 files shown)


RamilCDISC (approved):
The PR adds lock for cache operations in memory, where multiple processes was causing a race condition. This race condition was causing inconsistent outputs. The PR was validated by running the rules (mentioned in AC) for both DASK and PANDAS dataset implementations. 

1. Run rule CORE-000239 with DASK dataset.
2. Run rule CORE-000239 with PANDAS dataset.
3. Run rule CORE-000363 with DASK dataset.
4. Run rule CORE-000363 with PANDAS dataset
5. Run rule CORE-000370 with DASK dataset.
6. Run rule CORE-000370 with PANDAS dataset
7. Run rule CORE-000468 with DASK dataset.
8. Run rule CORE-000468 with PANDAS dataset
9. Run rule CORE-000291 with DASK dataset.
10. Run rule CORE-000291 with PANDAS dataset
11. Run rule CORE-000356 with DASK dataset.
12. Run rule CORE-000356 with PANDAS dataset.

All these rules showed inconsistent output before but now they run successfully. The result from PANDAS and DASK both are similar.

PR #1148, opened 2025-04-22
Variable value

SFJohnson24:
This PR refactor dataset builders to eliminate build_split_datasets code.  It adds a default implementation of build_split_datasets in BaseDatasetBuilder which replaces the abstract method.  This method will be inherited by the children and the child method's will override this parent method when the builder requires different build_split_dataset logic.
  It also adds two new rule types"Value Check with Dataset Metadata" and "Value Check with Variable Metadata"

to test:
test cg0019 -- still has 4 issues raised.  
Review the test validate--this new logic should not change how split datasets were executed makes all builders able to handle splits
no rules exists for the new rule types--make a rule with the two new rule types and experiment with the builders.


RamilCDISC (changes requested):
Two Things 
1. When we use these new rule types in editor they get yellow highlighting. We need to add these for editor too.

2. When executing SEND1 with the logic given in connected issue by elijanssens. The editor throws an error which reads as follow:
`{1 item
"CE":[1 item
0:{6 items
"executionStatus":"execution_error"
"dataset":"ae.xpt"
"domain":"CE"
"variables":[]0 items
"message":"rule execution error"
"errors":[1 item
0:{3 items
"dataset":"ae.xpt"
"error":"An unknown exception has occurred"
"message":"'PandasDataset' object has no attribute 'items'"
}
]
}
]
}`

The rule used: 

> Authorities:
>   - Organization: CDISC
>     Standards:
>       - Name: SENDIG
>         References:
>           - Citations:
>               - Cited Guidance: SDTM datasets are normally named to be consistent with the
>                   domain code.
>                 Document: SENDIG v3.1
>                 Section: 4.1.4
>             Origin: SENDIG Conformance Rules
>             Rule Identifier:
>               Id: 'SEND1'
>               Version: '1'
>             Version: '5'
>         Version: '3.1'
> Check:
>   all:
>     - name: dataset_name
>       operator: is_not_contained_by
>       value: $domain_value
> Core:
>   Id: CDISC.SENDIG.SEND1
>   Status: Draft
>   Version: '1'
> Description: Raise an error when the filename is not equal to the value of DOMAIN.
> Outcome:
>   Message: The filename is not equal to the 2-character domain code.
>   Output Variables:
>     - DOMAIN
> Rule Type: Value Check with Dataset Metadata
> Scope:
>   Classes:
>     Include:
>       - ALL
>   Domains:
>     Include:
>       - ALL
> Sensitivity: Dataset
> 
Dataset was negative dataset from sharepoint

SFJohnson24:
1. is your editor pointed at the correct schema when you run it locally?
2.  that rule logic doesnt make a ton of sense--it was hypothetical.  It also doesnt have either of the rule types in this PR.
[Rule_underscores.json](https://github.com/user-attachments/files/20004151/Rule_underscores.json)
this would be the rule logic

RamilCDISC (approved):
The PR adds a new rule type to the engine for getting dataset and variable metadata. The PR was validated by following test scenarios:

1. Running the engine against CG0019 for validating the changes to the dataset builders.
2. Running the engine against postive datasets with "Value Check with Dataset Metadata". 
3. Running the engine against negative datasets with "Value Check with Dataset Metadata".
4. Running the egnine against the positive datasets with "Value Check with Variable Metadata".
5. Running the engine against the negative datasets with "Value Check with Variable Metadata"
6. Running engine using the rule SEND1/TIG0037 with their positive and negative samples
7. Running the above mentioned tests in the editor for functionality check against for editor.

The rule FB1501 and FB1502 are still not complete and require new operators. The operators have been requested in separate ticket. The engine will be validated for those rules in that ticket. [https://github.com/cdisc-org/cdisc-rules-engine/issues/1181](url)

PR #1150, opened 2025-04-23
update tests

SFJohnson24:


RamilCDISC (approved):
Tests are updated so that the use case check is skipped. The reason to skip is that these tests do not require to validate the use case checks. Tests for use case checks are already available in test_rule_processor.py. As all the tests pass it shows that nothing is broken

Issue #1059, opened 2025-02-28
CG0073 (CORE-000364): only applicable to SDTMIG 3.2 executing for SDTMIG 3.4


mhungria:
CG0073 (CORE-000364): only applicable to SDTMIG 3.2 executing for SDTMIG 3.4.

Potentially related to https://github.com/cdisc-org/cdisc-rules-engine/issues/1054

mhungria:
@nickdedonder , is it possible to expedite a new CLI release to include the fix for https://github.com/cdisc-org/cdisc-rules-engine/pull/1114 as well as resolution for this issue?

nickdedonder:
@mhungria similar to #1054, when running this with the CLI for SDTMIG 3.3, this rule is not present in my Rules report so it hasn't been executed.

mhungria:
Please see note on https://github.com/cdisc-org/cdisc-rules-engine/issues/1054.

mhungria:
@SFJohnson24, thanks for addressing this issue!

mhungria:
@SFJohnson24, team:

I just tested with the v0.10.0 release and I got a clean result without this issue.

Thanks

SFJohnson24:
Glad to hear it @mhungria

PR #1151, opened 2025-04-23
local rule std check

SFJohnson24:
this PR adds a check for standard and version for rules given via -lr flag.  This ensures that rules given with the local flag outside of the cache match the standard and version given with the -s and -v flags specified with in the rule

see also: https://github.com/cdisc-org/cdisc-rules-engine/issues/1059

### DIFF
--- a/scripts/script_utils.py
+++ b/scripts/script_utils.py
@@ -345,24 +345,56 @@ def load_and_parse_local_rule(rule_file: str) -> dict:
         return None
 
 
+def rule_matches_standard_version(rule, standard, version, substandard=None):
+    normalized_version = version.replace("-", ".")
+    for standard_info in rule["standards"]:
+        std_name = standard_info.get("Name", "")
+        std_version = standard_info.get("Version", "")
+        std_substandard = standard_info.get("Substandard")
+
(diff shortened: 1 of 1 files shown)


mhungria:
@SFJohnson24, thanks for addressing this issue!

I'll test again as soon as a new CLI release is available.

RamilCDISC (approved):
The PR adds checks for standard and version while executing a rule. If the mentioned standard and version match the rule only then the rule is executed. The validation was done by following test cases:

1. Running SDTMIG rule using -r flag when the standard and version is matching the rule.
2. Running SDTMIG rule using -lr flag when standard and version is matching the rule.
3. Running SDTMIG rule using -r flag when the standard do not match the rule but version do.
4. Running SDTMIG rule using -lr flag when the standard do not match the rule but version do. 
5. Running SDTMIG rule using -r flag when the version do not match the rule but standard do. 
6. Running SDTMIG rule using -lr flag when the version do not match the rule but standard do.
7. Running all the same cases as above but with standard set as USDM, ADAM, TIG.
8. Running validation with cases where the rule has more than one standard.

PR #1152, opened 2025-04-23
updated rule tester actions

gerrycampion:
This update:
- allows us to manually publish the rule tester endpoint to dev or prod
- will auto publish the endpoint to prod on push to main (as before)
- updates the version of python from 3.9 to 3.10 to be consistent with everything else

To deploy to dev manually, you can:
- go here: https://github.com/cdisc-org/cdisc-rules-engine/actions/workflows/rule-tester.yml
- select the prod-rule-tester branch
- leave the env as "DEV"

The dev endpoint is already live in the dev version of rule editor here:  https://icy-flower-095494b10-dev.centralus.azurestaticapps.net/


RamilCDISC (approved):
The updated rule tester action works fine as can be seen in the ran actions and updated dev endpoint

PR #1158, opened 2025-05-01
Relrec fixed with idvarvals populated

gerrycampion:
Logic was added to fix the case where IDVARVAL columns were populated in RELREC.

Refer to the attached issue for test data and rule. You can also refer to the new unit test.

### DIFF
--- a/cdisc_rules_engine/utilities/data_processor.py
+++ b/cdisc_rules_engine/utilities/data_processor.py
@@ -140,6 +140,16 @@ class DataProcessor:
         return reference_data
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


RamilCDISC (approved):
This PR fixes the RELREC merge which was breaking a rule. The PR was validated by:

1. Running the editor locally and testing the engine against the first test data attached in the AC.
2. Running the editor locally and testing the engine against the second test data attached in AC.
3. Running the test with first test data directly in engine instead of editor.
4. Running the tests with second test data directly in engine instead of editor.

gerrycampion:
replied in jira https://jira.cdisc.org/browse/CORERULES-150

PR #1165, opened 2025-05-05
update docker trigger

SFJohnson24:
The docker image build gitaction triggers on release but it needs to be delayed as it uses the github api to grab the new release and see if it is latest.  It cannot run in parallel so I alterred its run conditions to trigger after the release action successfully runs
https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/14836565275


RamilCDISC (approved):
Merging to test the action

RamilCDISC:
Docker Image build works fine

PR #1166, opened 2025-05-05
added get_parent_model_column_order error handling

gerrycampion:
[CG0314.yml.txt](https://github.com/user-attachments/files/20040761/CG0314.yml.txt)
[unit-test-coreid-CG0314-negative-with-ae.xlsx](https://github.com/user-attachments/files/20040762/unit-test-coreid-CG0314-negative-with-ae.xlsx)
[unit-test-coreid-CG0314-negative-without-ae.xlsx](https://github.com/user-attachments/files/20040763/unit-test-coreid-CG0314-negative-without-ae.xlsx)

The test rule and test data are linked above. The data with ae should run and produce a report. The data without ae should show the following error:
> cdisc_rules_engine.exceptions.custom_exceptions.DomainNotFoundError: Operation get_parent_model_column_order requires parent Domain AE but Domain not found in datasets

### DIFF
--- a/cdisc_rules_engine/operations/parent_library_model_column_order.py
+++ b/cdisc_rules_engine/operations/parent_library_model_column_order.py
@@ -1,3 +1,4 @@
+from cdisc_rules_engine.exceptions.custom_exceptions import DomainNotFoundError
 from cdisc_rules_engine.operations.library_model_column_order import (
     LibraryModelColumnOrder,
 )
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
The PR adds error handling for get_parent_model_column_order. The issue was validated by using a dataset that had the parent available and with the case where it as unavailable. For unavailable case the error message was correctly reported.

PR #1167, opened 2025-05-06
Within

SFJohnson24:
forgot to update the readme.  The schema already has the optional parameter within for operators so no change needed there


RamilCDISC (approved):
The new logic is explained well and read me is now up to date

Issue #434, opened 2023-06-07
New operator needed for comparing sorted arrays


JozefAerts:
We can now obtain the array with the ordered variable names both from the dataset (using "get_column_order_from_dataset") and from the CDISC Library (using "get_column_order_from_library"). The former should be a subset of the latter.
It however looks that we don't have an operator to compare these, to see whether the array from the dataset is in the same order as the array from the CDISC Library.
We need this for rule SEND48 (already present in the rule editor), testfiles on SharePoint at:
https://cdisc.sharepoint.com/sites/CORERules/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2FCORERules%2FShared%20Documents%2FunitTesting%2FSENDIG%2FSENDIG048&viewid=ecd009a9%2Db73b%2D415b%2D9e41%2Df9d04a69edf4

Example of an array (for VS) delivered by the CDISC Library:
![grafik](https://github.com/cdisc-org/cdisc-rules-engine/assets/25853332/31a1037d-ac09-475c-b0f8-239839688953)
Example of an array (for VS) for the dataset:
![grafik](https://github.com/cdisc-org/cdisc-rules-engine/assets/25853332/b3601100-8717-406a-a91a-d9dcf8b24d06)
The second needs to be a subset of the first, in the same order.
The new operator should report an error when the second is not in the same order, or an element of it is not in the first.

nickdedonder:
also CORERULES-51 (CG0664)

marisawyckmans:
CG0664 links to TIG0698, so this issue also links to CORERULES-9843 (TIG0698). @nickdedonder could you add the TIG label? Thanks

PR #1172, opened 2025-05-07
Sorted arrays

SFJohnson24:
This adds the requested new operator.  Of note: 
image
the positive data has a VSLOBXFL which is not part of the library metadata so it is returning an error.  The operator is functioning properly and as expected.
[Rule_underscores.json](https://github.com/user-attachments/files/20091866/Rule_underscores.json)
[DatasetsN.json](https://github.com/user-attachments/files/20091867/DatasetsN.json) negative
[DatasetsP.json](https://github.com/user-attachments/files/20091868/DatasetsP.json) positive

### DIFF
--- a/cdisc_rules_engine/check_operators/dataframe_operators.py
+++ b/cdisc_rules_engine/check_operators/dataframe_operators.py
@@ -1600,3 +1600,36 @@ class DataframeType(BaseType):
             return len(target_set.intersection(comparator_set)) == 0
 
         return self.value.apply(check_no_shared_elements, axis=1).all()
+
+    @log_operator_execution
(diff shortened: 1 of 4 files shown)


SFJohnson24:
zenhub is acting up, it will not let me link #434 to it

RamilCDISC (approved):
The PR adds an operator for comparing the dataset column order and library column order. It was validated by the following test cases.

1. Running a validation using a dataset where the number of column are equal to and similar to the library.
2. Running a validation using a dataset where the number of the columns are less than the library.
3. Running validation with standard set to TIG. 
4. Running the validation with standard set to SDTIMG. 
5. Running the validation with standard set to SEND.

All validations pass.

PR #1178, opened 2025-05-12
Python

SFJohnson24:
This PR increments python and other dependencies to be compatible with python--mainly numpy, dask, pandas, flake8, etc.

The new versions of these dependencies changed a fair bit of underlying functionality--pandas is more strict on scalar and vector comparisons, dask requires metadata or asbool() more often as it used to infer metadata but now needs to be told expected output more often for its lazy execution.  The new version of dask updated its serialization strategy and was less stringend on serialization requirement.  It now does not allow the original validator object to be serialized.  I created a workaround that makes a lookup table + function.  This might benefit from caching the lookup table--we should make a ticket later if we note this is a bottleneck.

release: https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/15044795975


RamilCDISC (approved):
The PR updates the python version for the engine from 3.10 to 3.12. The PR was validated by following test scenarios:


1. Confirming version updated in CI/CD pipeline and all is working fine.
2. Confirming version updated in the release workflow and all working fine.
3. Confirming version updated in the related documentations.
4. Confirming version updated in toml file.
5. Confirming the outdated version of packages in updated in requirements file.
6. Confirming the tests are updated to be coherent with latest libraries version.
7. Passing of all unit tests.
8. Passing of regression tests.
9. Successful creation and working of release.

PR #1183, opened 2025-05-20
Fix TestRule

ASL-rmarshall:
PR #1176 included a change to `TestRule/__init__.py` to force population of the cache with all published CT packages.  Population of the cache uses a `cache_populator`, but this is only created if `standards_data` or `codelists` are specified with the test data.  However, `standards_data` and `codelists` are not specified for most USDM test data so the call to the undefined `cache_populator` caused an unhandled error, which is reported in the editor as "500 Internal Server Error".

 Creation of the `library_service` and `cache_populator` have therefore been made unconditional so that the `cache_populator` is always defined.

### DIFF
--- a/TestRule/__init__.py
+++ b/TestRule/__init__.py
@@ -66,9 +66,10 @@ def main(req: func.HttpRequest, context: func.Context) -> func.HttpResponse:  # 
         standard_substandard = standards_data.get("substandard")
         codelists = json_data.get("codelists", [])
         cache = InMemoryCacheService()
+        library_service = CDISCLibraryService(api_key, cache)
+        cache_populator: CachePopulator = CachePopulator(cache, library_service)
+        asyncio.run(cache_populator.load_available_ct_packages())
         if standards_data or codelists:
-            library_service = CDISCLibraryService(api_key, cache)
-            cache_populator: CachePopulator = CachePopulator(cache, library_service)
             if standards_data:
                 asyncio.run(
                     cache_populator.load_standard(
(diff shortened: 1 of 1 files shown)


TestRule/__init__.py
```diff
@@ -66,17 +66,17 @@ def main(req: func.HttpRequest, context: func.Context) -> func.HttpResponse:  #
         standard_substandard = standards_data.get("substandard")
         codelists = json_data.get("codelists", [])
         cache = InMemoryCacheService()
+        library_service = CDISCLibraryService(api_key, cache)
+        cache_populator: CachePopulator = CachePopulator(cache, library_service)
+        asyncio.run(cache_populator.load_available_ct_packages())
         if standards_data or codelists:
```
RamilCDISC, line 72: I believe this check was helpful earlier, but it might not be necessary anymore, since we’re already validating standards_data further down. What do you think about removing it?

RamilCDISC (changes requested):

TestRule/__init__.py
ASL-rmarshall: When you say "validating standards_data further down" do you mean the `if standards_data` on the next line? If so, I think the `if standards_data or codelists` is still needed because we want to populate the cache with codelists only if either standards or codelists are specified in the rule metadata.  The `if standards_data` on the next line is there to ensure that the cache is populated with the definition of the standard only when a standard is specified.

If would be possible to remove the `if standards_data or codelists`, but to get the same effect you'd then have to duplicate the call to populate the cache with codelists but ensure that it's not called twice - e.g.:
```python
        if standards_data:
            asyncio.run(
                cache_populator.load_standard(
                    standard, standard_version, standard_substandard
                )
            )
            asyncio.run(cache_populator.load_codelists(codelists))
        elif codelists:
            asyncio.run(cache_populator.load_codelists(codelists))
```
I think it's more succinct the way it is currently, so I don't think the `if standards_data or codelists` should be removed (unless I misunderstood your comment/question).

Issue #1177, opened 2025-05-09
Rule blocked: CORERULES-9552 - additional requirements for codelist_terms operation


ASL-rmarshall:
**Links to related JIRA Tickets**
- https://jira.cdisc.org/browse/CORERULES-9552
- https://jira.cdisc.org/browse/CORERULES-9609
- https://jira.cdisc.org/browse/CORERULES-9647
- https://jira.cdisc.org/browse/CORERULES-9653
- https://jira.cdisc.org/browse/CORERULES-9654
- https://jira.cdisc.org/browse/CORERULES-9655
- https://jira.cdisc.org/browse/CORERULES-9656
- https://jira.cdisc.org/browse/CORERULES-9659
- https://jira.cdisc.org/browse/CORERULES-9660
- https://jira.cdisc.org/browse/CORERULES-9661
- https://jira.cdisc.org/browse/CORERULES-9662
- https://jira.cdisc.org/browse/CORERULES-9663
- https://jira.cdisc.org/browse/CORERULES-9664
- https://jira.cdisc.org/browse/CORERULES-9665
- https://jira.cdisc.org/browse/CORERULES-9666
- https://jira.cdisc.org/browse/CORERULES-9671
- https://jira.cdisc.org/browse/CORERULES-9672
- https://jira.cdisc.org/browse/CORERULES-9673
- https://jira.cdisc.org/browse/CORERULES-9678
- https://jira.cdisc.org/browse/CORERULES-9679
- https://jira.cdisc.org/browse/CORERULES-9683
- https://jira.cdisc.org/browse/CORERULES-9684
- https://jira.cdisc.org/browse/CORERULES-9685
- https://jira.cdisc.org/browse/CORERULES-9686
- https://jira.cdisc.org/browse/CORERULES-9687
- https://jira.cdisc.org/browse/CORERULES-9688
- https://jira.cdisc.org/browse/CORERULES-9689
- https://jira.cdisc.org/browse/CORERULES-9690
- https://jira.cdisc.org/browse/CORERULES-9691
- https://jira.cdisc.org/browse/CORERULES-9692
- https://jira.cdisc.org/browse/CORERULES-9693
- https://jira.cdisc.org/browse/CORERULES-9696
- https://jira.cdisc.org/browse/CORERULES-9950
- https://jira.cdisc.org/browse/CORERULES-9959
- https://jira.cdisc.org/browse/CORERULES-9962
[...]

Issue #1170, opened 2025-05-07
CT cache submission_lookup issue - submission values are not unique


ASL-rmarshall:
**Describe the bug**
At the moment, a `submission_lookup` dictionary is created within each cached CT file. This dictionary uses the CT submission value a key, with values containing `codelist` code and `term` code (which is "N/A" for codelists), and is (intended to be) populated with all codelists **and** all terms.

However, submission values are not unique:
- A codelist may have the same submission value as a term (e.g., "TTYPE" is the submission value for both codelist "C66739" and term "C49660").
- Two (or more) terms may have the same submission value (e.g., "0" is the submission value for 53 terms in the SDTM CT 2024-09-27).

[Note that submission values for codelists are unique: no two codelists within the same version of CT for the same standard will have the same submission value]

It appears that the `submission_lookup` ends up being populated with the last occurrence of a submission value, regardless of whether it's for a codelist or for a term.  This means that:
- There may be no entry in `submission_lookup` for any codelist whose submission value matches the submission value of a (later) term.
- There will be only one entry in `submission_lookup` for any term submission value that is used for multiple terms.

**Expected behavior**

I think the `submission_lookup` dictionary needs to be updated so that the value is a list of the `codelist` code and `term` code combinations for all occurrences of the submission value key.  Any functionality that references `submission_lookup` should then be updated parse the value list - either to find codelist entries or to process each of the term entries.

SFJohnson24:
<img width="678" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDN2Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0d64dbe7edaeb57f19766e6abae5463110600495/image.png" alt="image.png" />
We can add codelist and term submission value to the lookup key (see lower example.  This could be stringified or use a delimiter to separate codelist/term) or we could create a list for terms with the codelists and term CCodes, then iterate through that list (see above).  Either is viable.

ASL-rmarshall:
[...]

Issue #939, opened 2024-12-09
Rule blocked: CORERULES-9609 - codelist_terms operation does not retrieve correct terms


ASL-rmarshall:
- A .txt file with the JSON containing the full request: [Request.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNkFUQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c5d377eae0d6186dce4dc5d88c079326da2c02c4/Request.txt)

**Links to related JIRA Tickets**
- https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-9609

**Rule Information**
- **Standard**: USDM
- **Rule ID**: DDF00015
- **Rule Description**: A study version's study phase must be specified according to the extensible Trial Phase Response (C66737) SDTM codelist (e.g. an entry with a code or decode used from the codelist should be consistent with the full entry in the codelist).

**Describe the bug**
The `codelist_terms` operation does not retrieve the correct set of terms for the specified codelist.  In this example rule, the specified codelist is "TPHASE". The `_execute_operation` method tries to retrieve the C-code for the specified codelist by looking up the submission value ("TPHASE") in `submission_lookup`, which appears to be created from the `submission_lookup` property of the first controlled terminology package that's found.  

However, the `submission_lookup` property of the CT package appears to have been created based on the assumption that there is a single C-code for each unique submission value. Unfortunately, this is an invalid assumption - there may be as many as 65 different C-codes for any given submission value. Also, the same value might be the submission value for both a codelist and a term, which is the case in this example:
| Code   | Codelist Code | Codelist Name                     | CDISC Submission Value | CDISC Definition                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
| ------ | ------------- | --------------------------------- | ---------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| C66737 |               | Trial Phase Response              | TPHASE                 | A terminology codelist relevant to a step in the clinical research and development of a therapy from initial clinical trials to post-approval studies.                                                                                                                                                                                                                                                                                                                                                 |
[...]

PR #1199, opened 2025-06-05
map, codelist_extensible and codelist_terms for usdm

gerrycampion:
The following rule and data show examples of how to use the new `map` operation and the modified `codelist_extensible` and `codelist_terms` operations

[Rule.yaml.txt](https://github.com/user-attachments/files/21023792/Rule.yaml.txt)
[Datasets.xlsx](https://github.com/user-attachments/files/20713782/Datasets.xlsx)

You can also refer to the new documentation examples, schema, and unit tests. I'm not 100% sure this covers everything needed in #1177 so please let me know what is missing.

Some differences between the suggested implementation and this implementation:
- `map` operation allows the use of a lookup table for codelists until it becomes available in the library. if we don't use this, it just means we need more rules. But I think ultimately, metadata-driven codelist checks with a single rule is the approach we should take.
- instead of providing lists of terms, merges are done directly in the codelist operations so that each source record has the matching codelist extensibility or codelist term code/decode (depending on which source param is provided).  This means that you should be able to use `empty` check operator instead of `is_contained_by` check operator. I think this provides better simplicity, performance, and specificity as to exactly which term matches with each record's codelist and version.


resources/schema/Operations.json
```diff
@@ -416,25 +426,26 @@
...
-          "MRCT",
-          "PROTOCOL",
-          "QRS",
-          "QS-FT",
-          "SDTM",
-          "SEND",
-          "TMF"
-        ]
-      },
+      "items": { "$ref": "#/properties/ct_package_type" },
```
ASL-rmarshall: This reference needs to be prefixed with the file name to work (i.e. `"$ref": "Operations.json#/properties/ct_package_type"`).

cdisc_rules_engine/models/library_metadata_container.py
```diff
@@ -59,3 +66,55 @@ def get_all_ct_package_metadata(self):
 
     def set_ct_package_metadata(self, key, value):
         self._ct_package_metadata[key] = value
+
+    def _load_ct_package_data(self, ct_package_type: str, version: str):
+        ct_package_version = f"{ct_package_type}-{version}"
+        ct_package_data = self.get_ct_package_metadata(ct_package_version)
+        if ct_package_data is None:
+            file_name = f"{ct_package_version}.pkl"
+            with open(join(self._cache_path, file_name), "rb") as f:
```
ASL-rmarshall: Validation fails with the following error when there is no file in the local cache for the specified `ct_package_type` and `version`:
```
"error":"An unknown exception has occurred"
"message":"[Errno 2] No such file or directory: 'ddfct-2023-12-16.pkl'"
```
I think there needs to be a check for the existence of the file before the attempt to open it.  If there is no file, I guess `ct_package_data` would be returned as `None` and whichever methods use the returned value would need to deal with the possibility of a returned `None`.

resources/schema/Operations.md
ASL-rmarshall, line 1: The first example for `codelist_terms` in Operations.md should be in YAML format, not JSON format.

cdisc_rules_engine/operations/codelist_terms.py
```diff
@@ -1,11 +1,70 @@
...
+                left_on=(
+                    self.params.ct_version,
+                    self.evaluation_dataset[left_on].str.lower(),
+                ),
+                right_on=("version", codelist[right_on].str.lower()),
+                how="left",
+            )
+        return result[target]
+
+    def _handle_single_version(self) -> pd.Series:
```
ASL-rmarshall, line 72: This PR has been attached to #939 and #1170, but it does not address to issue of non-unique submission values result in an incomplete `submission_lookup` dictionary (which is still used in `_handle_single_version`).  USDM terminology checking can be correctly applied using `_handle_multiple_versions` but other types of terminology checking - e.g. SDTM, SEND, etc. - will still be impacted by #939 and #1170.  It would probably be best to unlink this PR from #939 and #1170, and have a separate PR to address the non-unique submission value problem.

ASL-rmarshall (changes requested):

cdisc_rules_engine/operations/codelist_terms.py
gerrycampion: unlinked

cdisc_rules_engine/models/library_metadata_container.py
gerrycampion: Continue gracefully when missing ct package type or version. We will leave it to the rules to handle these conditions appropriately.

cdisc_rules_engine/models/library_metadata_container.py
```diff
@@ -71,10 +71,16 @@ def _load_ct_package_data(self, ct_package_type: str, version: str):
...
         if ct_package_data is None:
-            file_name = f"{ct_package_version}.pkl"
-            with open(join(self._cache_path, file_name), "rb") as f:
-                ct_package_data = load(f)
-                self.set_ct_package_metadata(ct_package_version, ct_package_data)
+            file_name = join(self._cache_path, f"{ct_package_version}.pkl")
+            try:
+                with open(file_name, "rb") as f:
+                    ct_package_data = load(f)
+                    self.set_ct_package_metadata(ct_package_version, ct_package_data)
```
ASL-rmarshall: Is this necessary here if it's done on line 83 anyway?

resources/schema/Operations.md
ASL-rmarshall: Looks good now.

cdisc_rules_engine/models/library_metadata_container.py
```diff
@@ -86,41 +86,39 @@ def _load_ct_package_data(self, ct_package_type: str, version: str):
...
-                        "term_code": term["conceptId"],
-                        "term_value": term["submissionValue"],
-                    }
-                    for codelist_code, codelist in ct_package_data.items()
-                    if "terms" in codelist
-                    for term in codelist["terms"]
-                ]
-            )
+            for codelist_code, codelist in ct_package_data.items():
+                for term in codelist.get("terms", []):
```
ASL-rmarshall: This throws an error when the item value, `codelist`, is not a dictionary (e.g., for `ct_package_data["package"]` which has a string value of "ddfct-2023-12-15"). There probably needs to be type checking or something.  Possibly consider for line 98 too.

ASL-rmarshall (changes requested):

cdisc_rules_engine/models/library_metadata_container.py
gerrycampion: forgot to delete

PR #1201, opened 2025-06-09
Cg0371

SFJohnson24:
this PR eliminated is_valid_relationship and is_valid_reference and their logical complements.
It also eliminates the relationship data passed to operators from the preprocessor.  Relationship data merges are already handled in the datasetpreprocessor _merge_datasets() function.

it adds the ability to match child to parent for relationship data with the 'child' argument under match dataset
it also adds the value_is_reference to equal to and not equal to, which allows for a check of the value referenced in the column such as IDVAR.

to test run cg0371:
[Rule_underscores.json](https://github.com/user-attachments/files/20731071/Rule_underscores.json)
 This is my version of the rule, it would need match conditions for each supp written into the rule.
[DatasetsN.json](https://github.com/user-attachments/files/20717040/DatasetsN.json)
[DatasetsP.json](https://github.com/user-attachments/files/20717041/DatasetsP.json)


SFJohnson24:
match dataset  is a space, not an underscore
I added the type_insensitive property

RamilCDISC (approved):
The PR removes the is valid relationship and is valid reference operators and adds the functionality directly to the preprocessor. The updated code was validated by:

1. Reviewing the updated code for any unwanted comments and code.
2. Reviewing the logic to see if there is any flaw or edge case that is not handled.
3. Reviewing the project code to confirm the removed operators reference is also removed from everywhere. 
4. Reviewing updated unit tests.
5. Validation of passing of all unit tests.
6. Running validation in the engine using terminal against positive and negative dataset.
7. Validating the updated operators are added to the schema and editor accepts the new properties. (bug was found and fixed)

PR #1207, opened 2025-06-13
Get the preview URL only from the action on the main branch

RamilCDISC:
This updates the end2end testing to use the last deploy link from the action on the main branch of rule editor instead of just the last run action. It also updates the python version to 3.12

EDIT: Now the end2end testing run on rule tester. When you sent to test a new branch:
1. Run the rule tester workflow to deploy it on dev.
2. Re-run the preview workflow to verify.


.github/workflows/test_rule_editor_preview.yml
```diff
@@ -15,10 +15,13 @@ jobs:
...
         with:
-          python-version: 3.9
+          python-version: 3.12
+          
+      - name: Ensure setuptools is installed
+        run: python -m ensurepip --upgrade && python -m pip install --upgrade pip setuptools
 
       - name: Install dependencies
-        run: pip install requests selenium webdriver-manager pyperclip==1.9.0 selenium-wire blinker==1.4 brotli
+        run: pip install requests selenium webdriver-manager pyperclip==1.9.0 selenium-wire blinker==1.4 brotli certifi h2 hyperframe kaitaistruct pyasn1 pyOpenSSL pyparsing pysocks selenium wsproto zstandard
```
gerrycampion: Are all of these dependencies required? For example, where is pyperclip used?

.github/test/extract_deploy_link.py
gerrycampion, line 1: The latest run after synchronizing is still giving a FileNotFoundError("No log file found"): https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/15854427257/job/44696020609

It is probably because of the other issue I mentioned below

.github/workflows/test_rule_editor_preview.yml
```diff
@@ -15,10 +15,13 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.12
```
gerrycampion, line 18: Use latest versions: `actions/checkout@v4` and `actions/setup-python@v5`

.github/test/extract_deploy_link.py
```diff
@@ -18,7 +18,17 @@
...
+first_push_run = next(
+    (run for run in workflow_data["workflow_runs"] if run["event"] == "pull_request"),
+    None,
+)
+
+if not first_push_run:
+    raise ValueError("No workflow run with event 'push' found")
+
+first_run_id = first_push_run["id"]
+print(f"First workflow run ID with 'push' event: {first_run_id}")
```
gerrycampion: comment says "push", but you are checking for "pull_request". I think you want:
```python
if run["event"] == "push" and run["head_branch"] == "main"
```

.github/workflows/test_rule_editor_preview.yml
SFJohnson24: still missing incrementing the action versions as above

.github/workflows/test_rule_editor_preview.yml
SFJohnson24: please address Gerry's comment above -- are all these dependencies necessary?

.github/test/selenium_test_editor.py
```diff
@@ -13,7 +13,7 @@
 import brotli
 
 # Get the Preview Deployment URL
-RULE_EDITOR_URL = os.getenv("RULE_EDITOR_URL")
+RULE_EDITOR_URL = "https://icy-flower-095494b10-dev.centralus.azurestaticapps.net/"
```
gerrycampion: This should revert to come from the env var.

.github/workflows/test_rule_editor_preview.yml
gerrycampion, line 1: I believe that this workflow first needs to call the `rule-tester` workflow on the current branch so that the current branch is deployed to the dev rule tester. I guess it should be called from the CI workflow, but don't forget to add `needs: rule_tester` under `test_rule_editor_preview` so that rule_tester runs before test_rule_editor_preview

gerrycampion (changes requested):

.github/test/selenium_test_editor.py
```diff
@@ -13,7 +13,7 @@
 import brotli
 
 # Get the Preview Deployment URL
-RULE_EDITOR_URL = os.getenv("RULE_EDITOR_URL")
+RULE_EDITOR_URL = os.getenv("RULE_EDITOR_PREVIEW_URL")
```
gerrycampion: I see this is currently a repo secret. Can you switch it to be a repo variable instead?

.github/test/selenium_test_editor.py
```diff
@@ -32,24 +32,74 @@
 driver = webdriver.Chrome(service=service, options=chrome_options)
 wait = WebDriverWait(driver, 20)
 
+username = os.getenv("RULE_EDITOR_USERNAME")
```
gerrycampion, line 38: I see this is currently a repo secret. Can you switch it to be a repo variable instead? (Note that RULE_EDITOR_PASSWORD should remain a secret.)

.github/workflows/rule-tester.yml
```diff
@@ -1,34 +1,64 @@
...
       - uses: azure/login@v2
         with:
           creds: ${{ env.creds }}
+
       - uses: azure/appservice-settings@v1
         with:
           app-name: ${{ env.functionAppName }}
           mask-inputs: false
-          app-settings-json: '{"WEBSITE_ENABLE_SYNC_UPDATE_SITE": "1"}' # wait for this step to complete before the webapps-deploy step
+          app-settings-json: '{"WEBSITE_ENABLE_SYNC_UPDATE_SITE": "1"}'
```
gerrycampion: Why did you remove this informational comment?

.github/workflows/rule-tester.yml
```diff
@@ -1,34 +1,64 @@
...
 on:
+  workflow_call:
+    inputs:
+      environment:
+        required: true
+        type: string
+      branch:
+        required: false
+        type: string
+
```
gerrycampion: Branch should not be an input param. The checkout action will use the current branch by default

.github/workflows/rule-tester.yml
```diff
@@ -1,34 +1,64 @@
...
     inputs:
       environment:
         required: true
         default: DEV
         type: environment
+      branch:
+        required: false
+        default: ''
+        description: 'Branch to deploy'
+
```
gerrycampion: Branch should not be an input param. The checkout action will use the current branch by default

.github/workflows/rule-tester.yml
```diff
@@ -1,34 +1,64 @@
...
+          ref: ${{ inputs.branch || github.ref_name }}
+
+      - name: Print branch being deployed
+        run: |
+          BRANCH_NAME="${{ inputs.branch }}"
+          if [ -z "$BRANCH_NAME" ]; then
+            BRANCH_NAME="${GITHUB_REF_NAME}"
+          fi
+          echo "Deploying branch: $BRANCH_NAME"
+
```
gerrycampion: This can be removed because it will default to the current branch

.github/workflows/test_rule_editor_preview.yml
```diff
@@ -5,29 +5,37 @@ on:
...
 
 jobs:
+  rule_tester:
+    uses: ./.github/workflows/rule-tester.yml
+    secrets: inherit
+    with:
+      environment: DEV
+      branch: ${{ github.head_ref || github.ref_name }}
   extract_and_test:
+    needs: rule_tester
```
gerrycampion: This should be added to automated-CI instead of here. Better modularity and less complexity

.github/test/selenium_test_editor.py
```diff
@@ -32,24 +32,74 @@
...
+password = os.getenv("RULE_EDITOR_PASSWORD")
+
 try:
     print("Opening Rule Editor site...")
     driver.get(RULE_EDITOR_URL)
 
+    time.sleep(10)  # wait for the page to load
```
gerrycampion: Why are theses `sleeps` necessary? Are there no better alternatives through the use of `waits`?

gerrycampion (changes requested):

RamilCDISC:
I have made the requested changes. For moving the rule tester deployment to CI file, we have to put at least one job that is not dependent on another job in a file. So if i remove the rule tester from the test preview file, the GitHub puts an error. I have replaced the sleeps calls where possible, rest are required for proper execution. @gerrycampion

gerrycampion:
> For moving the rule tester deployment to CI file, we have to put at least one job that is not dependent on another job in a file. So if i remove the rule tester from the test preview file, the GitHub puts an error.

@RamilCDISC When you move this line as well, you should be good:

https://github.com/cdisc-org/cdisc-rules-engine/blob/e47703c415f8edcb5511cf5da574628147ed21fe/.github/workflows/test_rule_editor_preview.yml#L14

RamilCDISC:
@gerrycampion if we will remove the needs rule tester line, the test rule editor preview job will not wait for the rule tester job to complete. We need the rule tester deployment to finish first.

gerrycampion:
> @gerrycampion if we will remove the needs rule tester line, the test rule editor preview job will not wait for the rule tester job to complete. We need the rule tester deployment to finish first.

@RamilCDISC , I'm not asking you to ___remove___ it. I'd like you to ___move___ it to the CI/CD workflow. This way, when it is run automatically as part of the CI/CD workflow, it will require rule tester deployment to finish first. If it is run manually via workflow_dispatch, it will be up to the user to have run the rule-tester deployment manually first.

RamilCDISC:
updated.

.github/workflows/test_rule_editor_preview.yml
```diff
@@ -5,29 +5,37 @@ on:
...
-      
-        run: python .github/test/selenium_test_editor.py
\ No newline at end of file
+        env:
+          RULE_EDITOR_USERNAME: ${{ vars.RULE_EDITOR_USERNAME }}
+          RULE_EDITOR_PASSWORD: ${{ secrets.RULE_EDITOR_PASSWORD }}
+          RULE_EDITOR_URL: ${{ vars.RULE_EDITOR_URL }}
+        run: python .github/test/selenium_test_editor.py
+
+      - name: Upload screenshot (always runs)
```
gerrycampion, line 36: This is triggering a warning here:
https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/17136918511/job/48615579756#step:8:17

gerrycampion (changes requested):

.github/workflows/test_rule_editor_preview.yml
RamilCDISC: fixed.

PR #1209, opened 2025-06-16
updated for supp

SFJohnson24:
This PR remedies the issues with expected, required and permissible operators via resolving a bug in the sdtm_utilities with library.  

To test: cg0016 should have issues with the supp dataset, currently it does not report these.


RamilCDISC (dismissed):
The PR handles the case for SUPP that was causing a bug for few tickets. It was discussed in meeting with team that the issue will be revisited later to add a test for this PR in the test suite. 
The PR is good for now

PR #1215, opened 2025-06-18
new operator, updated record count logic

SFJohnson24:


RamilCDISC (changes requested):
Could you please add test cases for Dask too in the unit tests.

SFJohnson24:
@RamilCDISC this is done

RamilCDISC (approved):
The PR adds a new operator and updates the record count logic. The new operation lets you filter variables from dataset based on  specific criteria. The PR is validated by:

1. Proof reading the updated code for any unwanted comments irrelevant logic.
2. Validating the code changes and updates are rigorous and handles edge cases.
3. Confirming and reviewing the updated unit tests. (Dask test cases were missing which were requested for addition).
4. Running edge cases such as variable present is dataset but missing from metadata. or Duplicate Variable in metadata with different roles.
5. Validating passing of regression tests.
6. Running a rule using the new operation against positive dataset.
7.  Running a rule using the operation against a negative dataset.

PR #1216, opened 2025-06-23
added aprelsupp logic

SFJohnson24:
this PR tweaks https://github.com/cdisc-org/cdisc-rules-engine/pull/1209 so that it will work for RELREC, RELSUP, but also APRELSUP by using contains instead of startswith


RamilCDISC (approved):
updated as required

PR #1218, opened 2025-06-23
Added filter to distinct operation

gerrycampion:
Filtering has been added to the distinct operation. Refer to the updated documentation and the unit tests for examples.

I've also added handling of usdm entities in the scope.

### DIFF
--- a/cdisc_rules_engine/services/data_services/excel_data_service.py
+++ b/cdisc_rules_engine/services/data_services/excel_data_service.py
@@ -102,7 +102,7 @@ class ExcelDataService(BaseDataService):
     def _get_dataset_name(
         self, metadata: pd.DataFrame, first_record: dict, dataset_filename: str
     ) -> str:
-        if DATASET_NAME_COLUMN in metadata.columns:
+        if DATASET_NAME_COLUMN in metadata.columns and not metadata.empty:
             return metadata[DATASET_NAME_COLUMN].iloc[0]
         if self.standard == "usdm":
             return first_record.get("instanceType", dataset_filename.split(".")[0])
--- a/cdisc_rules_engine/services/data_services/excel_data_service.py
+++ b/cdisc_rules_engine/services/data_services/excel_data_service.py
@@ -112,9 +112,10 @@ class ExcelDataService(BaseDataService):
             datasets_worksheet[DATASET_FILENAME_COLUMN] == dataset_name
         ]
         dataset = self.get_dataset(dataset_name=dataset_name)
+        first_record = dataset.data.iloc[0].to_dict() if not dataset.empty else {}
         return SDTMDatasetMetadata(
-            name=dataset_name.split(".")[0].upper(),
-            first_record=(dataset.data.iloc[0].to_dict() if not dataset.empty else {}),
+            name=first_record.get("instanceType", dataset_name.split(".")[0].upper()),
+            first_record=first_record,
             label=metadata[DATASET_LABEL_COLUMN].iloc[0] if not metadata.empty else "",
             modification_date=datetime.fromtimestamp(
                 os.path.getmtime(self.dataset_path)
--- a/cdisc_rules_engine/services/data_services/excel_data_service.py
+++ b/cdisc_rules_engine/services/data_services/excel_data_service.py
@@ -23,6 +23,7 @@ from .base_data_service import BaseDataService, cached_dataset
 DATASETS_SHEET_NAME = "Datasets"
 DATASET_FILENAME_COLUMN = "Filename"
 DATASET_LABEL_COLUMN = "Label"
+DATASET_NAME_COLUMN = "Dataset Name"
 
 
 class ExcelDataService(BaseDataService):
@@ -98,6 +99,15 @@ class ExcelDataService(BaseDataService):
         dataset = PandasDataset(dataframe)
         return dataset
 
+    def _get_dataset_name(
+        self, metadata: pd.DataFrame, first_record: dict, dataset_filename: str
+    ) -> str:
+        if DATASET_NAME_COLUMN in metadata.columns:
+            return metadata[DATASET_NAME_COLUMN].iloc[0]
+        if self.standard == "usdm":
+            return first_record.get("instanceType", dataset_filename.split(".")[0])
+        return dataset_filename.split(".")[0].upper()
+
     @cached_dataset(DatasetTypes.RAW_METADATA.value)
     def get_raw_dataset_metadata(
         self, dataset_name: str, **kwargs
@@ -114,7 +124,7 @@ class ExcelDataService(BaseDataService):
         dataset = self.get_dataset(dataset_name=dataset_name)
         first_record = dataset.data.iloc[0].to_dict() if not dataset.empty else {}
         return SDTMDatasetMetadata(
-            name=first_record.get("instanceType", dataset_name.split(".")[0].upper()),
+            name=self._get_dataset_name(metadata, first_record, dataset_name),
             first_record=first_record,
             label=metadata[DATASET_LABEL_COLUMN].iloc[0] if not metadata.empty else "",
             modification_date=datetime.fromtimestamp(
--- a/cdisc_rules_engine/check_operators/dataframe_operators.py
+++ b/cdisc_rules_engine/check_operators/dataframe_operators.py
@@ -1570,7 +1570,7 @@ class DataframeType(BaseType):
             )
             return bool(target_set.intersection(comparator_set))
 
-        return self.value.apply(check_shared_elements, axis=1).any()
+        return self.value.apply(check_shared_elements, axis=1)
 
     @log_operator_execution
     @type_operator(FIELD_DATAFRAME)
@@ -1591,7 +1591,7 @@ class DataframeType(BaseType):
             )
             return len(target_set.intersection(comparator_set)) == 1
 
-        return self.value.apply(check_exactly_one_shared_element, axis=1).any()
+        return self.value.apply(check_exactly_one_shared_element, axis=1)
 
     @log_operator_execution
     @type_operator(FIELD_DATAFRAME)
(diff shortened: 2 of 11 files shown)


RamilCDISC (changes requested):
I tried putting in the operations in the local editor for rule DDF00009. As seen in attached screenshot we need to update schema to accommodate these.

gerrycampion:
@RamilCDISC 
I created a [ticket ](https://github.com/cdisc-org/cdisc-rules-engine/issues/1219) because this is not a part of the current changes. Only the `filter` property was added and this is already allowed by the schema.

RamilCDISC:
I used the rule logic form the editor and ran the dataset attached in ticket. The editor gives scope skips for all three datasets in the excel file. See attached screenshot. @gerrycampion 

result

gerrycampion:
> I used the rule logic form the editor and ran the dataset attached in ticket. The editor gives scope skips for all three datasets in the excel file. See attached screenshot. @gerrycampion
> 

@RamilCDISC 
I made an update to fix this. This issue is not specific to `unique`, but impacts all operations that use "domain" for the usdm standard.

gerrycampion:
All 3 results are now returned

cdisc_rules_engine/services/data_services/excel_data_service.py
```diff
@@ -112,9 +112,10 @@ def get_raw_dataset_metadata(
             datasets_worksheet[DATASET_FILENAME_COLUMN] == dataset_name
         ]
         dataset = self.get_dataset(dataset_name=dataset_name)
+        first_record = dataset.data.iloc[0].to_dict() if not dataset.empty else {}
         return SDTMDatasetMetadata(
-            name=dataset_name.split(".")[0].upper(),
-            first_record=(dataset.data.iloc[0].to_dict() if not dataset.empty else {}),
+            name=first_record.get("instanceType", dataset_name.split(".")[0].upper()),
```
RamilCDISC: There can be dataset where the name also contains a dot. for that case we might have a wrong named returned here. This is an edge case. Could you please update to use some built in or library function to get the name more robustly @gerrycampion

gerrycampion: This part of the logic has not changed. And at the moment, i am not aware of any of our standards that allow having a dot in the filename, other than the extension indicator. Do you know of any examples?

RamilCDISC: I was not aware of that fact. My concern was that user can add such a dataset. This can be gracefully handled by engine if we update the logic a bit. But if its fixed rule that dataset names cannot have any extra dot then its fine.

RamilCDISC:
right now the editor shows following result when negative dataset (attached in AC) is run against the logic in AC.
```
{
  "ScheduledActivityInstance": [
    {
      "executionStatus": "skipped",
      "dataset": "ScheduledActivityInstance.xpt",
      "domain": "",
      "variables": [],
      "message": "Rule skipped - doesn't apply to entity for rule id=, dataset=ScheduledActivityInstance",
      "errors": []
    }
  ],
  "ScheduleTimeline": [
    {
      "executionStatus": "skipped",
      "dataset": "ScheduleTimeline.xpt",
      "domain": "",
      "variables": [],
      "message": "rule evaluation skipped - operation domain not found",
      "errors": [
        {
          "executionStatus": "skipped",
          "dataset": "ScheduleTimeline.xpt",
          "domain": null,
          "variables": [],
          "message": "Operation distinct requires Domain TIMING but Domain not found in dataset",
          "errors": []
        }
      ]
    }
  ],
  "Timing": [
    {
      "executionStatus": "skipped",
      "dataset": "Timing.xpt",
      "domain": "",
      "variables": [],
      "message": "Rule skipped - doesn't apply to entity for rule id=, dataset=Timing",
      "errors": []
    }
  ]
}
```
I checked the positive dataset too on the sharepoint both negative and positive datasets don't have domain column. I am not sure here dataset is wrong or rule logic. Because according to current rule logic in AC domain will be required in datasets. But if this USDM dataset don't have domain then we need to fix the error

cdisc_rules_engine/services/data_services/excel_data_service.py
ASL-rmarshall: I agree that there needs to be a fall-back if `instanceType` is missing (because we have a separate dataset generated for each primitive data type found in the data - `string`, `boolean`, `float` - and for null values, and none of these has `instanceType`).  However, USDM datasets are generally in mixed case and using `upper()` for the fall-back here might cause problems for later comparison - there doesn't seem to be anything in `rule_applies_to_entity` to account for upper case dataset names.  I'm also not aware of any situations where either `instanceType` or the dataset name will contain a dot.

ASL-rmarshall: @gerrycampion More general question about this - why are you referencing `instanceType` in the `ExcelDataService`? As far as I can see, this data service would only be used if you were running the CLI against Excel data files and I don't think we've ever done that for USDM rules.

gerrycampion:
> right now the editor shows following result when negative dataset (attached in AC) is run against the logic in AC.
> 
> 
> I checked the positive dataset too on the sharepoint both negative and positive datasets don't have domain column. I am not sure here dataset is wrong or rule logic. Because according to current rule logic in AC domain will be required in datasets. But if this USDM dataset don't have domain then we need to fix the error

@RamilCDISC 
I tested by running this branch of engine locally with rule editor locally and I do not get the same result. Domain is no longer required for usdm. This is what I get:

```json
{
  "ScheduledActivityInstance": [
    {
      "executionStatus": "skipped",
      "dataset": "ScheduledActivityInstance.xpt",
      "domain": "",
      "variables": [],
      "message": "Rule skipped - doesn't apply to entity for rule id=DDF00009, dataset=ScheduledActivityInstance",
      "errors": []
    }
  ],
  "ScheduleTimeline": [
    {
      "executionStatus": "success",
      "dataset": "ScheduleTimeline.xpt",
      "domain": "ScheduleTimeline",
      "variables": [
        "id",
        "name",
        "parent_entity",
        "parent_id",
        "parent_rel"
      ],
      "message": "The schedule timeline does not contain an anchor (i.e., a scheduled activity instance that is referenced by a Fixed Reference timing).",
      "errors": [
        {
          "value": {
            "parent_id": "StudyDesign_1",
            "name": "Adverse Event Timeline",
            "parent_rel": "scheduleTimelines",
            "parent_entity": "StudyDesign",
            "id": "ScheduleTimeline_1"
          },
          "dataset": "ScheduleTimeline.xpt",
          "row": 2
        },
        {
          "value": {
            "parent_id": "StudyDesign_1",
            "name": "Early Termination Timeline",
            "parent_rel": "scheduleTimelines",
            "parent_entity": "StudyDesign",
            "id": "ScheduleTimeline_2"
          },
          "dataset": "ScheduleTimeline.xpt",
          "row": 3
        },
        {
          "value": {
            "parent_id": "StudyDesign_1",
            "name": "Vital Sign Blood Pressure Timeline",
            "parent_rel": "scheduleTimelines",
            "parent_entity": "StudyDesign",
            "id": "ScheduleTimeline_3"
          },
          "dataset": "ScheduleTimeline.xpt",
          "row": 4
        }
      ]
    }
  ],
  "Timing": [
    {
      "executionStatus": "skipped",
      "dataset": "Timing.xpt",
      "domain": "",
      "variables": [],
      "message": "Rule skipped - doesn't apply to entity for rule id=DDF00009, dataset=Timing",
      "errors": []
    }
  ]
}
```

Are you using the updated branch of rule engine?

cdisc_rules_engine/services/data_services/excel_data_service.py
gerrycampion: @ASL-rmarshall, the fallback is trying to say: "If the dataset is usdm (has instanceType), then use instanceType. Otherwise, it is sdtm, so use the dataset name.".  I overlooked the other possible primitive datasets like string, boolean, float and I did not intend to upper() for usdm. Do you have any suggestion for extracting the names for the primitive dataset types for usdm without breaking the sdtm functionality?

ASL-rmarshall: @gerrycampion Would this work?:
```python
            name=(
                first_record.get("instanceType", dataset_name.split(".")[0])
                if self.standard == "usdm"
                else dataset_name.split(".")[0].upper()
            )
```

ASL-rmarshall: @gerrycampion ... alternatively, the "Datasets" sheet in the standard Excel file is supposed to have a "Dataset Name" column, so it should be possible to use something like this for any standard:
```python
DATASET_NAME_COLUMN = "Dataset Name"
...
            name=metadata[DATASET_NAME_COLUMN].iloc[0]
```

gerrycampion: thanks, i've added logic to use both. I think a lot of sdtm test data does not have the Dataset Name column

RamilCDISC:
After recent updates the previous bug is solved. Review is all good on my side.

cdisc_rules_engine/services/data_services/excel_data_service.py
```diff
@@ -98,6 +99,15 @@ def get_dataset(self, dataset_name: str, **params) -> DatasetInterface:
         dataset = PandasDataset(dataframe)
         return dataset
 
+    def _get_dataset_name(
+        self, metadata: pd.DataFrame, first_record: dict, dataset_filename: str
+    ) -> str:
+        if DATASET_NAME_COLUMN in metadata.columns:
```
ASL-rmarshall: As the retrieval of the dataset label from `metadata` has logic to deal with the dataset (file) name not appearing in the Datasets sheet (line 129: `if not metadata.empty`), I think this should also be used here:
```python
       if DATASET_NAME_COLUMN in metadata.columns and not metadata.empty:
```

SFJohnson24:
@gerrycampion @RamilCDISC I have traced failing USDM part of the test suite back to this ticket. If you look at the comparison--the exit code and comparison chart was not being created for the test suite. I have updated this in https://github.com/cdisc-org/cdisc-rules-engine/pull/1294 I am going to reopen this ticket to resolve the issue with the skipped rules.

@ASL-rmarshall I believe we need to update the test suite data as there is no scheduleddecisioninstance which is causing 3 rules to skip

Issue #305, opened 2023-03-14
As VALIDATOR, I want to cross-check my study define.xml against metadata in CDISC Library for inconsistent variable role so that I can stay compliant with CG0010


chowsanthony:
Rule: Metadata attribute of Variable Role = IG Role for domains in IG, Metadata attribute of Variable Role = Model Role for custom domains

`ItemRef[@Role]` matches those in CDISC Library.

See https://github.com/cdisc-org/cdisc-rules-engine/issues/302 for definition of custom domain.
See https://github.com/cdisc-org/cdisc-rules-engine/issues/303 for definition of modeled domain.

Custom domain wills check against the corresponding SDTM. Modeled domains will check against the corresponding SDTMIG.

nickdedonder:
For custom domains the information needs to be taken from the class in the model and for non-custom domain it merges with the domain metadata from the IG as well as the model. 
We need a new dataset builder for define with library metadata

SFJohnson24:
I believe the new dataset builder I created will fit this rule  #604  
https://github.com/cdisc-org/cdisc-rules-engine/pull/879

I will need to do a bit of touching up for the role logic to perform what Nick is describing above but it does fetch role from the define and library for the variables.

gerrycampion:
@nickdedonder is this accurate:

To start, the SDTM class hierarchy looks like:
- General Observations
  - Interventions
  - Events
  - Findings
    - Findings About
- Special-Purpose
- Associated Persons
- Trial Design
- Study Reference
- Relationship

You are only allowed to have custom domains for the 4 General Observations' classes, aka the "DETECTABLE_CLASSES" in the code. AP and SUPPQUAL (and maybe others) are special because they interact with the other domain structures to make a domain, but those domains are not individually defined in the model or IG.

I think there are 3 possible cases:
- Standard General Observations domain: Metadata should come from model hierarchy and any metadata in the IG should overwrite the model metadata. For example, LB should take all class variables from model "Findings", model "General Observations" and overwrite matching metadata with IG "LB" dataset variables.
- Custom General Observations domain: Metadata should come from model hierarchy only
- Standard non-General Observations domain: Metadata should come from the IG only

nickdedonder:
@gerrycampion yes, this is accurate.
You can only create custom domain on the 3 general observation classes (4 is you consider findings about separate). You could also create it for AP, but it would follow the rules as it would be an AP based on one of the 3 general observation classes.

[...]

Issue #560, opened 2023-10-10
Rule blocked: CORERULES-707 (SEND-270): library_variable_data_type fails when variable from model added


JozefAerts:
When a model variable (such as VSLOBXFL to VS in SEND) is added, the operator "library_variable_data_type" returns the empty string, which then causes a false positive as --LOBXFL normally has the type "Char".
I tried to make a temporary workaround, also testing that "library_variable_data_type" is not empty, which however generates another false positive.
The same applies when there is a variable that is invalid, e.g. "VSINVALID".
Also disturbing is that only the first discrepancy is reported.
Files are attached.
[Rule_271.zip](https://github.com/cdisc-org/cdisc-rules-engine/files/12856439/Rule_271.zip)

SFJohnson24:
This is also applicable to TIG0191

PR #1220, opened 2025-06-24
Lib var data type

SFJohnson24:
this PR resolves several issues
- sort keys throughout the sdtm_utility--it is not ordinal, order_number logic is removed except when renaming ordinal for the dataset builders.
- extends() causing duplicates when a noncustom domain is given to library.
- get_allow_class_variables having mixed logic with the model and IG.   get_variables_metadata_from_standard_model) was also reworked to have clear dataflow/variable naming conventions.  The logic now correctly follows that of Gerry's comment: https://github.com/cdisc-org/cdisc-rules-engine/issues/305
- deleted the two test_get_model_filtered_variables findings_about class as they do not accurately reflect FA class data.

to test
--trial Jozef's data in the linked PR.  It is positive
--an expected/permissible/required operation rule (I used CORE-000334)
--cg0219 (while it still has the SEND variables from the model, LBPDUR is picked up).  I changed model filter variables operator to reflect the logic from the implementation guide: 'Fetches variable level library model properties filtered by the provided key_name and key_value.' https://jira.cdisc.org/browse/CORERULES-183?jql=project%20%3D%20CORERULES%20AND%20resolution%20%3D%20Unresolved%20AND%20text%20~%20%22cg0219%22%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC


RamilCDISC (changes requested):
As the PR is connected to issue 1185 while related to SEND48, I deployed this branch to dev rule tester and used the dev instance of editor to run a test on send48 using the positive and negative datasets from Sharepoint. 

The editor gave '500 internal server Error". I tested the production editor forSEND48 and same result. Below is screenshot from dev and production editor in order.

dev:
dev-branch

production:
main

SFJohnson24:
@RamilCDISC the PR fixes the issues with the metadata laid out by jozef in the ticket.  If you look at the dataflow in get_variables_metadata_from_standard you can see that it addresses the ordering issues with identifiers and timing variables laid out and described in the ticket.   The rule does work for cg0664 and its datasets so this issue is not related to my code but likely something with the dataset.  Does the request look normal?

RamilCDISC:
Rest all looks good. Could you please add a unit test for the bug in #560? The  bug should be fixed by the updated code but a unit test would be good for future. @SFJohnson24

SFJohnson24:
@RamilCDISC that issue involves library, it would need to have the rule published and be integrated into our system test suite as it needs to make api calls for library Metadata

PR #1224, opened 2025-06-27
Record column

SFJohnson24:
this PR does a few things:
- updates schema so grouping can be an operation result $TIMING_VARIABLES
- update record count to process TIMING VARIABLES as the grouping:
  -  this involves IDing an operation result as the grouping (done via checking if it is a column name and if it starts with '$')
  -  we then parse the column to get the list of columns from it that we want to use for the grouping; we also keep track of blank ('' or None for all columns) as pd will replace these with NaN.
  -  group on this list of columns
  -  restore the '' / None values to the columns that became NaN when grouped
-   when the grouping is restored, engine tries to merge on the target which is $TIMING_VARIABLES.  Logic is added to ID if it is an operation result (done via checking if it is a column name and if it starts with '$');  We then parse the operation result and return the columns to replace the grouping ID so that the merge can be performed between the original dataset and the grouped result

to test: 
[Rule_underscores.json](https://github.com/user-attachments/files/20954374/Rule_underscores.json)

(I believe the rule may need to be altered, my logic: 
[Rule_underscores.json](https://github.com/user-attachments/files/20955726/Rule_underscores.json)

 negative:
[DatasetsN.json](https://github.com/user-attachments/files/20954382/DatasetsN.json)
[DatasetsN2.json](https://github.com/user-attachments/files/20955741/DatasetsN2.json)

positive
[DatasetsP.json](https://github.com/user-attachments/files/20955742/DatasetsP.json)
[DatasetsP2.json](https://github.com/user-attachments/files/20955743/DatasetsP2.json)

### DIFF
--- a/cdisc_rules_engine/operations/base_operation.py
+++ b/cdisc_rules_engine/operations/base_operation.py
@@ -198,7 +198,7 @@ class BaseOperation:
             and target_metadata.is_supp
         ):
             domain_for_library = "SUPPQUAL"
-        elif target_metadata and target_metadata.name.lower().startswith("REL"):
+        elif target_metadata and "rel" in target_metadata.name.lower():
             domain_for_library = target_metadata.name
         else:
             domain_for_library = self.params.domain
--- a/cdisc_rules_engine/operations/base_operation.py
+++ b/cdisc_rules_engine/operations/base_operation.py
@@ -172,18 +172,51 @@ class BaseOperation:
         )
 
     def _get_grouping_columns(self) -> List[str]:
-        return (
-            self.params.grouping
-            if not self.params.grouping_aliases
-            else [
-                (
-                    self.params.grouping_aliases[i]
-                    if 0 <= i < len(self.params.grouping_aliases)
-                    else v
-                )
-                for i, v in enumerate(self.params.grouping)
-            ]
-        )
+        if any(item.startswith("$") for item in self.params.grouping):
+            return self._expand_operation_results_in_grouping(self.params.grouping)
+        else:
+            return (
+                self.params.grouping
+                if not self.params.grouping_aliases
+                else [
+                    (
+                        self.params.grouping_aliases[i]
+                        if 0 <= i < len(self.params.grouping_aliases)
+                        else v
+                    )
+                    for i, v in enumerate(self.params.grouping)
+                ]
+            )
+
+    def _expand_operation_results_in_grouping(self, grouping_list):
+        expanded = []
+        for item in grouping_list:
+            if item.startswith("$") and item in self.evaluation_dataset.columns:
+                operation_col = self.evaluation_dataset[item]
+                first_val = operation_col.iloc[0]
+                if operation_col.astype(str).nunique() == 1:
+                    if isinstance(first_val, (list, tuple)):
+                        expanded.extend(first_val)
+                    else:
+                        expanded.append(item)
+                else:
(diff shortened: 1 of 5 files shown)


RamilCDISC (changes requested):
I made yaml of the rule attached in PR and pointed to this branch's CORE-base.json for schema. The local editor shows yellow lines under key name and key value properties and do not accept them. 

The schema i built is following:
```
Authorities:
  - Organization: CDISC
    Standards:
      - Name: SDTMIG
        Version: '3.4'
        References:
          - Citations:
              - Document: SDTM v2.0
                Item: --REPNUM
                Section: Findings
                Cited Guidance: >
                  The instance number of a test that is repeated within a given timeframe for the same test.
                  The level of granularity can vary (e.g., within a time point or within a visit).
            Origin: SDTM and SDTMIG Conformance Rules
            Version: '2.0'
            Rule Identifier:
              Id: CG0562
              Version: '1'
      - Name: SDTMIG
        Version: '3.3'
        References:
          - Citations:
              - Document: Model v1.7
                Item: --REPNUM
                Section: Table 2.2.3
                Cited Guidance: >
                  The instance number of a test that is repeated within a given timeframe for the same test.
                  The level of granularity can vary (e.g., within a time point or within a visit).
                  For example, multiple measurements of blood pressure or multiple analyses of a sample.
            Origin: SDTM and SDTMIG Conformance Rules
            Version: '2.0'
            Rule Identifier:
              Id: CG0562
              Version: '1'
      - Name: TIG
        Version: '1.0'
        Substandard: SDTM
        References:
          - Citations:
              - Document: SDTM v2.1
                Item: --REPNUM
                Section: 3.1.3 The Findings Observation Class
                Cited Guidance: >
                  The instance number of a test that is repeated within a given timeframe for the same test.
                  The level of granularity can vary (e.g., within a time point or within a visit).
            Origin: TIG Conformance Rules
            Version: '1.0'
            Rule Identifier:
              Id: TIG0648
              Version: '1'

Check:
  any:
    - all:
        - name: --REPNUM
          operator: exists
        - name: --REPNUM
          operator: empty
    - all:
        - name: --REPNUM
          operator: exists
        - name: $record_count_per_subject_test_timing
          operator: greater_than
          value: 1
        - name: --REPNUM
          operator: is_not_unique_set
          value:
            - USUBJID
            - --TESTCD
            - $TIMING_VARIABLES

Core:
  Id: CDISC.SDTMIG.CG0562
  Status: Draft
  Version: '1'

Description: >
  Raise an error when REPNUM is in the dataset and there are multiple records for a subject for a test
  within the timeframe identified by the timing variables on the record, and --REPNUM is null or not unique
  per subject per test per timing variables.

Operations:
  - id: $TIMING_VARIABLES
    key name: role
    key value: Timing
    operator: get_dataset_filtered_variables
  - id: $record_count_per_subject_test_timing
    group:
      - $TIMING_VARIABLES
    operator: record_count

Outcome:
  Message: --REPNUM is null or is not unique per subject per test per timing variables.
  Output Variables:
    - USUBJID
    - --TESTCD
    - --REPNUM
    - $TIMING_VARIABLES

Scope:
  Classes:
    Include:
      - FINDINGS
  Domains:
    Include:
      - ALL
  Use Case: INDH

Sensitivity: Record
Rule Type: Record Data

```

SFJohnson24:
@RamilCDISC key name and value need underscores.  Should be key_name and key_value.  You can always doublecheck the readme or the schema if you are think there is an issue just to confirm

PR #1226, opened 2025-06-30
fixed tab

SFJohnson24:
else statement was improperly indented so it was always logging the warning


RamilCDISC (approved):
updated as required

PR #1227, opened 2025-06-30
update null handling

SFJohnson24:
changed the logic from adding 'null' to None.  When json.dumps() None, it becomes json compatible null

image
I ran cg0433 on negative data.  you can see the null displayed in results

### DIFF
--- a/cdisc_rules_engine/models/actions.py
+++ b/cdisc_rules_engine/models/actions.py
@@ -284,13 +284,13 @@ class COREActions(BaseActions):
         for key, value in row_dict.items():
             if isinstance(value, list):
                 filtered_dict[key] = (
-                    "null"
+                    None
                     if any(val in NULL_FLAVORS for val in value) or pd.isna(value).any()
                     else value
                 )
             else:
                 filtered_dict[key] = (
-                    "null" if (value in NULL_FLAVORS or pd.isna(value)) else value
+                    None if (value in NULL_FLAVORS or pd.isna(value)) else value
                 )
         error_object = ValidationErrorEntity(
             dataset=(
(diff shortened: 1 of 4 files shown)


cdisc_rules_engine/models/actions.py
```diff
@@ -284,13 +284,13 @@ def _create_error_object(
         for key, value in row_dict.items():
             if isinstance(value, list):
                 filtered_dict[key] = (
-                    "null"
+                    None
                     if any(val in NULL_FLAVORS for val in value) or pd.isna(value).any()
```
ASL-rmarshall: @SFJohnson24 This will replace a value list with `None` if any value in the list is some sort of null - so a list like 
`["Valid value 1", "OK", NaN]` would be entirely replaced with `None`.  Is this intentional?  I would have expected each of the values in the list to be replaced individually (e.g., to give `["Valid value 1", "OK", None]`)

SFJohnson24: @ASL-rmarshall I missed this part of the ticket, correct it and the test suite

RamilCDISC:
@ASL-rmarshall The PR is good from my side. Since you requested a change before I am not merging right now so you can still have a look.

cdisc_rules_engine/models/actions.py
```diff
@@ -283,14 +283,16 @@ def _create_error_object(
         filtered_dict = {}
         for key, value in row_dict.items():
             if isinstance(value, list):
-                filtered_dict[key] = (
-                    "null"
-                    if any(val in NULL_FLAVORS for val in value) or pd.isna(value).any()
-                    else value
-                )
+                filtered_list = []
```
ASL-rmarshall: I'm happy that lists containing nulls are now being handled as I'd expect, but I'm wondering if there's any reason not to use a list comprehension for this:
```python
filtered_dict[key] = [None if (val in NULL_FLAVORS or pd.isna(val)) else val for val in value]
```

RamilCDISC:
based on conversation hold in todays scrum meeting the PR is ready to be closed.

PR #1230, opened 2025-07-03
added custom list functionality

SFJohnson24:
this PR adds a flag to list-rule-sets to allow listings of custom standards and versions.

to test: add custom rules and standards into the cache, use the --custom or -o shorthand flag.


RamilCDISC (approved):
The PR adds flag to filter custom standards. The custom standards successfully show up

PR #1231, opened 2025-07-03
add supp logic

SFJohnson24:
This PR allows for SUPP-- to be used as the match when the child property is given.
[Rule_underscores.json](https://github.com/user-attachments/files/21046075/Rule_underscores.json)
see above rule.
It also resolves an issues with is_reference where None was added instead of raising an error.

to test: run above rule against cg0371 data.
[Rule_underscores2.json](https://github.com/user-attachments/files/21046089/Rule_underscores2.json)
this rule can also be used as it only specifies SUPPDM, not SUPPLB so a result is removed from report.  This will flag an error with the dataset and skip it.  (see logs)


SFJohnson24:
@RamilCDISC I have updated the tests per your request

PR #1235, opened 2025-07-08
Meddra bug

SFJohnson24:
There is an issue with the meddra logic specifically--since the CD is not alphanumeric as with other external dictionary codes, some sample data was made designating the AEBDSYCD as a number.  The cache is populated using string logic and thus the keys are strings.  I have added logic to check the type to allow for both character or number typing for the code so that an error does not occur when the cache lookup occurs.
image

to test: 
I have the test data I can send as I cannot paste XPT/asc files in github.  ae is the file with numeric code and should be positive data (no errors).  ae2 is negative data and should return issues as below.
image


RamilCDISC (approved):
updated as required

PR #1239, opened 2025-07-11
added ccode parsing logic

SFJohnson24:
this PR adds parsing for dataset builders that use library metadata.  It checks if a submission value exists, meaning a codelist is present and then parses to get the ccode, adding it to the variable which then becomes a column on merge.

to test:
[Rule_underscores.json](https://github.com/user-attachments/files/21189252/Rule_underscores.json)
[DatasetsN.json](https://github.com/user-attachments/files/21189253/DatasetsN.json)
I used the negative define from cg0001
[Book1.xlsx](https://github.com/user-attachments/files/21189265/Book1.xlsx)  This shows the dataset for analysis -- the ccodes from library match with https://library.cdisc.org/browser/#/mdr/sendig/3-1/classes/Findings/datasets/VS

### DIFF
--- a/cdisc_rules_engine/dataset_builders/define_variables_with_library_metadata.py
+++ b/cdisc_rules_engine/dataset_builders/define_variables_with_library_metadata.py
@@ -28,6 +28,7 @@ class DefineVariablesWithLibraryMetadataDatasetBuilder(BaseDatasetBuilder):
         "library_variable_data_type",
         "library_variable_role",
         "library_variable_core",
+        "library_variable_ccode",
         "library_variable_order_number"
         """
         # get Define XML metadata for domain and use it as a rule comparator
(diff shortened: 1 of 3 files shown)


RamilCDISC (approved):
The PR adds the code parsing logic to the base dataset builder. The updated code was validated by:

1. Reviewing and proof reading the code for any unwanted code or comments.
2. Making sure the best coding practices are followed.
3. Making sure all references are updated, added or changed in all appropriate places where required. 
4. Making sure documentation is updated.
5. Making sure all unit tests and regression tests pass.
6. Verifying the deployment of rule tester endpoint.
7. Running the tests locally in CLI against negative dataset.
8. Verifying each report issue in the report by comparing the codelist value in define and library, to validate that engine reads the values correctly and compare them correctly.
9. Verify the behavior of engine when define file is not provided.

PR #1244, opened 2025-07-15
CG0235

SFJohnson24:
This PR addresses two issues:
- it moves the operator helpers and vectorized operations from business_rules into engine's helpers/operators file for easier management
- it corrects an error where isoparse is unable to parse a column when it doesnt exist--this was added to the missing column errors.
before:
image
after:
image

To test: run editor locally and test cg0235 negative data

### DIFF
--- a/cdisc_rules_engine/check_operators/dataframe_operators.py
+++ b/cdisc_rules_engine/check_operators/dataframe_operators.py
@@ -1,7 +1,7 @@
 from business_rules.operators import BaseType, type_operator
 from typing import Union, Any, List, Tuple
 from business_rules.fields import FIELD_DATAFRAME
-from business_rules.utils import (
+from cdisc_rules_engine.check_operators.helpers import (
     flatten_list,
     is_valid_date,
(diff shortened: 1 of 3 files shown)


SFJohnson24:
> Could you please update the business_rule version in requirements.txt and pyproject.toml. I think we should have the same unit tests also copied here for the copied operators

this is done @RamilCDISC

RamilCDISC (approved):
The PR adds the removed operators of business_rule to engine for better maintenance and usability. The operators are already well tested and validated. The PR only copies them to Engine. 

The integration of these tests in engine was validated by:

1. Validate that all relevant files were updated for imports of these operators.
2. Validate and match that all operators that were deleted from business rules were added to the engine.
3. Validate that the version numbers were increased in both business rules and engine repositories.
4. Validate that all unit tests pass.
5. Validate that test suite passes, as test suite executes all the operators in the engine. Successful execution mean successful integration.
6. Validate that the rule editor works fine with the new update.
7. Validate that no previous functionality is affected.

PR #1250, opened 2025-07-18
dask readme

SFJohnson24:
I added an update to our docset outlining setting the dataset size limit for dask as it came up in emails.  I have tested both windows commands as well as the .env strategy for implementation coercion to dask.  @RamilCDISC I was hoping you could test the linux/max bash command which should work.  
.env
image
powershell command
image
windows command prompt
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,7 +486,7 @@ $env:DATASET_SIZE_THRESHOLD=0; core.exe validate -rest -of -config -commands
 ### Linux/Mac (Bash)
 
 ```bash
-DATASET_SIZE_THRESHOLD=0 ./core.exe -rest -of -config -commands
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -464,3 +464,41 @@ When submitting an issue, please include:
...
+### Windows (PowerShell)
+
+```powershell
+$env:DATASET_SIZE_THRESHOLD=0; core.exe validate -rest -of -config -commands
+```
+
+### Linux/Mac (Bash)
+
+```bash
+DATASET_SIZE_THRESHOLD=0 core.exe -rest -of -config -commands
```
RamilCDISC: The command for mac for executing will be ./core instead of core.exe Also the latest release i tried running it on mac. It did not execute and gives an error that the core is damaged.

PR #1254, opened 2025-07-21
update commands

SFJohnson24:


SFJohnson24:
@RamilCDISC fixed--it was CORE in the pathing wasnt capitalized correctly

RamilCDISC:
The comparison scripts has some error and testing is still failing because of that.

RamilCDISC (approved):
The test suite is updated and fixed

PR #1260, opened 2025-07-22
update null handling

SFJohnson24:
this PR makes null handling consistent across reporting- FB 3205 avoids getting into our current null handling given its sensitivity.  This resulted in a 500 error when executing the negative data with the rule.  This PR makes the NaN handling consistent so as to prevent these errors.

use FB3205 and negative data to test

### DIFF
--- a/cdisc_rules_engine/constants/__init__.py
+++ b/cdisc_rules_engine/constants/__init__.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 # a message like: [INFO 2021-12-29 17:10:26,575 - module.py:44] - Log Message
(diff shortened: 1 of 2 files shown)


RamilCDISC (approved):
The PR adds the handling of NULL value occurrence for FDA business rules. The null flavors were updated accurately and the processing code was also updated accordingly. The validation was done by:

1. Reviewing the PR for any unwanted comments or code.
2. Reviewing the PR to validate the solution logic.
3. Validating all unit and regression tests pass.
4. Deploying the rule tester using the PR's branch and testing the negative dataset and positive dataset in the editor.
5. Validating the result according to the operators in the rule.

PR #1270, opened 2025-07-25
error

SFJohnson24:
It was brought up with the 2 rule ids which were unpublished and deleted--if a rule is specified and its not in the cache when being grabbed by the ID-- it should have an error thrown, not a log.


RamilCDISC (approved):
updated as required.

PR #1273, opened 2025-07-28
Define metadata

SFJohnson24:


SFJohnson24:
@RamilCDISC the excel file needs to specify CT.  It works fine when you do this.

SFJohnson24:
@RamilCDISC I added error handling for the 1 dataset that is has no issues in the negative data as well as handling if you enter incorrect ct to help troubleshoot.  Try using an incorrect CT in the library tab of the excel and running cg0001.  It should now tell you the CT is not correct in editor.

PR #1276, opened 2025-07-29
operator rework

SFJohnson24:
this PR reworks the is_not_unique_relationship check operator.  The original method was missing 1 expected violation due to null handling in pandas' duplicated() function, flagging only 3 records instead of 4.  This PR handles nulls and '' better.  This PR maintains checks from both target and comparator perspectives while properly excluding null values as base grouping keys to prevent false positives.

copilot summary:This pull request refactors the `is_not_unique_relationship` method in `dataframe_operators.py` to improve clarity and maintainability by introducing a helper method. The changes streamline the logic for identifying violations of one-to-one relationship constraints.

### Refactoring of `is_not_unique_relationship`:

* Simplified the main logic by introducing the `_find_relationship_violations` helper method to encapsulate the identification of target values that violate one-to-one relationship constraints. This reduces code duplication and improves readability.
* Updated the handling of null or empty values in both `target` and `comparator` columns to ensure consistent treatment during the relationship violation checks.

### New helper method `_find_relationship_violations`:

* Added `_find_relationship_violations` to encapsulate logic for identifying violated targets based on one-to-one relationship constraints between `target` and `comparator`. This method ensures separation of concerns and makes the code easier to test and maintain.

to test: check the FB0904.  The empty TSVAL part of the rule must be removed.


cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -1121,26 +1121,53 @@ def is_not_unique_relationship(self, other_value):
...
+            if clean_targets:
+                result = result | self.value[target].isin(clean_targets)
+            if has_null_target:
+                result = result | self.value[target].isna()
         return result
 
+    def _find_relationship_violations(self, df_without_duplicates, target, comparator):
+        """Find all target values that violate one-to-one relationship constraints."""
+        violated_targets = set()
+        for target_val in df_without_duplicates[target].dropna().unique():
```
RamilCDISC, line 1146: The nested for loops here can make the processing very slow especially for when datasets size is large. Could you please vectorize these using groupby?

RamilCDISC (changes requested):
Please see the comments below.  Could you also please add some unit tests covering the new logic especially the case where two different flavors of nulls are compared

cdisc_rules_engine/check_operators/dataframe_operators.py
SFJohnson24: I do not believe I can vectorize this operator given the constraints of what it needs to do, especially in relation to null, while also making it work with Dask and Pandas.  The logic of this rule requires that we not take null value's perspective, otherwise we will flag all null values as violations as null is matched with many values (not 1:1).  This requires condition directional logic for target-->comparator and vice versa.  Distributed computing and partition boundaries makes this conditional directional logic very difficult.  The nested for loops work well with distributed computing and despite the bottleneck, are an unfortunate necessity for this to play well with dask and pandas

SFJohnson24:
@RamilCDISC  I forgot to mention, the rule as written will skip the 4th issue as  
   - name: TSVALCD
      operator: non_empty 
will rule it out.  Remove this and you will get the 4 issues

RamilCDISC (approved):
The PR reworks the unique relationship operation. The changes were validated by:

1. Reviewing the updated code for any unwanted comments or code logic.
2. Reviewing the updated logic for its efficiency and integration.
3. Validating passing of all unit-tests and regression tests.
4. Validating by running a validation locally in the rules engine.
5. Validating by running the validation using editor and deploying the branch to rule tester
6. Validating the edge case where both comparator and targets are null.
7. Validating the edge case where comparator is null but target is not.
8. Validating the null case where comparator is null and target is a duplicate in the column.
9. Validating the edge case operator and target same value.

PR #1279, opened 2025-07-30
update reporting

SFJohnson24:
this PR updates None handling the rule report.  The issue encountered by Els is described:
For CG0265: First the rule filters only completed TSVAL, and then completed TSVALCD. And then it checks the 1-1 relationship. But still we have a study with a TS where TSVAL for TSPARMCD = TCNRTL is ‘NONE’ with TSVALCD = C411132 is flagged, while this is really the only combination. When looking at the output report, it reports TSVAL = NONE as 'null'. So maybe something is going wrong in the translation from NONE to accidentally 'null' and the filter for 'non_empty' is then not taken into account? Because if I remove the TSVAL/TSVALCD = non_empty filters, then there are different TSVALCDs for the same TSVAL.
image
image

i made test data with TSVAL of 'None'.
[Datasets.json](https://github.com/user-attachments/files/21514753/Datasets.json)

### DIFF
--- a/cdisc_rules_engine/services/reporting/base_report.py
+++ b/cdisc_rules_engine/services/reporting/base_report.py
@@ -124,10 +124,13 @@ class BaseReport(ABC):
                         "row": error.get("row", ""),
                         "SEQ": error.get("SEQ", ""),
                     }
-                    values = [
-                        str(error.get("value", {}).get(variable))
-                        for variable in variables
-                    ]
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
The Null flavor handling has been added appropriately. The passing of unit tests show proper integration. Test suite fails because of a change in rule by an author. All regression tests pass

PR #1290, opened 2025-08-08
"Applied performance fix and formats"

pendingintent:
These are all fixes made to usdm_data_service.py and added to this branch of cdsic-rules-engine

### DIFF
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -1,11 +1,11 @@
 name: Build and Push Docker Image
 on:
-  workflow_run:
-    workflows: ["Release CORE Rules Engine"]
-    types:
-      - completed
-    branches:
-      - main
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+        description: "Version tag for the Docker image"
 jobs:
   check-if-latest:
     runs-on: ubuntu-latest
@@ -43,6 +43,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
+          ref: ${{ needs.check-if-latest.outputs.version }}
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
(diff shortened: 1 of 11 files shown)


pendingintent:
The test_suite is failing.  Used copilot to explain:

The job failed due to a TypeError: "Cannot perform 'rand_' with a dtyped [float64] array and scalar of type [bool]". This error occurs when using the bitwise AND operator (&, which maps to rand_) between a float64 array and a boolean scalar. In Python and pandas, bitwise operations like & are only valid between compatible dtypes (e.g., bool & bool, int & int).

This error appears in several rule IDs (CORE-000691, CORE-000692, CORE-000693, CORE-000694, CORE-000695).

Solution: You must ensure that both operands of the & operator are of boolean dtype. In your code, where you apply a bitwise AND (&) between a float64 array and a boolean, you should cast the float64 array to boolean first.

RamilCDISC:
@pendingintent If you will see the USDM sections in the test suite job run on this PR, you will see a trace of an error that occurred while executing the USDM validation. This mean validation for some dataset is not working properly. If you need the dataset and rules you can clone the test suite repository else let me know I can share the datasets with you.

pendingintent:
@RamilCDISC, I am going to need to request some assistance with this issue.  I am not familar with the CORE rules and believe I am reaching my limits.  At the very least, I need assistance with understanding the test suite.

pendingintent:
OK.  @RamilCDISC, the test suite is now running successfully!  Thank you to @SFJohnson24 for VSCode test and debug baptism-by-fire :)

RamilCDISC:
@pendingintent Thank you for your fixes. If you would look into the usdm_positive_comparision excel file in the artifacts of the test suite run, you will notice that some rules are skipping on this branch.

SFJohnson24:
note @pendingintent @RamilCDISC those 3 rules appear to have been skipping for some time but the test suite was not capturing the error code.  I made a ticket for #1295 The changes in this ticket may be causing this but it is also likely that it is not and the changes from the linked ticket I attached to that issue are the cause

PR #1301, opened 2025-08-15
update

SFJohnson24:
This PR brings -lr in line with -r.  This allows the user to specify multiple -lr.  It parses them as they can be a mix of directories of local rules and file paths.  

This pull request introduces improvements to how local rule directories are handled and updates configuration for code linting. The main changes focus on supporting multiple local rule directories and ensuring correct rule file loading, along with a minor update to the `.flake8` configuration.

**Local rules handling:**

* The `--local_rules` CLI option in `core.py` now supports multiple directories, allowing users to specify more than one path for local rules.
* The `load_rules_from_local` function in `scripts/script_utils.py` was updated to iterate over all provided local rule paths, loading files from each directory or file path. This enables proper aggregation of rule files from multiple sources.

**Linting configuration:**

* The `.flake8` configuration was updated to ignore the `E713` error code, in addition to the previously ignored codes.

to test - try iterations of -lr

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,7 @@
 [flake8]
 max-line-length = 120
 max_complexity = 10
-ignore = E203, W503
+ignore = E203, W503, E713
 
 exclude = .github,
     .pytest_cache,
--- a/core.py
+++ b/core.py
@@ -176,6 +176,7 @@ def cli():
     required=False,
     type=click.Path(exists=True, readable=True, resolve_path=True),
     help="path to directory containing local rules.",
+    multiple=True,
 )
 @click.option(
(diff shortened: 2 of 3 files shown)


RamilCDISC (approved):
The PR updates -lr flag to allow users specify multiple local rule paths. The updated flag was validated by:

1. Passing only one local rule.
2. Passing multiple local rules when all rule files are valid and exist.
3. Passing multiple local rules where one or more files do not exist.
4. Passing multiple local rules that exist but are of different standards and versions.
5. Passing multiple local rules where at least one file has invalid extension.
6. Passing no local rules specified.

PR #1302, opened 2025-08-18
Performance 1233

pendingintent:
These changes address the parent_entity value missing for the top-level attributes:

1. In the Study dataset, parent_entity should be populated with "Wrapper"
2. In the string dataset, parent_entity should be:
- "Wrapper" where parent_rel is "usdmVersion", "systemName" or "systemVersion"
- "Study" where:
- - parent_rel is "name" and value is "Study_CDISC PILOT - LZZT"
- - parent_rel is "instanceType" and value is "Study"
- In the StudyVersion dataset, parent_entity should be "Study" for any record where rel_type is "definition"
- In the StudyProtocolDocument (USDM v3) or StudyDefinitionDocument (USDM v4) dataset, parent_entity should be "Study" for any record where rel_type is "definition".

### DIFF
--- a/cdisc_rules_engine/services/data_services/usdm_data_service.py
+++ b/cdisc_rules_engine/services/data_services/usdm_data_service.py
@@ -287,12 +287,13 @@ class USDMDataService(BaseDataService):
         rel_type = getattr(node, "type", "")
         # 'Wrapper' for all top-level Study attributes (any path starting with '`this`' or with no dot)
         if path.startswith("`this`") or "." not in path:
-            parent_entity = self.entity_dict["`this`"]
+            # parent_entity = self.entity_dict["`this`"]
+            parent_entity = self.entity_dict.get("`this`")
         # If still not set, use schema mapping for node type or key
         if not parent_entity:
             node_type = getattr(node, "type", None)
             if node_type and node_type in self.entity_dict:
-                parent_entity = self.entity_dict[node_type]
+                parent_entity = self.entity_dict.get(node_type)
             elif path:
                 key = path.split(".")[-1] if "." in path else path
                 parent_entity = self.entity_dict.get(key, key)


cdisc_rules_engine/services/data_services/usdm_data_service.py
```diff
@@ -267,9 +267,12 @@ def __get_record_metadata(self, node) -> dict:
         parent_entity = ""
         parent_id = ""
         while parent and isinstance(parent, dict):
-            if parent.get("instanceType") and parent.get("id"):
-                parent_entity = parent["instanceType"]
-                parent_id = parent["id"]
+            if parent.get("instanceType"):
```
RamilCDISC, line 270: Could you please mention why did you remove the parent.get("id") check here?

cdisc_rules_engine/services/data_services/usdm_data_service.py
```diff
@@ -282,6 +285,18 @@ def __get_record_metadata(self, node) -> dict:
             path_no_index.split(".")[-1] if "." in path_no_index else path_no_index
         )
         rel_type = getattr(node, "type", "")
+        # 'Wrapper' for all top-level Study attributes (any path starting with '`this`' or with no dot)
+        if path.startswith("`this`") or "." not in path:
+            parent_entity = self.entity_dict["`this`"]
```
RamilCDISC: for consistency in code could you please make this using .get()?

cdisc_rules_engine/services/data_services/usdm_data_service.py
```diff
@@ -282,6 +285,18 @@ def __get_record_metadata(self, node) -> dict:
...
+        # If still not set, use schema mapping for node type or key
+        if not parent_entity:
+            node_type = getattr(node, "type", None)
+            if node_type and node_type in self.entity_dict:
+                parent_entity = self.entity_dict[node_type]
```
RamilCDISC: same here could you please use .get()

cdisc_rules_engine/services/data_services/usdm_data_service.py
pendingintent: The particular line as well as the logic has been replaced.  I now use `parent_id = parent.get("id", "")`, which is a safer pattern preventing a KeyError.  parent_id is always set to the actual value or an empty string.  For example, when parent_entity is "Wrapper", the parent_id will be an empty string.

cdisc_rules_engine/services/data_services/usdm_data_service.py
pendingintent: Changed to `parent_entity = self.entity_dict.get("`this`")`

cdisc_rules_engine/services/data_services/usdm_data_service.py
pendingintent: Changed to `parent_entity = self.entity_dict.get(node_type)`

pendingintent:
The parent_entity values were not being populated for the following datasets:

- Study (A2 is now populated with 'Wrapper')
- string (A2:A4 now populated with 'Wrapper'; A5:A6 now populated with 'Study')
- null (A2:A4 now populated with 'Study')
- StudyVersion (A2 now populated with 'Study')
- StudyProtocolDocument (A2 now populated with 'Study')

Following this fix, the expected parent_entity values are populated as they were in the previous release version of usdm_data_service.py.

The output can be confirmed by opening the XLSX file that is produced: ./<study_name>-YYYY-MM-ddTHH-mm-ss.xlsx.

The worksheets where changes appear are named after the data sets, i.e., Study, string, null, StudyVersion, StudyProtocolDocument.

RamilCDISC (approved):
The PR fixes the parent_entity value missing issue for USDM datasets. The validation was done by:

1. Reviewing the code for any unwanted code or comments. 
2. Reviewing the updated logic for any potential or hidden bugs.
3. Validating that all unit test pass.
4. Validating all regression tests pass.
5. Validating the successful running of test suite and proper comparison of generated reports. 
6. Validating that the performance of engine for USDM rules is not affected.

Issue #682, opened 2024-03-11
Create DDF specific reporting template


nickdedonder:

ASL-rmarshall:
@ASL-rmarshall This is the ticket for the template

ASL-rmarshall:
Here's an updated version of the template for reporting USDM rule output: [CORE-ReportDDF_NewProposed_v2.0.xlsx](https://github.com/user-attachments/files/22388020/CORE-ReportDDF_NewProposed_v2.0.xlsx)

Although the final goal is to have all USDM rules implemented as JSONata rules, for now this template will need to handle output from both "normal" rules (based on a tabular dataset structures) and JSONata rules (when the functionality is available).

When defined, we'll need to know the names of the output attributes that will be mapped to the "Entity" and "Instance id" columns in the Issue Details tab (so we can include these in JSONata rule output).

FYI @nickdedonder @gerrycampion

PR #1304, opened 2025-08-19
1104 dataservice and datareader for arbitrary json files

gerrycampion:


cdisc_rules_engine/utilities/jsonata_processor.py
```diff
@@ -0,0 +1,72 @@
...
+        errors = (
+            [
+                ValidationErrorEntity(
+                    value=result,
+                    dataset=dataset_metadata.name,
+                    row=result.get("path"),
+                    usubjid=result.get("id"),
+                    sequence=result.get("iid"),
+                )
+                for result in results
```
RamilCDISC: After going through JSONata library my understanding is that the expr.evaluate() can return return list, numeric number or boolean depending on the query and data. But here we assume it will be a list or None. If a query returns single element, the code will break.

cdisc_rules_engine/utilities/jsonata_processor.py
```diff
@@ -0,0 +1,72 @@
...
+        return [validation_error_container.to_representation()]
+
+    @staticmethod
+    @cache
+    def get_custom_functions(jsonata_functions_path):
+        if not jsonata_functions_path:
+            return ""
+        functions = []
+        for filepath in glob(f"{jsonata_functions_path}/*.jsonata"):
+            with open(filepath, "r") as file:
```
RamilCDISC: There is no handling of bad jsonata files. Maybe we can introduce a try except block with a clean message if some jsonata files are bad.

cdisc_rules_engine/utilities/jsonata_processor.py
```diff
@@ -0,0 +1,72 @@
...
+                function_definition = file.read()
+                function_definition = function_definition.replace("{", "", 1)
```
RamilCDISC: Also a bad jsonata file may break tis logic as it may have multiple braces

RamilCDISC (changes requested):
please see the comments below.

cdisc_rules_engine/services/data_readers/json_reader.py
```diff
@@ -4,7 +4,7 @@
 )
 
 
-class USDMJSONReader(DataReaderInterface):
+class JSONReader(DataReaderInterface):
     def from_file(self, file_path):
         with open(file_path) as fp:
```
ASL-rmarshall: This is not specifically related to this PR, but as the JSON reader is being updated, I thought I'd mention it.  I've often found that reading of USDM JSON files fails if the file contains special characters.  I've had to fix this locally by updating this line in the reader to `with open(file_path, encoding='utf-8') as fp:`.

Could this change be implemented here?  According to the [JSON specification](https://www.ietf.org/rfc/rfc4627.txt):
> JSON text SHALL be encoded in Unicode.  The default encoding is UTF-8.

cdisc_rules_engine/utilities/jsonata_processor.py
```diff
@@ -0,0 +1,72 @@
...
+        expr = Jsonata(full_string)
+        results = expr.evaluate(dataset)
+        errors = (
+            [
+                ValidationErrorEntity(
+                    value=result,
+                    dataset=dataset_metadata.name,
+                    row=result.get("path"),
+                    usubjid=result.get("id"),
+                    sequence=result.get("iid"),
```
ASL-rmarshall: I don't think `iid` should be referenced here - `iid` is a temporary attribute that I've only added into some of the rules where a `path` needs to be calculated.  I added it because there's a chance that the file contains duplicate `id` values, so I needed a way to reliably identify unique object instances (which is needed by the `get_path` function because it builds the path by iteratively searching for the unique object from the top down through the JSON hierarchy).

If we're going to rely on something like `iid` to unique identify instances for the reporting of errors (which I'm not sure would be particularly helpful as it's not present in the original JSON), then I think the "instance identifer" would need to be added as part of the generic JSON pre-processing, and not as part of the rule logic.  In this case, the attribute would need a name that's not likely to be used in the rule logic (e.g., `_iid`).

Is a value for `sequence` required? If so, we'd either need to figure out what would be a helpful value that's available for all rules, or we'd need to allow this to be assigned on a rule-by-rule basis.

cdisc_rules_engine/utilities/jsonata_processor.py
```diff
@@ -0,0 +1,72 @@
...
+                ExecutionStatus.SUCCESS.value
+                if results
+                else ExecutionStatus.EXECUTION_ERROR.value
+            ),
+        )
+        return [validation_error_container.to_representation()]
+
+    @staticmethod
+    @cache
+    def get_custom_functions(jsonata_functions_path):
```
ASL-rmarshall: As some of these functions will be required for successful execution of rules published in the CDISC Library, shouldn't any external jsonata function be available within the engine file structure (e.g., in `resources` or a specific sub-folder)?  

Users should also be able specify a location for custom jsonata functions referenced by their own custom rules.  Different function variables might be a good idea - e.g., `$utils` for "standard" functions that are required for published CDISC rules and something like `$custom_utils` for user-defined functions.

resources/schema/CORE-base.json
```diff
@@ -41,6 +41,9 @@
         },
         {
           "$ref": "Operator.json"
+        },
+        {
+          "type": "string"
```
ASL-rmarshall: I think this is the wrong place to specify that "string" is allowed if the JSONata statements are expected immediately after `Check` - this would mean that you'd have to have `all`, `any` or `not` (`Boolean`) under `Check` and it would then allow a string to be specified at any level within the nested condition.

Instead, line 362-364 should be changed to:
```json
    "Check": {
      "anyOf": [
        {
          "$ref": "#/$defs/Boolean"
        },
        {
          "type": "string"
        }
      ]
    },
```

resources/schema/Rule_Type.md
```diff
@@ -179,6 +179,97 @@ all:
...
+```
+
+### Output Variables and Report column mapping
+
+You can use `Outcome.Output Variables` to specify which properties to display from the result JSON. The following result property names will map to the column names in the Excel output report.
+
+Result property name -> Report Issue Details Column Name:
+
+- `id` -> `USUBJID`
+- `path` -> `Record`
```
ASL-rmarshall: As mentioned, `path` is expensive to calculate so it would be better not to assume that it will be present for all outputs

cdisc_rules_engine/utilities/jsonata_processor.py
```diff
@@ -0,0 +1,72 @@
...
+)
+
+
+class JSONataProcessor:
+
+    @staticmethod
+    def execute_jsonata_rule(
+        rule: dict,
+        dataset: dict,
+        dataset_metadata: SDTMDatasetMetadata,
```
ASL-rmarshall: It doesn't seem right to use `SDTMDatasetMetadata` here - none of the properties of `SDTMDatasetMetadata` is relevant for USDM entities.

cdisc_rules_engine/services/data_readers/json_reader.py
ASL-rmarshall: Note that the `is_valid_data` method of the `USDMDataService` also reads in the specified JSON path and has the same encoding issue.  It would be better if `is_valid_data` used the `JSONReader`.

It would also be better if errors were handled more elegantly when the specified file is not valid JSON.

cdisc_rules_engine/utilities/jsonata_processor.py
gerrycampion: This will give an "Error occurred during validation." Is that a problem?

RamilCDISC: Maybe we can better handle the case and output a more explanatory error message.

cdisc_rules_engine/utilities/jsonata_processor.py
gerrycampion: the variable names are more closely tied to their column names in the report now and none are required. The names may change when we get a new report template.

cdisc_rules_engine/utilities/jsonata_processor.py
gerrycampion: i no longer need this and dropped it.

cdisc_rules_engine/utilities/jsonata_processor.py
gerrycampion: I was also thinking maybe a preprocessing step could add the jsonpath property for each object. Do you think this is worthwhile?

resources/schema/Rule_Type.md
gerrycampion: this is not required

cdisc_rules_engine/utilities/jsonata_processor.py
ASL-rmarshall: The new proposed template is now available [here](https://github.com/cdisc-org/cdisc-rules-engine/issues/682#issuecomment-3303359692).

Berber and I discussed the option of adding a pre-processing step to add jsonpath. We don't think this is really necessary - having the entity/class name and the instance id will usually be enough.  If we think the path is needed for specific rules (e.g., the one that checks for duplicate id values), then this can be added as a specific output variable for the rule.

cdisc_rules_engine/services/data_readers/json_reader.py
gerrycampion: - I updated to use open(file_path, "rb"), so that it is opened in binary instead of using the locale decoding and decoding is detected correctly by the json loader instead.
- I changed other code to use the shared json loader
- added error handling - better messaging when loading json in a rule and during complete fail due to loading the service.

cdisc_rules_engine/utilities/jsonata_processor.py
gerrycampion: updated

cdisc_rules_engine/utilities/jsonata_processor.py
gerrycampion: updated

cdisc_rules_engine/utilities/jsonata_processor.py
gerrycampion: updated

resources/schema/CORE-base.json
gerrycampion: good catch. fixed

cdisc_rules_engine/utilities/jsonata_processor.py
gerrycampion: builtins are now added as $utils. External can be added with `-jcf myutils path/to/myutils -jcf yourutils path/to/yourutils`

resources/jsonata/get_idmap.jsonata
ASL-rmarshall, line 1: This file can be removed - it is used for test data generation, not for rules.

resources/jsonata/get_maxids.jsonata
ASL-rmarshall, line 1: This file can be removed - it is used for test data generation, not for rules.

resources/jsonata/get_rule.jsonata
ASL-rmarshall, line 1: This file can be removed - it is used for rule testing, not for data checking.

resources/jsonata/make_dup.jsonata
ASL-rmarshall, line 1: This file can be removed - it is used for test data generation, not for rules.

resources/jsonata/report.jsonata
ASL-rmarshall, line 1: This file can be removed - it is used for rule testing, not for data checking.

cdisc_rules_engine/rules_engine.py
```diff
@@ -101,16 +106,23 @@ def validate_single_rule(self, rule: dict, datasets: Iterable[SDTMDatasetMetadat
...
-                include_split = rule["domains"].get("include_split_datasets", False)
-                if not include_split:
-                    continue  # handling split datasets
-            results[dataset_metadata.unsplit_name] = self.validate_single_dataset(
+        if rule.get("rule_type") == RuleTypes.JSONATA.value:
+            results["json"] = self.validate_single_dataset(
                 rule,
                 datasets,
-                dataset_metadata,
+                SDTMDatasetMetadata(name="json"),
```
ASL-rmarshall, line 114: Assigning the dataset name as "json" causes all rules to be skipped unless their Scope is Entities: Include: ALL.  I think `rule_applies_to_entity` needs to be updated to account for JSONata rules - something like:
```python
if rule.get("rule_type") == RuleTypes.JSONATA.value and dataset_metadata.name == "json":
    return any(rule_applies_to_entity(dm,rule) for dm in datasets)
```
i.e., if this is a JSONata rule and you're passing the whole json, then check that there's at least one entity in the json that satisfies the Scope condition.

resources/schema/Rule_Type.md
```diff
@@ -179,6 +179,106 @@ all:
...
+### Output Variables and Report column mapping
+
+You can use `Outcome.Output Variables` to specify which properties to display from the result JSON. The following result property names will map to the column names in the Excel output report.
+
+Result property name -> Report Issue Details Column Name:
+
+- `dataset` -> `Dataset`
+- `usubjid` -> `USUBJID`
+- `record` -> `Record`
+- `sequence` -> `Sequence`
```
ASL-rmarshall: I think these will need to be reviewed and possibly changed to align with the updated USDM Excel output report.  The column names will definitely need to be changed.  We may also need to update the property names to align better with what we want to report for USDM data (e.g., `id` instead of `usubjid`) - but this may depend on the engine functionality when running rules for USDM tabular datasets (are the properties just picked up from the results dataset columns?).

It might also be useful to specify the mapping(s) to JSON output properties.  I noticed above that `record` seems to be mapped to `row` and I assume that a `dataset` properties in the jsonata output would be represented as `dataset` in the JSON output.  Are there any other JSON output properties that could/should be assigned?

resources/schema/Rule_Type.md
```diff
@@ -179,6 +179,106 @@ all:
...
+
+### Scope
+
+`Scope` should always `Include` `ALL` to ensure the rule will be run. The rule is only run once for the entire JSON file. The `Dataset` determination must come from the rule's jsonata result property.
```
ASL-rmarshall: I think the expectation is that Scope will be defined in the usual way and that the rule will only run if the JSON file contains at least one instance of an entity that satisfies the Scope condition.

resources/schema/Rule_Type.md
```diff
@@ -179,6 +179,106 @@ all:
...
+### Example
+
+#### Rule
+
+```yaml
+Check: |
+  **.$filter($, $myutils.equals).{"record":path, "A":A, "B":B}
+Core:
+  Id: JSONATA Test
+Status: Draft
```
ASL-rmarshall: Status should be indented

ASL-rmarshall (changes requested):
I'm checking with Berber to confirm if she has the same expectations about Scope for JSONata rules - I'll update here when I get a response.

cdisc_rules_engine/rules_engine.py
ASL-rmarshall: @gerrycampion I discussed this with Berber. She agrees that Scope should be specified as usual, but thinks that it is not necessary to prevent execution of the rule if there is no entity that meets the Scope criteria.  For example the suggested code above could become just:
```python
if rule.get("rule_type") == RuleTypes.JSONATA.value and dataset_metadata.name == "json":
    return true
```
i.e., if this is a JSONata rule and you're passing the whole json, then run the rule (once).

resources/schema/Rule_Type.md
ASL-rmarshall: @gerrycampion As mentioned above, it should be possible for Scope to be defined in the usual way with the JSONata still only executing once.  However, this no longer an immediate requirement:
>the rule will only run if the JSON file contains at least one instance of an entity that satisfies the Scope condition.

The JSONata rule should always execute once, regardless of what is specific in the Scope.

resources/schema/Rule_Type.md
gerrycampion: I updated this to simplify it even further I think. Now, the jsonata columns will be mapped to the json report properties of the same name/capitalization. I converted this mapping to a table. We will add additional rows/columns to this table in the usdm template update. There are currently no additional JSON output properties to be assigned, but we will need to add some for the usdm report.

cdisc_rules_engine/rules_engine.py
gerrycampion: I followed this logic, but I pulled it outside of the "Entity" logic so that it will work the same way for any standard.

resources/jsonata/get_path.jsonata
ASL-rmarshall, line 1: As agreed at the daily stand-up meeting on 2025-10-09, a `path` attribute will be added to every instance of every class within the imported JSON so the `get_path` JSONata function will no longer be necessary.  Please remove this file.

When adding the `path` attribute, no other attributes should be changed (or removed).  I wrote a JSONata function to do this ([here](https://github.com/cdisc-org/cdisc-jsonata-rules/blob/main/utils/add_path.jsonata)), but native python processing might be quicker.

resources/jsonata/parent_rel.jsonata
ASL-rmarshall, line 1: This function is currently not used in any rule, so the file can be removed.

PR #1306, opened 2025-08-20
cg0314

SFJohnson24:


cdisc_rules_engine/models/rule.py
```diff
@@ -54,6 +54,7 @@ def from_cdisc_metadata(cls, rule_metadata: dict) -> dict:
                 "conditions": cls.parse_conditions(rule_metadata.get("Check")),
                 "actions": cls.parse_actions(rule_metadata.get("Outcome")),
                 "use_case": rule_metadata.get("Scope", {}).get("Use_Case"),
+                "status": rule_metadata.get("Core", {}).get("Status", {}),
```
RamilCDISC, line 57: here if rule is missing status {} is returned. Later is script utils we have .lower() on status. If you could please update either the script util logic or this here to better handle this case.

RamilCDISC (changes requested):
please see the comment below

cdisc_rules_engine/models/rule.py
SFJohnson24: i updated to return a blank string instead of None to avoid the error, nice catch

RamilCDISC: thanks Sam

RamilCDISC (approved):
The PR updates the engine to let run draft rules without checking for standard and version mentioned. It also updated to add support for special purpose dataset and implements some fixes for dataset access. The PR was validated by:

1. Reviewing the updates for any unwanted code or comments.
2. Reviewing the new logic implemented to find any potential bugs.
3. Running validation after deploying the updated branch on test endpoint.
4. Running a validation locally for certifications datasets.
5. Running a validation for rule marked as both draft and published.
6. Ensuring relevant unit test has been updated.
7. Validating passing of all unit tests.
8. Validating passing of the test suite.

PR #1314, opened 2025-08-27
updated docs

SFJohnson24:


resources/schema/Operator.md
```diff
@@ -277,6 +277,52 @@ Complement of `suffix_equal_to`
...
+
+Example:
+
+```yaml
+- name: "--TOXGR"
+  operator: "does_not_contain_case_insensitive"
+  value: "grade" # True if no case variation of 'grade' exists in the list
+```
+
+## contains
```
RamilCDISC: There is repetition of the updated operators and they show the same example. I guess the old ones are yet to be removed.

RamilCDISC (changes requested):
Please see the below comment

PR #1334, opened 2025-09-12
1326: check codelist terms by preferred term

alexfurmenkov:
This PR extends the `CodelistTerms` operation with the opportunity to merge the dataset and the library metadata based on the preferred term, this is the operation example:

```
  - codelist_code: studyPhase.standardCode.code
    ct_package_type: SDTM
    id: $code_for_decode
    operator: codelist_terms
    term_pref_term: studyPhase.standardCode.decode
    ct_version: studyPhase.standardCode.codeSystemVersion
```


cdisc_rules_engine/operations/codelist_terms.py
```diff
@@ -15,16 +15,24 @@ def _execute_operation(self) -> pd.Series:
...
     def _handle_multiple_versions(self) -> pd.Series:
-        if self.params.term_code and self.params.term_value:
+        if (
+            self.params.term_code
+            and self.params.term_value
+            and self.params.term_pref_term
+        ):
             raise RuleExecutionError(
-                "Both term_code and term_value cannot be specified at the same time."
+                "term_code, term_pref_term and term_value cannot be specified at the same time."
```
ASL-rmarshall: I think the checking logic and message need to be modified a bit here.  The requirement is that only one of `term_code`, `term_value` or `term_pref_term` may be specified so the error message should be something like:
> More than one of term_code, term_pref_term and term_value cannot be specified at the same time.

At the moment, the error would not be raised if (for example) `term_code` and `term_value` were specified and `term_pref_term` was not.

tests/unit/test_operations/test_codelist_terms.py
```diff
@@ -190,29 +206,60 @@ def test_empty_terms(operation_params):
...
+            ("T1", nan, "T3", "T4"),
+        ),
+        (
+            "mock_package",
+            "codelist_code",
+            "t_code",
+            "t_value",
+            "t_pref_term",
+            RuleExecutionError,
+        ),
```
ASL-rmarshall, line 234: It would be good also to have test cases showing that specifying two of `term_code`, `term_value` and `term_pref_term` also raise a `RuleExecutionError`.

cdisc_rules_engine/operations/codelist_terms.py
```diff
@@ -34,6 +42,10 @@ def _handle_multiple_versions(self) -> pd.Series:
             left_on = self.params.term_value
             right_on = "term_value"
             target = "term_code"
+        elif self.params.term_pref_term:
+            left_on = self.params.term_pref_term
+            right_on = "term_pref_term"
+            target = "term_code"
```
ASL-rmarshall: Looking at this now, I realise that we will need to use the same `returntype` parameter (used by `_handle_single_version`) to indicate whether it's the `term_code` (`returntype`="code"), `term_value` (`returntype`="value") or `term_pref_term` (`returntype`="pref_term", which is a new option) that's returned.  At the moment:
- `term_code` is returned if you specify either `term_value` or `term_pref_term`
- `term_value` is returned if you specify `term_code`

There needs to be a way to indicate that `term_pref_term` is returned (instead of `term_value`) when you specify `term_code` (or even that you want to look up value/pref_term by pref_term/value respectively).

Theoretically, it would be OK to use any `returntype` value with any of `term_code`, `term_value` or `term_pref_term` - even to indicate the same return type as lookup type (e.g., using `returntype`="code" with `term_code`).  This is because the match-merging is case-insensitive so you could use the operation to look up the correct case-sensitive value from the codelist for the corresponding (case-insensitive) data value.

ASL-rmarshall:
Also, [Operations.md](https://github.com/cdisc-org/cdisc-rules-engine/blob/main/resources/schema/Operations.md) will need to be updated to describe the additional/changed functionality.

tests/unit/test_operations/test_codelist_terms.py
alexfurmenkov: Added, pls check

cdisc_rules_engine/operations/codelist_terms.py
alexfurmenkov: Updated, please check

PR #1352, opened 2025-09-23
1326: codelist preferred term check

alexfurmenkov:


tests/unit/test_operations/test_codelist_terms.py
```diff
@@ -190,29 +206,68 @@ def test_empty_terms(operation_params):
 
 
 @pytest.mark.parametrize(
-    "package_type, codelist_code, term_code, term_value, expected",
+    "package_type, codelist_code, term_code, term_value, term_pref_term, expected",
```
ASL-rmarshall: Ideally, there should be some tests to confirm that the `returntype` parameter has the expected effect.

resources/schema/Operations.json
```diff
@@ -533,6 +533,9 @@
     "term_code": {
       "type": "string"
     },
+    "term_pref_term": {
```
ASL-rmarshall, line 536: I think line 528 also needs to be updated to allow "pref_term" as a value for the `returntype` parameter.

cdisc_rules_engine/operations/codelist_terms.py
```diff
@@ -15,25 +15,42 @@ def _execute_operation(self) -> pd.Series:
...
+        if params_count > 1:
             raise RuleExecutionError(
-                "Both term_code and term_value cannot be specified at the same time."
+                "More than one of term_code, term_pref_term and term_value cannot be specified at the same time."
             )
         elif self.params.term_code:
             left_on = self.params.term_code
             right_on = "term_code"
-            target = "term_value"
+            target = f"term_{self.params.returntype or 'value'}"
```
ASL-rmarshall, line 45: Now that we allow `returntype` to be "pref_term" (in addition to "code" or "value"), I think the alternative functionality of the `codelist_terms` operation (handling a single version) will need to be updated to prevent unexpected results - either by adding a check and returning an error if "pref_term" is used or, preferably, by updating it to return the preferred term when "pref_term" is specified. For example, change `_get_codelist_values` from
```python
            if codelist_level == "codelist":
                if code_obj.get("term") == "N/A":
                    if check == "code":
                        values.append(codelist_id)
                    else:
                        values.append(codelist_info["submissionValue"])
            elif codelist_level == "term":
                terms = codelist_info.get("terms", [])
                for term in terms:
                    if check == "value":
                        values.append(term["submissionValue"])
                    else:
                        values.append(term["conceptId"])
```
to
```python
            if codelist_level == "codelist":
                if code_obj.get("term") == "N/A":
                    if check == "code":
                        values.append(codelist_id)
                    elif check == "pref_term":
                        values.append(codelist_info["preferredTerm"])
                    else:
                        values.append(codelist_info["submissionValue"])
            elif codelist_level == "term":
                terms = codelist_info.get("terms", [])
                for term in terms:
                    if check == "value":
                        values.append(term["submissionValue"])
                    elif check == "pref_term":
                        values.append(term["preferredTerm"])
                    else:
                        values.append(term["conceptId"])
```

tests/unit/test_operations/test_codelist_terms.py
alexfurmenkov: Updated

resources/schema/Operations.json
alexfurmenkov: Updated

cdisc_rules_engine/operations/codelist_terms.py
alexfurmenkov: Updated

PR #1358, opened 2025-09-30
Pympler fallback

SFJohnson24:
This PR adds a fallback for the asizeof() pympler which estimates negative memory by falling back onto sys.getsizeof 

This pull request improves the robustness of the in-memory cache service by introducing a safer way to calculate object sizes, preventing potential errors when estimating memory usage. The main change is the addition of a custom size calculation function that gracefully falls back to a default method if the preferred one fails.

**Improvements to object size calculation:**

* Added a new `cust_asizeof` function that tries to use `asizeof.asizeof` for object size estimation, but falls back to `sys.getsizeof` and logs a warning if an exception occurs. This prevents crashes when `asizeof` cannot handle certain object types.
* Updated the cache initialization in the `InMemoryCacheService` class to use `cust_asizeof` instead of `asizeof.asizeof` for the LRU cache's `getsizeof` parameter, improving reliability.
* Modified the `get_data_size` function to use `cust_asizeof` instead of `asizeof.asizeof` for non-`DatasetInterface` objects, ensuring consistent error handling.

to test: run cg0006 with test data I sent to @RamilCDISC via slack

### DIFF
--- a/cdisc_rules_engine/services/cache/in_memory_cache_service.py
+++ b/cdisc_rules_engine/services/cache/in_memory_cache_service.py
@@ -15,11 +15,7 @@ from cdisc_rules_engine.services import logger
 def cust_asizeof(obj):
     try:
         return asizeof.asizeof(obj)
-    except (ValueError, TypeError, AttributeError) as e:
-        logger.warning(
-            f"asizeof failed for object type {type(obj).__name__}: {e}. "
-            "Using sys.getsizeof fallback."
-        )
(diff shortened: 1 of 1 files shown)


RamilCDISC (commented):
I was running the validation with the data provided. When i run with using -d flag for all the datasets in one folder. I can see the warning being printed and "asizeof" of function failing. Then i tried running for each dataset separately. I did not see any warning message appearing. Although the behavior should be same here. Kindly see the screenshot. 

Image

PR #1359, opened 2025-10-02
update

SFJohnson24:
This pull request introduces two major new configuration options for validation and reporting: limiting the number of rows per Excel report file and capping the number of errors reported per rule. These options can be set via CLI flags or environment variables, and the code now supports splitting Excel reports into multiple files when data exceeds the configured row limit. The changes are reflected in the CLI, reporting logic, and rule validation flow.

### Reporting enhancements

* Added support for splitting Excel reports into multiple files when the number of rows exceeds a configurable maximum (`max_report_rows`). This can be set via the CLI (`-mr`) or the `MAX_REPORT_ROWS` environment variable, and is documented in `README.md` and `env.example`. The logic ensures the larger value between CLI and env is used, and disables the limit if set to 0. [[1]](diffhunk://#diff-2b9e25f1b7868b885dd673f93e87baa01ebcc24c29f828f8cdc1f5f302b330dfR32-R33) [[2]](diffhunk://#diff-2b9e25f1b7868b885dd673f93e87baa01ebcc24c29f828f8cdc1f5f302b330dfR47-R224) [[3]](diffhunk://#diff-2b9e25f1b7868b885dd673f93e87baa01ebcc24c29f828f8cdc1f5f302b330dfL140-R334) [[4]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R155-R167) [[5]](diffhunk://#diff-0521255bbda96fd681ef7be9c0d23f04cb3838b3a905ed7f25b041ec34cf547cR3-R4) [[6]](diffhunk://#diff-935bc67832a22823701c35975421947be81615771e83d7ca168792058a15a347R208-R221)
* Refactored `ExcelReport` to handle multiple output files, updating file naming and metadata to reflect report parts when splitting occurs. [[1]](diffhunk://#diff-2b9e25f1b7868b885dd673f93e87baa01ebcc24c29f828f8cdc1f5f302b330dfR47-R224) [[2]](diffhunk://#diff-2b9e25f1b7868b885dd673f93e87baa01ebcc24c29f828f8cdc1f5f302b330dfL140-R334)

### Validation logic improvements

* Introduced a configurable soft maximum for errors reported per rule (`max_errors_per_rule`), settable via CLI (`-me`) or `MAX_ERRORS_PER_RULE` env variable, with logic to halt validation for a rule once the limit is reached. Defaults to 1000 if unspecified; disables the limit if set to 0. [[1]](diffhunk://#diff-61bbdda116cb3d9e8bce503b115f0c16fe154902e296785925d922821eae917bR105-R135) [[2]](diffhunk://#diff-61bbdda116cb3d9e8bce503b115f0c16fe154902e296785925d922821eae917bR95) [[3]](diffhunk://#diff-b11312ab2759e9cf8aa448be78c80f1cb9e6be99642ea9c78a93fc7a2e850146R26-R27) [[4]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R155-R167) [[5]](diffhunk://#diff-0521255bbda96fd681ef7be9c0d23f04cb3838b3a905ed7f25b041ec34cf547cR3-R4) [[6]](diffhunk://#diff-645421107f064022b54581bdaf972ee1baa090acef121e83d4b27be3f50ed802R487-R508)
* Added helper function `set_max_errors_per_rule` to centralize max error logic, and integrated it into both CLI and script-based validation flows. [[1]](diffhunk://#diff-645421107f064022b54581bdaf972ee1baa090acef121e83d4b27be3f50ed802R487-R508) [[2]](diffhunk://#diff-4a7a926a43d976bc2fc0cfb78839df350e02f9e6191061859727d48c01101f66R47) [[3]](diffhunk://#diff-4a7a926a43d976bc2fc0cfb78839df350e02f9e6191061859727d48c01101f66R80) [[4]](diffhunk://#diff-4a7a926a43d976bc2fc0cfb78839df350e02f9e6191061859727d48c01101f66R94)

### CLI and documentation updates

* Updated CLI (`core.py`) to accept new flags for controlling report row and error limits, and ensured environment variables are loaded for these options. Also updated test and validation argument handling to support the new parameters. [[1]](diffhunk://#diff-935bc67832a22823701c35975421947be81615771e83d7ca168792058a15a347R208-R221) [[2]](diffhunk://#diff-935bc67832a22823701c35975421947be81615771e83d7ca168792058a15a347R253-R254) [[3]](diffhunk://#diff-935bc67832a22823701c35975421947be81615771e83d7ca168792058a15a347R266) [[4]](diffhunk://#diff-935bc67832a22823701c35975421947be81615771e83d7ca168792058a15a347R341-R342) [[5]](diffhunk://#diff-935bc67832a22823701c35975421947be81615771e83d7ca168792058a15a347R688-R689) [[6]](diffhunk://#diff-935bc67832a22823701c35975421947be81615771e83d7ca168792058a15a347R713-R714) [[7]](diffhunk://#diff-935bc67832a22823701c35975421947be81615771e83d7ca168792058a15a347R741-R742) [[8]](diffhunk://#diff-b11312ab2759e9cf8aa448be78c80f1cb9e6be99642ea9c78a93fc7a2e850146R26-R27)
* Expanded `README.md` and `env.example` to document the new options and their behaviors. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R155-R167) [[2]](diffhunk://#diff-0521255bbda96fd681ef7be9c0d23f04cb3838b3a905ed7f25b041ec34cf547cR3-R4)

### DIFF
--- a/core.py
+++ b/core.py
@@ -579,7 +579,7 @@ def list_rule_sets(ctx: click.Context, cache_path: str, custom: bool):
     for standard in sorted(rule_sets.keys()):
         versions = sorted(rule_sets[standard])
         for version in versions:
-            print(f"{standard.upper()}, {version}")
+            print(f"{standard}, {version}")
 
 
 @click.command()
--- a/tests/QARegressionTests/test_core/test_list_rule_sets.py
+++ b/tests/QARegressionTests/test_core/test_list_rule_sets.py
@@ -13,10 +13,10 @@ class TestListRuleSets(unittest.TestCase):
         result = self.runner.invoke(
             list_rule_sets, ["-c", os.path.join("resources", "cache")]
         )
-        self.assertIn("SDTMIG, 3-2", result.output)
-        self.assertIn("SDTMIG, 3-3", result.output)
-        self.assertIn("SDTMIG, 3-4", result.output)
-        self.assertIn("SENDIG, 3-1", result.output)
+        self.assertIn("sdtmig, 3-2", result.output)
+        self.assertIn("sdtmig, 3-3", result.output)
+        self.assertIn("sdtmig, 3-4", result.output)
+        self.assertIn("sendig, 3-1", result.output)
 
     def test_list_rule_sets_invalid_cache_path(self):
         result = self.runner.invoke(
@@ -26,10 +26,10 @@ class TestListRuleSets(unittest.TestCase):
 
     def test_list_rule_sets_no_cache_path(self):
         result = self.runner.invoke(list_rule_sets, [])
-        self.assertIn("SDTMIG, 3-2", result.output)
-        self.assertIn("SDTMIG, 3-3", result.output)
-        self.assertIn("SDTMIG, 3-4", result.output)
-        self.assertIn("SENDIG, 3-1", result.output)
+        self.assertIn("sdtmig, 3-2", result.output)
+        self.assertIn("sdtmig, 3-3", result.output)
+        self.assertIn("sdtmig, 3-4", result.output)
+        self.assertIn("sendig, 3-1", result.output)
 
     def tearDown(self):
         tearDown()


cdisc_rules_engine/services/reporting/excel_report.py
```diff
@@ -41,27 +44,172 @@ def __init__(
...
+        standard,
+        version,
+        dictionary_versions,
+        **kwargs,
+    ) -> List[Workbook]:
+        """
+        Create multiple workbooks when data exceeds Excel's row limit.
+        """
+        workbooks = []
+        detailed_chunks = self._chunk_data(detailed_data, self.MAX_ROWS_PER_SHEET)
```
RamilCDISC: Could you lowercase the max_rows_per_sheet here as it is defined in the constructor.

RamilCDISC (changes requested):
Could you please see the below comment and will the rule editor also have a way to set the size of records in the report?

cdisc_rules_engine/services/reporting/excel_report.py
SFJohnson24: rule editor will not have a way to set the size--this is purely a feature for the CLI and the executable.  editor uses excel data and should not be getting a dataset with issues that would reach the row limit of excel

SFJohnson24: > Could you lowercase the max_rows_per_sheet here as it is defined in the constructor.

this is done

SFJohnson24:
> 1. I ran multiple validation with different test cases. I found that the issue summary page is being left empty from part2 onwards. Only the first part shows content on issue summary page.

I changed this so if it is shorter than the row limit, it will print it on every wb.  If it is longer than the limit, it will populate as much as it can but may leave blanks

>     2. Could we make the env variable name and CLI flag same?

done

>     3. Is the rules report sheet omitted from the max number of rows check?

yes--I could add this if we think this is necessary?  Number of issues and the issue reports seemed like the logical things to do as those could get quite long.  I didnt think that rules run would get to the point it needs to be split across multiple workbooks

SFJohnson24:
@RamilCDISC -- the limit was met or exceeded for a single dataset.  for suppec, it produced 80 errors which is >1 so it did not validate any of the other datasets against that rule.  If you look at the results, it is only validating 1 dataset per rule because once the first dataset is validated with issues, it exceeds the limit and the validation for that rule ceases for tthe remaining datasets.  This is expected behavior based off what you just pasted.

setting a limit and stopping validaition mid-dataset is much more difficult that after each dataset runs--which is why it is 'met or exceeded'.  It runs a dataset against a rule, checks the errors returns and sees if it is at or over the limit

env.example
```diff
@@ -1,2 +1,4 @@
 CDISC_LIBRARY_API_KEY = 012345
-DATASET_SIZE_THRESHOLD = size_in_bytes to force dask implementation
\ No newline at end of file
+DATASET_SIZE_THRESHOLD = size_in_bytes to force dask implementation
+MAX_REPORT_ROWS = maximum number of issues per excel sheet (plus headers) in result report
+MAX_ERRORS_PER_RULE = maximum number of errors to report per rule during a validation run.
```
RamilCDISC, line 4: here the env variable is MAX_ERRORS_PER_RULE but the readme file says MAX_REPORT_ERRORS. Could you please update the readme?

SFJohnson24: it was incorrect in the readme.  MAX_ERRORS_PER_RULE is correct--i believe that is more descriptive.

PR #1371, opened 2025-10-08
897: check that values of specified variable is valid XHTML 

alexfurmenkov:


cdisc_rules_engine/operations/get_xhtml_errors.py
```diff
@@ -0,0 +1,90 @@
...
+    Steps:
+      1. Make sure the column is a valid HTML -> on failure emit "Invalid HTML fragment:".
+      2. Make sure the column is a valid XML -> on failure emit "Invalid XML fragment:".
+      3. DTD validation for <usdm:ref> (attributes: attribute, id, klass) -> on failure emit "Invalid XML fragment:".
+
+    Empty / None values return an empty list.
+    """
+
+    usdm_dtd = """\
+    <!ELEMENT ref EMPTY>
```
ASL-rmarshall: We'll also need a definition for the `usdm:tag` element (which must have a `name` attribute) as described [here](https://wiki.cdisc.org/display/USDMIGv4/Syntax+Templates) and mentioned [here](https://cdisccore.slack.com/archives/D09EX2UGL01/p1758624283682419).

cdisc_rules_engine/operations/get_xhtml_errors.py
```diff
@@ -0,0 +1,90 @@
...
+            refs = root.xpath("//usdm:ref", namespaces={"usdm": "usdm"})
+        except Exception:
+            refs = []
+        if not refs:
+            return None
+        dtd = etree.DTD(io.StringIO(self.usdm_dtd))
+        for ref in refs:
+            ref_copy = etree.Element("ref", attrib=ref.attrib)
+            if not dtd.validate(ref_copy):
+                return f"Invalid XML fragment: {dtd.error_log.filter_from_errors()}"
```
ASL-rmarshall: This will stop the checking of refs as soon as the first invalid one is found.  However, it would be preferable if all invalid refs (and/or tags) are reported in one go - so maybe add the error for each invalid ref/tag to a list and return the list of errors if it's not empty.

cdisc_rules_engine/operations/get_xhtml_errors.py
```diff
@@ -0,0 +1,90 @@
...
+        id        ID    #REQUIRED
+        klass     CDATA #REQUIRED
+    >
+    """
+
+    def _execute_operation(self):
+        dataset = self.evaluation_dataset
+        target = self.params.target
+        if target not in dataset:
+            error_list = [f"Target column '{target}' not found"]
```
ASL-rmarshall: This seems strange to me - I don't think you should get an error reported for every record in the dataset if the specified column does not exist. I would expect this operation to behave in the same way as other operations - i.e., if the column specified in the `name` parameter is not present in the dataset, then this is dealt with by the error handling in `BaseOperation.execute` (e.g., the rule would be skipped because the specified variable does not exist).

cdisc_rules_engine/operations/get_xhtml_errors.py
```diff
@@ -0,0 +1,90 @@
...
+        text = value.strip()
+        if not text:
+            return []
+
+        html_error = self._validate_html(text)
+        if html_error:
+            return [html_error]
+
+        xml_text = text
+        if "usdm:ref" in text and "xmlns:usdm" not in text:
```
ASL-rmarshall: Will need to handle `usdm:tag` too (here and in any other place that deals with `usdm:ref`)

cdisc_rules_engine/operations/get_xhtml_errors.py
```diff
@@ -0,0 +1,90 @@
...
+        if dtd_error:
+            return [dtd_error]
+        return []
+
+    def _validate_html(self, text: str) -> Optional[str]:
+        try:
+            html5lib.parse(text)
+            return None
+        except Exception as e:
+            return f"Invalid HTML fragment: {e}"
```
ASL-rmarshall: If it's possible, it would be better to report all the HTML parsing issues in one go instead of stopping when the first is encountered.  However, this may depend on how the parser works and whether the parser can differentiate between errors that prevent continued processing and those that don't.

cdisc_rules_engine/operations/get_xhtml_errors.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def _validate_xml(
+        self, xml_text: str
+    ) -> Tuple[Optional[etree._Element], Optional[str]]:
+        try:
+            parser = etree.XMLParser(ns_clean=True)
+            root = etree.fromstring(xml_text.encode("utf-8"), parser=parser)
+            return root, None
+        except etree.XMLSyntaxError as e:
+            return None, f"Invalid XML fragment: {e}"
```
ASL-rmarshall: If it's possible, it would be better to report all the XML parsing issues in one go instead of stopping when the first is encountered.  However, this may depend on the type of error encountered and how the parser works - I think this particular parser stops when malformed XML is encountered.

tests/unit/test_operations/test_get_xhtml_errors.py
```diff
@@ -0,0 +1,88 @@
...
+        assert cell == []
+    elif expected_type == "invalid_html":
+        assert len(cell) == 1 and cell[0].startswith("Invalid HTML fragment:")
+    elif expected_type == "invalid_xml":
+        assert len(cell) == 1 and cell[0].startswith("Invalid XML fragment:")
+    else:
+        pytest.fail(f"Unknown expected_type {expected_type}")
+
+
+def test_get_xhtml_errors_missing_column(base_services):
```
ASL-rmarshall: This shouldn't be necessary if a missing variable is handled in the usual way (i.e., leading to an "absent variable skip"

cdisc_rules_engine/operations/get_xhtml_errors.py
alexfurmenkov: Added

tests/unit/test_operations/test_get_xhtml_errors.py
alexfurmenkov: I've updated the test to make sure that ColumnNotFoundError is raised

cdisc_rules_engine/exceptions/custom_exceptions.py
```diff
@@ -10,6 +10,11 @@ class DatasetNotFoundError(EngineError):
     description = "Dataset not found"
 
 
+class ColumnNotFoundError(EngineError):
```
ASL-rmarshall: I think the creation of this new custom error and the code to raise it is unnecessary.  Although I agree that it is cleaner to check for the referenced column and raise this new custom error explicitly if it's not found, there is already code available that handles errors raised by references to missing columns - the resultant non-custom `KeyError` error is handled by `handle_operation_exceptions` which reports "Column not found in data" - here: https://github.com/cdisc-org/cdisc-rules-engine/blob/c261062297d41776baf13e0f4a9084a4ee36d89d/cdisc_rules_engine/rules_engine.py#L431-L437

If we start using this new custom error here, it might suggest it should be used in other places that already rely on the `KeyError` handling.

cdisc_rules_engine/operations/get_xhtml_errors.py
```diff
@@ -0,0 +1,143 @@
...
+    <!ATTLIST tag
+        name CDATA #REQUIRED
+    >
+    """
+
+    def _execute_operation(self):
+        dataset = self.evaluation_dataset
+        target = self.params.target
+        if target not in dataset:
+            raise ColumnNotFoundError
```
ASL-rmarshall: It's not necessary to use a new custom exception here - `raise KeyError(target)` would give exactly the same results without all the additional code. If fact, you could just remove lines 41 and 42 completely and the operation would still work in the same way, with missing column exceptions being handled when `KeyError` is raised by referencing `dataset[target]` in line 44.

cdisc_rules_engine/operations/base_operation.py
```diff
@@ -73,6 +74,7 @@ def execute(self) -> DatasetInterface:
         except (
             EngineError,
             DatasetNotFoundError,
+            ColumnNotFoundError,
```
ASL-rmarshall: Unnecessary

resources/schema/Operations.json
```diff
@@ -393,6 +393,13 @@
       },
       "required": ["id", "operator"],
       "type": "object"
+    },
+    {
+      "properties": {
+        "operator": { "const": "get_xhtml_errors" }
+      },
+      "required": ["id", "operator", "name", "namespace"],
```
ASL-rmarshall, line 401: The `namespace` property also needs to be defined.  Please insert the following at line 530 (because properties are defined alphabetically):
```json
    "namespace": {
      "type": "string"
    },
```

resources/schema/Operations.md
```diff
@@ -1233,3 +1233,15 @@ Operations:
     name: USUBJID
     id: $aeterm_is_null
 ```
+
+### get_xhtml_errors
+
+Validates XHTML fragments in the target column against the defined XSD schema.
```
ASL-rmarshall: I think it would be better to say "specified namespace" instead of "defined XSD schema".

Also, after the yaml example, please add:
> Note that a local XSD file is required for validation. The file must be stored in the folder indicated by the value of the `LOCAL_XSD_FILE_DIR` default file path and the mapping between the namespace and the local XSD file's sub-folder/name must be included in the value of the `LOCAL_XSD_FILE_MAP` default file path.

PR #1373, opened 2025-10-09
1333: sync requirements.txt and pyproject.toml

alexfurmenkov:


alexfurmenkov:
I have validated that on python3.12 `pip install .` and `pip install -r requirements.txt` works fine

core.py
```diff
@@ -751,3 +751,6 @@ def test_validate():
 if __name__ == "__main__":
     freeze_support()
     cli()
+
+
+
```
alexfurmenkov: Just a test to make sure that linting works after my change. Will roll it back soon

.github/workflows/lint_format.yml
```diff
@@ -34,8 +34,7 @@ jobs:
           python-version: "3.12"
       - name: Install linters
         run: |
-          pip install flake8==6.1.0
-          pip install black==24.10.0
+          pip install -r requirements-dev.txt
```
SFJohnson24: I would do 
```    pip install black flake8 -c requirements-dev.txt```
to install just those two packages and not all

File: requirements-dev.txt
Comment (SFJohnson24) [inline requirements-dev.txt:1]: I would add a line ```-r requirement.txt``` to this file at the top so users can install all requirements for dev and runtime with just ```pip install requirement-dev.txt```

File: .github/workflows/test_unit.yml
Comment (SFJohnson24) [inline .github/workflows/test_unit.yml:1]: this only needs the requirements-dev.txt if we reference requirements.txt in the dev file

Comment (alexfurmenkov) [reply .github/workflows/test_unit.yml:1]: sorry not getting your point... we basically need to install dev dependencies here to make pytest work

Comment (SFJohnson24) [reply .github/workflows/test_unit.yml:1]: this is in reference to what i wrote below:
"I would add a line -r requirement.txt to this file at the top so users can install all requirements for dev and runtime with just pip install requirement-dev.txt"
If installing dev requirements has a reference to requirement.txt, then we only need ```pip install -r requirements-dev.txt```

Comment (alexfurmenkov) [reply .github/workflows/test_unit.yml:1]: gotcha, just updated

File: README.md [hunk 3]
```diff
@@ -103,6 +103,8 @@ NOTE: if you have multiple versions of python on your machine, you can call pyth
 
 `python -m pip install -r requirements.txt` # From the root directory
```
SFJohnson24: eliminate this line-- just need to install python -m pip install -r requirements-dev.txt

alexfurmenkov: removed

RamilCDISC:
Could you please resolve the conflict here?

PR #1375, opened 2025-10-09
715: usdm schema validation

alexfurmenkov:


gerrycampion:
@alexfurmenkov @ASL-rmarshall
For the json files, this is how I handle running the entire json a single time.


First, create the fule json "dataset" with the name "json":
https://github.com/cdisc-org/cdisc-rules-engine/blob/233c00e8870df0245e33779bb14b01fed3004426/cdisc_rules_engine/rules_engine.py#L124-L131


Here, we allow this single full json to run. (I think in these locations, we should change from "json" to a constant so that we can associate these names):
https://github.com/cdisc-org/cdisc-rules-engine/blob/233c00e8870df0245e33779bb14b01fed3004426/cdisc_rules_engine/utilities/rule_processor.py#L639-L643


Here we call the jsonata processor:
https://github.com/cdisc-org/cdisc-rules-engine/blob/233c00e8870df0245e33779bb14b01fed3004426/cdisc_rules_engine/rules_engine.py#L366-L369


This is where I get the jsonata results and map them into error entities and an error container. You will want to do something similar for the schema results:
https://github.com/cdisc-org/cdisc-rules-engine/blob/233c00e8870df0245e33779bb14b01fed3004426/cdisc_rules_engine/utilities/jsonata_processor.py#L42-L89

resources/schema/Rule_Type.md
```diff
@@ -556,3 +556,17 @@ Attach define xml metadata at variable level
...
+
+- `json_path`
+- `error_attribute`
+- `error_value`
+- `validator`
+- `validator_value`
+- `message`
+- `dataset`
+- `id`
+- \_`path`
```
gerrycampion: shouldn't it be `_path` ?

resources/cache/usdm-3-0-schema.pkl
gerrycampion, line 1: What is the source for creating these files? If possible, the logic for creating these should be added to `update-cache` call.

resources/schema/Rule_Type.md
alexfurmenkov: Yes, thanks for noticing the typo. Updated

resources/cache/usdm-3-0-schema.pkl
alexfurmenkov: A combination of https://github.com/cdisc-org/DDF-RA/blob/main/Deliverables/API/USDM_API.json and this script:

```
import os
import argparse
import json

def parse_arguments():
    parser = argparse.ArgumentParser()
    parser.add_argument("-i", "--input_file", help="USDM OpenAPI JSON file")
    args = parser.parse_args()
    return args

args = parse_arguments()

filename = os.path.split(args.input_file)[-1]

outfname = "".join(filename.split(".")[0])+"_schemas"

with open(args.input_file) as f:
    openapi = json.load(f)

jschema = {"$defs": {}}

def replace_deep(data, a, b):
    if isinstance(data, str):
        return data.replace(a, b)
    elif isinstance(data, dict):
        return {k: replace_deep(v, a, b) for k, v in data.items()}
    elif isinstance(data, list):
        return [replace_deep(v, a, b) for v in data]
    else:
        # nothing to do?
        return data
    
for sn,sd in openapi["components"]["schemas"].items():
    if sn == "Wrapper-Input":
        for k, v in sd.items():
            jschema[k] = replace_deep(replace_deep(v,"components/schemas","$defs"),"-Input","")
    elif not sn.endswith("-Output"):
        #jschema["$defs"][sn] = to_json_schema(replace_deep(sd,"components/schemas","$defs"))
        jschema["$defs"][sn.replace("-Input","")] = replace_deep(replace_deep(sd,"components/schemas","$defs"),"-Input","")

for v in jschema["$defs"].values():
    v.update({"additionalProperties": False})
    for pn, pd in v.get("properties", {}).items():
        if pn in v.get("required", []) and pd.get("type","") == "array":
            pd.update({"minItems": 1})

with open(os.path.join(''.join(os.path.split(args.input_file)[0:-1]),outfname+'.json'), "w", encoding="utf-8") as f:
    json.dump(jschema, f, ensure_ascii=False, indent=4)

```

Btw, this command was used to retrieve the 3rd version of the schema:
`git --no-pager --git-dir DDF-RA.git show --format=format:"%B" v3.0.0:Deliverables/API/USDM_API.json > USDM_API_v3-0-0.json`

alexfurmenkov: @gerrycampion I am not sure if it should be placed into the update cache call since the schema definition is being loaded into the `LibraryMetadataContainer` inside of the `get_library_metadata_from_cache` function:

https://github.com/cdisc-org/cdisc-rules-engine/blob/ea6bd8e4d17d79975de0e73a7bac2c4966ad8db5/scripts/run_validation.py#L134

https://github.com/cdisc-org/cdisc-rules-engine/pull/1375/files#diff-645421107f064022b54581bdaf972ee1baa090acef121e83d4b27be3f50ed802R146

ASL-rmarshall: For consistency, you could - and probably should - also get the v4 spec (and any subsequent versions) in the same way as you get the v3 spec:
```
git --no-pager --git-dir DDF-RA.git show --format=format:"%B" v4.0.0:Deliverables/API/USDM_API.json > USDM_API_v4-0-0.json
```
(It's worth noting that `git ... show --format=format:"%B"` is used because this does a binary transfer that avoids any platform-specific encoding).

Though ideally the specs would be in the Library already...

alexfurmenkov: @ASL-rmarshall, Yes, I was using the git command to also retrieve the 4th version

gerrycampion: @alexfurmenkov  If you can't put it in the update-cache, I think you should at least add a github action for it, like "prerelease-update-usdm-schema.yml"

gerrycampion (approved):

resources/cache/usdm-3-0-schema.pkl
alexfurmenkov: Updated readme with USDM Schema update instructions. We've agreed that this is going to be updated rarely, so just manual instructions should be good for now

PR #1377, opened 2025-10-14
Update jsonata_processor.py

gerrycampion:


DmitryMK:
All good, but I'm not merging as I think I should not do it.

Issue #1211, opened 2025-06-18
Value 'NA' is treated as NaN (Data in Excel format)


HajimeShimizu:
Issue: 
When '**NA**' is included in data in excel, the value is treated as NaN in data loading process.

'NA' is one of valid terms (ex. NY codelist) and this text value (="NA") can be found in raw data. However, if 'NA' is included in Excel file, the value is treated as NaN.

Cause:
We have following code in cdisc_rules_engine/services/data_services/excel_data_service.py (def get_dataset)

        dataframe = pd.read_excel(
            self.dataset_path,
            sheet_name=dataset_name,
            dtype=dtypes,
            skiprows=(1, 2, 3),
        )

As default, several values are treated as missing value in Pandas ( 'NA', 'NaN', 'N/A', 'NULL', 'null', etc.). Here we need to specify which value should be recognized as missing value. (I've already tested the code below and, it looks fine.)

        dataframe = pd.read_excel(
            self.dataset_path,
            sheet_name=dataset_name,
            dtype=dtypes,
            skiprows=(1, 2, 3),
            # suggested change
            na_values=[''],
            keep_default_na=False,
        )

best regards

P.S. The same behavior happens in other method (ex. read_csv())

PR #1383, opened 2025-10-15
1211-fix-na-vs-nan-excel

RakeshBobba03:
Fixed issue #1211 where the string value 'NA' was incorrectly converted to NaN when reading Excel files. By adding na_values=[''] and keep_default_na=False parameters to all five pd.read_excel() calls in excel_data_service.py, the string 'NA' is now preserved as a valid data value instead of being treated as missing. This allows 'NA' to be used as a legitimate term while empty strings continue to be properly handled as missing values.


gerrycampion (changes requested):
Add a unit test for this

gerrycampion (approved):

Issue #1379, opened 2025-10-14
Rule blocked: FB4607, FB4608 - New Rule Sensitivity


SFJohnson24:
Currently we only are able to report at a dataset or a record level for a rule.  It would be useful to report on a 'GROUP' level that is between dataset and record.  FB4607 either reports only one issue or for all the records in each set.  It would be useful to be able to set the sensitivity so we report once for each set for these rules.

The new sensitivity should allow for the author to pass what they are reporting on-- this would be done in actions.py and the error objects could be parsed by the group that is passed in the rule to select issues for each group to report on.

JozefAerts:
Great! I will try to set up a list of (SEND) rules where this will be useful.

SFJohnson24:
this is the same issue as #831

SFJohnson24:
SEND rules 266 and 266.1, probably also SDTM rule CG0013, when "Sensitivity" is set to "Dataset"
The current implementation only reports the first violation for the dataset.
For example, in our negative SEND Excel dataset for VS, we have the variables VSNOTEST, VSLOBXFL and VSREPNUM.
The first one is clearly invalid, but also VSLOBXFL and VSRESPNUM are not allowed in SEND (they are in SDTM).

When executed, the system only reports VSNOTEST, no others.
When VSNOTEST is corrected (in VSTESTCD), the system only reports VSLOBXFL.
When the latter is removed, VSRESPNUM is reported.

However, when Sensitivity is set to "Record", then a violation is reported for EACH record in the set/group for which there is a violation. So for example, if there are 20 records in the set/group that violates the rule, 20 violation messages are created.

Maybe we need an additional "Sensitivity" named "Group"???

This is non-blocking, but it would be nice when there is a way to take care that for each group that violates a rule, exactly one record in the output is generated.

SFJohnson24:
I got the same behaviour with SEND234 and SEND235 where groups with the same value for SETCD are checked for having a row with TXPARMCD=SPGRPCD or ARMCD respectively.
When there are two groups that violate the rule, only the first violation is reported.

PR #1388, opened 2025-10-20
1379 fb4607,fb4608 new rule sensitivity

RakeshBobba03:
Add GROUP sensitivity level for rules to report once per logical group

Implements issue #1379 by adding a new "Group" sensitivity option that reports validation errors once per group of records rather than once per dataset or once per record. Rule authors can now specify Grouping_Variables in their rule definitions to control how errors are aggregated. This solves the problem where rules like FB4607 and FB4608 were either reporting only the first violation with Dataset sensitivity or generating excessive duplicate errors (one per record) with Record sensitivity. The implementation includes updates to the sensitivity enum, actions.py grouping logic, rule model and serialization, schema definitions, and comprehensive tests.

### DIFF
--- a/core.py
+++ b/core.py
@@ -276,7 +276,7 @@ def validate(
     validate_xml: str,
     jsonata_custom_functions: tuple[()] | tuple[tuple[str, str], ...],
     max_report_rows: int,
-    max_errors_per_rule: tuple[int | bool],
+    max_errors_per_rule: tuple[int, bool],
 ):
     """
     Validate data using CDISC Rules Engine


cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -879,6 +879,62 @@ def shorter_than(self, other_value: dict):
     def shorter_than_or_equal_to(self, other_value: dict):
         return ~self.longer_than(other_value)
 
+    @log_operator_execution
```
SFJohnson24, line 882: what is this operator and how does it relate to this ticket?

cdisc_rules_engine/models/actions.py
```diff
@@ -166,6 +166,41 @@ def generate_targeted_error_object(
             errors_list = self._generate_errors_by_target_presence(
                 data, targets_not_in_dataset, all_targets_missing, errors_df
             )
+        elif self.rule.get("sensitivity") == Sensitivity.GROUP.value:
+            # Handle GROUP sensitivity
```
SFJohnson24, line 170: please remove the comments from this code block -- I dont think they add any context that isn't easily gleaned from reading the code

cdisc_rules_engine/models/actions.py
```diff
@@ -166,6 +166,41 @@ def generate_targeted_error_object(
...
+            grouping_variables = self.rule.get("grouping_variables", [])
+
+            if not grouping_variables:
+                # Error: Group sensitivity requires Grouping_Variables
+                error_entity = ValidationErrorEntity(
+                    dataset=self._get_dataset_name(data),
```
SFJohnson24, line 176: I believe this should and USUBJID should be "N/A" similar to when sensitivity is not correctly defined below

cdisc_rules_engine/models/actions.py
```diff
@@ -267,6 +302,145 @@ def _generate_errors_by_target_presence(
...
+        """
+        missing_vars = {target: "Not in dataset" for target in targets_not_in_dataset}
+        errors_list = []
+
+        # Validate grouping variables exist
+        available_grouping_vars = [
+            var for var in grouping_variables if var in data.columns
+        ]
+
+        if not available_grouping_vars:
```
SFJohnson24, line 334: we should be throwing an error here if one or all of the grouping they specify doesnt exist--not falling back to a different sensitivity.  We also shouldnt be filtering the grouping variables for only the ones that exist.

resources/schema/CORE-base.json
```diff
@@ -608,6 +608,13 @@
     },
     "Sensitivity": {
       "$ref": "Sensitivity.json"
+    },
+    "Grouping_Variables": {
+      "items": {
+        "$ref": "#/$defs/VariableReference"
+      },
+      "type": "array",
+      "markdownDescription": "Variables to group by when using Group sensitivity. Required when Sensitivity is set to Group."
```
SFJohnson24, line 617: can we add a conditional block after properties, that if sensitivity is Group, Grouping_variables is required.
That way, our schema will flag when Group sensitivity is chosen in rule editor but grouping_variables is not given

cdisc_rules_engine/models/actions.py
```diff
@@ -267,6 +302,145 @@ def _generate_errors_by_target_presence(
...
+            ]
+
+        # Group by the specified variables
+        if all_targets_missing:
+            grouped_data = data.groupby(available_grouping_vars, dropna=False)
+        else:
+            # Merge data with errors_df to preserve grouping info
+            # Use inner join to only get rows that have errors
+            merged_data = data.loc[errors_df.index].copy()
+            for col in errors_df.columns:
```
SFJohnson24, line 369: what does this loop on 369-371 do?

error_df is a subset of data, so why are you checking when a column in it isnt in data

cdisc_rules_engine/models/actions.py
```diff
@@ -267,6 +302,145 @@ def _generate_errors_by_target_presence(
...
+                error_value = {**error_value, **missing_vars}
+
+            return [
+                ValidationErrorEntity(
+                    value=error_value,
+                    dataset=self._get_dataset_name(data),
+                )
+            ]
+
+        # Group by the specified variables
```
SFJohnson24, line 362: please remove unnecessary comments

SFJohnson24 (changes requested):
see comments for some changes.  
_generate_errors_by_group() has a ton of branching that is checking the same code repeatedly.  Null filtering is done 3 times in this code.  Can you parse some of this redundant code into helpers
Also note-- sensitivity.md needs to updated as this is used in our reference guide.

Issue #1372, opened 2025-10-08
Proposal for datetime comparison where precision of the datetimes is different.


JozefAerts:
Unfortunately, sometimes, we need to compare datetimes for wich the granularity/precision is different. For example, compare: 
datetime1 = 2025-06-25
datetime2 = 2025-06-25T17:22

Therefore I propose a function that allows to compare the "common" part of the two datetimes.
For the above, this means datetime2 is reduced to "2025-06-27" which has the same character length as datetime1. The result of the comparison would then be "equal".
Whereas when datetime2 would be "2025-06-24T16:40:33. the result would be that datetime2 comes before datetime1.

I am pretty bad in Python, so I asked ChatGPT to write such a function for me. The code is below.
So  I propose to develop some new operators based on this function, similar to "date_equal_to", "date_greater_than", etc.. 
I let people with a better knowledge of the English language to come up with appropriate names ...

Python: code (many thanks to ChatGPT ...):

from datetime import datetime

def compare_iso_datetimes(dt1: str, dt2: str) -> int:
    """
    Compare two ISO-8601 datetime strings by truncating both to the length
    of the shorter string before parsing.

    Returns:
        -1 if dt1 < dt2
         0 if dt1 == dt2
         1 if dt1 > dt2
    """
    # Truncate both to the length of the shortest string
    min_len = min(len(dt1), len(dt2))
    dt1_trunc = dt1[:min_len]
    dt2_trunc = dt2[:min_len]

    # Attempt to parse truncated strings
    # Fallback: if parsing fails, compare lexicographically
    try:
        d1 = datetime.fromisoformat(dt1_trunc)
        d2 = datetime.fromisoformat(dt2_trunc)
    except ValueError:
        # Fall back to lexicographic comparison if parsing fails
        if dt1_trunc < dt2_trunc:
            return -1
        elif dt1_trunc > dt2_trunc:
            return 1
        else:
            return 0

    # Compare datetimes
    if d1 < d2:
        return -1
    elif d1 > d2:
        return 1
    else:
        return 0

SFJohnson24:
we would likely want this behavior as native for the vectorized date comparison that underlies the date check operators

JozefAerts:
I have no objections ... ;-)

PR #1390, opened 2025-10-21
1372 datetime comparison with different precision levels

RakeshBobba03:
Add automatic precision detection for date comparison operators (Issue #1372 )
Extends the existing date_component parameter to accept "auto" as a value, enabling automatic precision detection when comparing dates with different granularity levels. When specified, the system detects the precision of both dates and compares them at the common (less precise) level. This solves the use case of comparing date-only fields like RFSTDTC ("2025-06-25") with datetime fields like AESTDTC ("2025-06-25T17:22") at the appropriate precision level. Implementation includes support for CDISC partial dates and uncertainty markers, comprehensive test coverage, and updated documentation in README.md and Operator.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,53 @@ The possible rule run statuses are:
 - `SUCCESS` - The rule ran and data was validated against the rule. May or may not produce results
 - `SKIPPED` - The rule was unable to be run. Usually due to missing required data, but could also be cause by rule execution errors.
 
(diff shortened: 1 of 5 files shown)


SFJohnson24 (changes requested):
Hey Rakesh,
please resolve the base branch so this PR is just from this issue.  Happy to help you untangle things if you need help with git.

PR #1399, opened 2025-10-24
cg0546

SFJohnson24:
current cg0546 fails if you try to use test data (the data in cg0545 works for this rule) because the argument is order and not sort_order (see reference guide and ).  This fixes the issue and allows the operator to execute


mhungria:
Thanks, @SFJohnson24 !!

PR #1404, opened 2025-10-27
cg0370

SFJohnson24:
This pull request introduces support for using reference values in the `Distinct` operation, allowing the operation to treat the value in the target column as a reference to another column in the same row. This is controlled via a new `value_is_reference` boolean parameter, which is now supported throughout the operation pipeline. The changes also include updates to the schema, new tests for this functionality, and a minor improvement to domain matching logic.
[Datasets.json](https://github.com/user-attachments/files/23172683/Datasets.json)
[Rule_underscores.json](https://github.com/user-attachments/files/23172684/Rule_underscores.json)
this is one of CG0370 sub-rules using this logic and negative data for it.

Key changes:

**Distinct Operation Reference Value Support:**
* Added a new `value_is_reference` boolean parameter to `OperationParams`, and updated the `Distinct` operation logic to use the value in the target column as a reference to another column when this flag is set. This includes support for both grouped and ungrouped distinct operations. [[1]](diffhunk://#diff-a8c3ea5072f0809ba706bc8b5ca1aa810d71a56b0bbb88bb49fc86cf6db89bb2R60) [[2]](diffhunk://#diff-ce2092de5cc87f43f1f577740e3f54c800b01b134fd43de0511db192669a974fR5-R51)
* Updated the JSON schema (`Operations.json`) to include the new `value_is_reference` parameter.
* Updated the rule processor to pass the `value_is_reference` parameter when constructing operation parameters.

**Testing:**
* Added unit tests to verify the new reference value behavior in both grouped and ungrouped contexts for the `Distinct` operation, covering both Pandas and Dask datasets.

**Domain Matching Logic:**
* Improved domain lookup logic in the rule processor to support matching domains with a trailing double dash (`--`) to all domains with a common prefix.

### DIFF
--- a/tests/unit/test_operations/test_distinct.py
+++ b/tests/unit/test_operations/test_distinct.py
@@ -229,7 +229,7 @@ def test_distinct_value_is_reference(
 
     data_service.data = [MockDataset("LB", "lb.xpt")]
 
-    def mock_get_dataset(filename, **kwargs):
+    def mock_get_dataset(dataset_name, **kwargs):
         return referenced_data
 
     data_service.get_dataset = mock_get_dataset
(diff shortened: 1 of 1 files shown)


SFJohnson24:
updated docs @RamilCDISC

PR #1405, opened 2025-10-28
1379 fb4607,fb4608 rule sensitivity

RakeshBobba03:
Add GROUP sensitivity level for rules to report once per logical group

Implements issue #1379 by adding a new "Group" sensitivity option that reports validation errors once per group of records rather than once per dataset or once per record. Rule authors can now specify Grouping_Variables in their rule definitions to control how errors are aggregated. This solves the problem where rules like FB4607 and FB4608 were either reporting only the first violation with Dataset sensitivity or generating excessive duplicate errors (one per record) with Record sensitivity. The implementation includes updates to the sensitivity enum, actions.py grouping logic, rule model and serialization, schema definitions, and comprehensive tests.


resources/schema/Sensitivity.md
```diff
@@ -5,3 +5,7 @@ Report one result per dataset generated by the `Rule Type`, where a dataset or r
 ## Record
 
 Report one result per record within the dataset generated by the `Rule Type`, where the record matches the rule failure criteria
+
+## Group
+
+Report one result per logical group of records within the dataset generated by the `Rule Type`, where the group matches the rule failure criteria. Groups are defined by the `Grouping_Variables` specified in the rule. This sensitivity level requires `Grouping_Variables` to be specified and provides a middle ground between Dataset (too few reports) and Record (too many reports) sensitivity levels.
```
SFJohnson24: can we include a yaml example of how this sensitivity with the grouping variables is entered in a rule
we should also explain it is grabbing the first record for each of the distinct groups so they understand the report behavior

RakeshBobba03: Added YAML examples to the Group sensitivity documentation showing both the basic syntax and rule example.

resources/schema/Sensitivity.md
```diff
@@ -5,3 +5,52 @@ Report one result per dataset generated by the `Rule Type`, where a dataset or r
...
+
+### Example
+
+```yaml
+Sensitivity: Group
+Grouping_Variables:
+  - SETCD
+```
+
+### Complete Rule Example
```
SFJohnson24: a full rule example isnt needed, just the grouping variable (I also think in yaml it is a space and not an underscore for Grouping_Variables)

resources/schema/Sensitivity.md
```diff
@@ -5,3 +5,14 @@ Report one result per dataset generated by the `Rule Type`, where a dataset or r
...
+
+## Group
+
+Report one result per logical group of records within the dataset generated by the `Rule Type`, where the group matches the rule failure criteria. Groups are defined by the `Grouping_Variables` specified in the rule. This sensitivity level requires `Grouping_Variables` to be specified and provides a middle ground between Dataset (too few reports) and Record (too many reports) sensitivity levels.
+
+### Example
+
+```yaml
+Sensitivity: Group
+Grouping_Variables: SETCD
```
SFJohnson24: Grouping_Variables should be Grouping Variables for yaml

resources/schema/Sensitivity.md
SFJohnson24: also need: 'we should also explain it is grabbing the first record for each of the distinct groups so they understand the report behavior'

PR #1406, opened 2025-10-28
1372 datetime comparison

RakeshBobba03:
Add automatic precision detection for date comparison operators (Issue https://github.com/cdisc-org/cdisc-rules-engine/issues/1372 )
Extends the existing date_component parameter to accept "auto" as a value, enabling automatic precision detection when comparing dates with different granularity levels. When specified, the system detects the precision of both dates and compares them at the common (less precise) level. This solves the use case of comparing date-only fields like RFSTDTC ("2025-06-25") with datetime fields like AESTDTC ("2025-06-25T17:22") at the appropriate precision level. Implementation includes support for CDISC partial dates and uncertainty markers, comprehensive test coverage, and updated documentation in README.md and Operator.md.


cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -26,8 +27,6 @@ def is_valid_date(date_string: str) -> bool:
     except Exception as e:
         uncertainty_substrings = ["/", "--", "-:"]
         if any([substr in date_string for substr in uncertainty_substrings]):
-            # date_string contains uncertainty
```
RamilCDISC, line 29: Could you please mention the reason for removing these comments?

RakeshBobba03: I initially removed these comments because I treated them as redundant explanatory comments. I am now realising I was incorrect. Restoring the comments

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -116,6 +117,85 @@ def get_microsecond(date_string: str):
...
+    "day": 2,
+    "hour": 3,
+    "minute": 4,
+    "second": 5,
+    "microsecond": 6,
+}
+
+PRECISION_LEVELS = [
+    "year",
+    "month",
```
RamilCDISC: Could you please move these to top of file along with date_regex pattern for consistency?

tests/unit/test_check_operators/test_date_comparison_checks.py
```diff
@@ -8,43 +8,39 @@
     "data,dataset_type,expected_result",
     [
         (
-            {"target": ["2021", "2099", "2022", "2023"]},
+            ["2021", "2099", "2022", "2023"],
```
RamilCDISC: Could you please explain why updating the test if no new test cases are being added?

tests/unit/test_check_operators/test_date_comparison_checks.py
```diff
@@ -747,3 +743,129 @@ def test_is_incomplete_date(target, dataset_type, expected_result):
...
     )
+
+
+@pytest.mark.parametrize(
+    "operator_name,data,comparator,expected_result",
+    [
+        (
+            "date_equal_to",
+            ["2025-06-25", "2025-06-24", "2025-06-25", "2025-06-26"],
+            "2025-06-25T17:22",
```
RamilCDISC: Could you please add here the test case where data list also has a data with 'T'? You can add the same date as comparator to test. Same for other test cases below.

tests/unit/test_check_operators/test_date_comparison_checks.py
RakeshBobba03: I was told in a previous ticket: "In tests instead of defining a dictionary here like {"target": [1,2,3]}, you could just define the inner array like [1,2,3] and build the target below like: df = dataset_type.from_dict({"target": data}) It just simplifies the test cases a bit. Same for other locations wherever this applies"
I misunderstood that and thought I could modify similar pre-existing tests. I changed test_invalid_date because it was the only other single-column test that could use the same simplification as test_auto_precision_operators. That was inconsistent and selective.
I should have only touched the new test_auto_precision_operators test and left all pre-existing tests unchanged. I'll revert test_invalid_date to maintain consistency with the codebase.

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -17,6 +18,26 @@
...
+    "day",
+    "hour",
+    "minute",
+    "second",
+    "microsecond",
+]
+
```
gerrycampion: Can these two variables be replaced by a single enum?

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -116,6 +137,64 @@ def get_microsecond(date_string: str):
...
+        return "day"
+    elif len(date_parts) == 2:
+        return "month"
+    elif len(date_parts) == 1:
+        return "year"
+    return None
+
+
+@lru_cache(maxsize=1000)
+def detect_date_precision(date_str: str) -> str:
```
gerrycampion: this should be called `detect_datetime_precision`

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -116,6 +137,64 @@ def get_microsecond(date_string: str):
     return timestamp.microsecond
 
 
+def _detect_time_precision(time_part: str) -> str:
```
gerrycampion: Rather than parsing the datetimes yourself, I would try something like these steps:
1. Use the date_regex to break the datetime into its components
2. Iterate through the components to find the first missing or "-" component

I think this will simplify things.

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -187,13 +266,15 @@ def case_insensitive_is_in(value, values):
...
-        # Comparison should return false if either is empty or None
         return False
-    else:
-        return operator(
-            get_date_component(component, target),
-            get_date_component(component, comparator),
-        )
+
+    if component == "auto":
+        component = get_common_precision(target, comparator)
```
gerrycampion: Operators that involve equals like eq, ne, geq, leq should require an explicit `auto` like you have here, but the strictly less than or strictly greater than operators should implicitly use the common precision by default.

tests/unit/test_check_operators/test_date_comparison_checks.py
```diff
@@ -747,3 +747,189 @@ def test_is_incomplete_date(target, dataset_type, expected_result):
...
+            [True, False, True, False],
+        ),
+        (
+            "date_equal_to",
+            ["2025-06-25", "2025-06-24", "2025-06-25T17:22", "2025-06-26"],
```
gerrycampion: There are a lot of repetitive test cases, like "2025-06-25". Each test scenarios only needs to be present once.

tests/unit/test_check_operators/test_date_comparison_checks.py
```diff
@@ -747,3 +747,189 @@ def test_is_incomplete_date(target, dataset_type, expected_result):
...
+
+@pytest.mark.parametrize(
+    "operator_name,data,comparator,expected_result",
+    [
+        (
+            "date_equal_to",
+            ["2025-06-25", "2025-06-24", "2025-06-25", "2025-06-26"],
+            "2025-06-25T17:22",
+            [True, False, True, False],
+        ),
```
gerrycampion: Instead of test cases like this:
```python
[
        (
            "date_equal_to",
            ["2025-06-25", "2025-06-24", "2025-06-25", "2025-06-26"],
            "2025-06-25T17:22",
            [True, False, True, False],
        ),
...
]
```
It would be better to organize test cases like this:
```python
[
        (
            "date_equal_to",
            "2025-06-25",
            "2025-06-25T17:22",
            "auto"
            True,
        ),
        (
            "date_equal_to",
            "2025-06-24",
            "2025-06-25T17:22",
            "auto"
            False,
        ),
        (
            "date_equal_to",
            "2025-06-25",
            "2025-06-25T17:22",
            "auto"
            True,
        ),
        (
            "date_equal_to",
            "2025-06-26",
            "2025-06-25T17:22",
            "auto"
            False,
        ),
...
]
```
Easier to read and each case is a separate unit that can pass/fail. Also, I'm not sure why they are in groups of 3 or 4. I added the "auto" param so that you can have "auto" or "None" so that you can test both for the strictly greater or less than scenarios.

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -116,6 +137,64 @@ def get_microsecond(date_string: str):
...
+        return "year"
+    return None
+
+
+@lru_cache(maxsize=1000)
+def detect_date_precision(date_str: str) -> str:
+    if not date_str or not isinstance(date_str, str):
+        return None
+
+    if "T" not in date_str and "--" not in date_str and "-:" not in date_str:
```
gerrycampion: You are missing unit tests for datetimes with missing components like the ones handled in this case. The SDTMIG has some examples:

Image

tests/unit/test_check_operators/test_date_comparison_checks.py
```diff
@@ -747,3 +747,189 @@ def test_is_incomplete_date(target, dataset_type, expected_result):
...
+        (
+            "date_equal_to",
+            ["2025-06-25T17:22", "2025-06-25T17:21", "2025-06-25T17:22:30"],
+            "2025-06-25T17:22:30",
+            [True, False, True],
+        ),
+        (
+            "date_equal_to",
+            ["2025-06-25T", "2025-06-24T", "2025-06-25T"],
```
gerrycampion: These are invalid ISO dates. I would expect these to give errors or False (I don't know which is the default behavior). But I wouldn't expect to see True's.
It is good to have tests with invalid tests.

README.md
```diff
@@ -266,6 +266,19 @@ The possible rule run statuses are:
...
+
+```yaml
+- name: "AESTDTC"
+  operator: "date_greater_than_or_equal_to"
+  value: "RFSTDTC"
+  date_component: "auto"
+```
+
+This compares `"2025-06-25"` with `"2025-06-25T17:22"` at day precision, `"2025-06"` with `"2025-06-25"` at month precision, etc.
+
```
gerrycampion: You don't need this section in the general readme. It should only be in Operator.md

tests/unit/test_check_operators/test_date_comparison_checks.py
```diff
@@ -747,3 +782,95 @@ def test_is_incomplete_date(target, dataset_type, expected_result):
...
+        ("2025-06", "2025-06-25", "auto", False),
+        ("2025-07", "2025-06-25", "auto", True),
+    ],
+}
+
+AUTO_PRECISION_PARAMS = [
+    (operator_name, target, comparator, date_component, expected_result)
+    for operator_name, scenarios in AUTO_PRECISION_CASES.items()
+    for target, comparator, date_component, expected_result in scenarios
+]
```
gerrycampion, line 877: great job parametrizing these tests. This is very easy to read

tests/unit/test_check_operators/test_date_comparison_checks.py
```diff
@@ -50,6 +71,19 @@ def test_invalid_date(data, dataset_type, expected_result):
     assert result.equals(df.convert_to_series(expected_result))
 
 
+@pytest.mark.parametrize(
+    "value,expected_precision",
+    [
+        ("2003---15", "day"),
+        ("--12-15", "day"),
+        ("-----T07:15", "minute"),
```
gerrycampion: These are wrong. The expected precision for these should be "year", None, None. You want the precision before the first missing component. And can you add some more of these test cases as well? Like `"2003-12-15T-:15", "day"` and `"2003-12-15T13:-:17", "hour"` from the example. It would also be good to include some of these in the cases for `test_auto_precision_operators`

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -185,15 +463,116 @@ def case_insensitive_is_in(value, values):
...
+        return dt.replace(day=1, hour=0, minute=0, second=0, microsecond=0)
+    if precision == "day":
+        return dt.replace(hour=0, minute=0, second=0, microsecond=0)
+    if precision == "hour":
+        return dt.replace(minute=0, second=0, microsecond=0)
+    if precision == "minute":
+        return dt.replace(second=0, microsecond=0)
+    if precision == "second":
+        return dt.replace(microsecond=0)
+    return dt
```
gerrycampion: Instead of if's, use a match statement here. The `precision` parameter's type should be passed as a DatePrecision member instead of str.

You could also use a dict of lambdas like you did for `_precision_check_functions`, but I think a match statement is slightly less verbose.

cdisc_rules_engine/check_operators/helpers.py
gerrycampion, line 1: in all instances where you are using strings for the date components ("year", "minute", etc), you should be using the DatePrecision enum instead

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -116,6 +138,262 @@ def get_microsecond(date_string: str):
...
+        elif second_part not in ("", "-"):
+            components["second"] = second_part
+
+
+def _extract_datetime_components(date_str: str) -> dict:
+    """Extract datetime components using regex pattern matching."""
+    components = _empty_datetime_components()
+    if not date_str or not isinstance(date_str, str):
+        return components
+    if not date_regex.match(date_str):
```
gerrycampion: you are using the regex to check for matches, but i don't see you using it to extract the components. See [regex groups](https://docs.python.org/3/howto/regex.html#grouping). I suspect a lot of this parsing code could be simplified using regex groups. https://regex101.com/ is a great tool for regexes and groups.

gerrycampion (changes requested):

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -116,31 +143,411 @@ def get_microsecond(date_string: str):
...
+    if not hour or hour == "-":
+        return None
+    if not minute or minute == "-":
+        return DatePrecision.hour
+    if second is None:
+        return DatePrecision.minute
+    second_part = second
+    if microsecond:
+        second_part = f"{second}.{microsecond}"
+    return _precision_from_second_component(second_part)
```
gerrycampion: it looks like a bunch of methods are no longer being used. can you remove these before i take a look at the rest?

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -56,91 +93,210 @@ def is_valid_duration(duration: str, negative) -> bool:
...
+
+def _date_and_time_precision(
+    date_components, time_components, has_time
+) -> DatePrecision | None:
+    found_missing, date_result = _check_date_precision(date_components)
+    if found_missing:
+        return date_result
+    if not has_time or not time_components:
+        return DatePrecision.day
+    return _check_time_precision(time_components)
```
gerrycampion: Can you explain why these need to be separate? If they don't need to be separate, can they be combined to a single method? Remember to take advantage of iterating through enums like I've done elsewhere instead of handwriting each case.

RakeshBobba03: I've combined the three functions into a single `_date_and_time_precision` method that uses enum iteration throughout, following the same pattern used in the codebase. The function now iterates through `list(DatePrecision)[DatePrecision.year : DatePrecision.day + 1]` for date components and` list(DatePrecision)[DatePrecision.hour : DatePrecision.microsecond + 1]` for time components, taking out the hardcoded lists and redundant helper functions.

cdisc_rules_engine/check_operators/helpers.py
```diff
@@ -56,91 +93,195 @@ def is_valid_duration(duration: str, negative) -> bool:
...
+def _get_precision_before(precision: DatePrecision) -> DatePrecision | None:
+    if precision == DatePrecision.year:
+        return None
+    prev_index = precision.value - 1
+    if prev_index >= 0:
+        return DatePrecision(prev_index)
+    return None
+
+
+def _date_and_time_precision(
```
gerrycampion, line 183: I still don't understand why the dates and times are being handled separately here. Can you explain this? I think I'm missing something.

RakeshBobba03: I separated date and time handling because, within `_date_and_time_precision`, time components are optional while date components are always present.

Time-only cases like "----T13:15" are handled earlier in `detect_datetime_precision`. If all date components are missing, it returns `None` before calling `_date_and_time_precision`. So within `_date_and_time_precision`, we can assume at least one date component exists.

The function first checks date components (year, month, day). If any are missing, it returns the precision before that component.

Then it checks whether a time component exists. If there's no time (e.g., "2003-12-15"), it returns `DatePrecision.day` since that's the last valid date component.

Only if time components exist do we check them (hour, minute, second, microsecond) for missing parts.

I kept them separate because:
1. Within this function's context, dates are always present (the caller ensures at least one date component exists; time-only cases return `None` earlier).
2. Times are optional.
3. When there's no time, the precision should be `DatePrecision.day`, not a time precision.

I can refactor it if you have a different approach in mind.

gerrycampion: Can you give an example test case that would require the date and time precision detection to be separate? Because all of the current test cases seem like they would work fine if the logic was combined/generalized.

RakeshBobba03: Right now there isn't a test case that requires the separation. I separated them for clarity.
If we combined the logic into a single loop through all precisions (year through microsecond), we'd need to check has_time when we reach time precisions. The current approach handles the no-time case upfront, which I feel is clearer. The efficiency difference is minimal (one less iteration for date-only strings). The separation is a code organization choice, not a correctness requirement.

gerrycampion: I'm wondering why we need `has_time` at all? Why is "no-time" a special case here but "no-day" or "no-minutes" are not a special cases?

gerrycampion: sorry for being nitpicky. i committed changes to show what i'm suggesting. let me know your thoughts.

RakeshBobba03: I see it now. Missing time components are handled the same as missing date components: if a component is None or "-", _check_date_component_missing returns True, and we return the precision before it. There's no need for a special has_time case. The unified loop through all DatePrecision values is cleaner and more consistent. Thanks for the commit showing the approach. I was totally looking at it the wrong way before.

PR #1411, opened 2025-10-29
CG0001

SFJohnson24:
this PR updates the is_contained_by
to test CG0001:
negative1 should have 2 issues but shows 3.  CM dataset has C49563 in both define_variable_codelist_coded_codes and domain_lib_ccode so it should not flag this as an issues as it is_not_contained_by.  The issue stems from it not parsing the values from the target.  it is looking for the list ['C49563 '] instead of looking for the value within the other list.

This pull request updates the logic for checking containment within the `is_contained_by` method in `dataframe_operators.py`. The main improvement is more robust handling of columns containing iterable values, ensuring that containment checks work correctly for lists and other iterables in DataFrame columns.

Containment logic improvements:

* Enhanced the `is_contained_by` method to iterate over columns of iterables, using the new `is_in` function to check if any item in a list is contained in the comparator value, ensuring correct handling of columns with lists.
* Updated the import section to include the new `is_in` function, which is now used for containment checks.


RamilCDISC (approved):
The PR updates the is_in operator to handle an edge case. The updated PR was validated by:

2. Reviewing the updated logic in accordance with the AC.
4. Running manual validation against positive datasets.
5. Running manual validations against negative datasets.

PR #1415, opened 2025-10-30
list-rule-set update

SFJohnson24:
image
this PR updates the cache (updates several rules that were cached as tig/1-0 without a substandard).  It updates the schema for tig--see above screen shot which tells users if they put TIG, a substandard is needed
It also updates how the list-set-rules is done to list the substandard as part of the tuple set key.


cdisc_rules_engine/services/cache/cache_populator_service.py
```diff
@@ -168,7 +168,13 @@ def process_rules_lists(self, rules_directory, rules_by_core_id, rules_lists):
                                 rules_directory[sub_key] = []
                             if core_id not in rules_directory[sub_key]:
                                 rules_directory[sub_key].append(core_id)
-
+                        else:
+                            key = f"{std_name}/{std_version}"
```
RamilCDISC: Could you please explain why are we repeating this here? The block 

```
key = f"{std_name}/{std_version}"
 if (
     key in rules_directory
     and core_id not in rules_directory[key]
):
      rules_directory[key].append(core_id)
```
is same as on line 159 - 164

SFJohnson24: that is artifact--removed it

PR #1418, opened 2025-11-04
720: Ignore not existing domains in the operations

alexfurmenkov:


RamilCDISC:
I was testing this branch using SEND105. I changed the rule to this 

> Check:
> 
>   any:
> 
>     - all:
>         - name: TX
>           operator: non_empty
>         - name: $distinct_txparmcd
>           operator: does_not_contain
>           value: SPECIES
>           value_is_literal: true
> Core:
>   Id: CDISC.SENDIG.105
>   Status: Draft
>   Version: '1'
> Description: Raise an error when the variable SPECIES is not present in DM, and
>   there is no record TSPARMCD=SPECIES in TS and there is no record
>   TXPARMCD=SPECIES in TX
> Executability: Fully Executable
> Operations:
>   - domain: TS
>     id: $distinct_tsparmcd
>     name: TSPARMCD
>     operator: distinct
>   - domain: TX
>     id: $distinct_txparmcd
>     name: TXPARMCD
>     operator: distinct

The editor put out this result:

> {
>   "DM": [
>     {
>       "executionStatus": "execution_error",
>       "dataset": "dm.xpt",
>       "domain": "DM",
>       "variables": [],
>       "message": "rule execution error",
>       "errors": [
>         {
>           "dataset": "dm.xpt",
>           "error": "Column not found in data",
>           "message": "TX"
>         }
>       ]
>     }
>   ],
>   "TS": [
>     {
>       "executionStatus": "execution_error",
>       "dataset": "ts.xpt",
>       "domain": "TS",
>       "variables": [],
>       "message": "rule execution error",
>       "errors": [
>         {
>           "dataset": "ts.xpt",
>           "error": "Column not found in data",
>           "message": "TX"
>         }
>       ]
>     }
>   ]
> }

I used the excel datset from the zip in connected issue. @gerrycampion is this output what you would intend the engine to give?

gerrycampion:
> TX

@RamilCDISC , Yes I would expect this result. You are trying to do a column operator, `non_empty` on a dataset `TX` instead of a column. Because the `TX` column does not exist, you get this error.

cdisc_rules_engine/utilities/rule_processor.py
```diff
@@ -480,10 +479,11 @@ def _execute_operation(
                 ),
             )
             if domain_details is None:
-                raise DomainNotFoundError(
-                    f"Operation {operation_params.operation_name} requires Domain "
-                    f"{operation_params.domain} but Domain not found in dataset"
+                logger.info(
+                    f"Domain {operation_params.domain} doesn't exist in the dataset"
```
gerrycampion, line 482: " in the dataset" doesn't make sense to me

tests/QARegressionTests/test_Issues/test_CoreIssue720.py
gerrycampion, line 1: I would prefer if we used the json output format and compared those results instead of the excel results. Using excel is a bit harder to visualize and seems like it could break easily with minor updates to the excel template.

gerrycampion (changes requested):

Issue #1402, opened 2025-10-24
Request to change 'within' to list instead of single string in target_is_sorted_by operator (Rule CG0546)


mhungria:
Hi Team,

This is a request to change 'within' to list instead of single string in target_is_sorted_by operator.

Trying to address Rule CG0546 at the moment. I envision this functionality would apply to other cases.

Thanks,
Marcelina

gerrycampion:
We may want to go one of two routes:
- Ensure we go through the existing rules in the rule editor that contain this operator and parameter to make sure we aren't breaking existing functionality.
- Overload the parameter to accept both single values or a list of values instead of just replacing the param type.

mhungria:
Hi @SFJohnson24 / Team,

This ticket appears as closed; however, I just tested and I got execution error on the rule editor, zip file with all files enclosed here:

[CG0546_20251117.zip](https://github.com/user-attachments/files/23588225/CG0546_20251117.zip)

SFJohnson24:
@RakeshBobba03 can you confirm the above?  
<img width="1667" height="863" alt="Image" src="https://github.com/user-attachments/assets/3bf5cfb4-f55f-449f-9c6e-f60dfc1ca5ae" />  It appears there is an error and SM does not execute when run against this rule.

PR #1429, opened 2025-11-11
1402 CG0546 support multi-key param update

RakeshBobba03:
#1402: Updated `target_is_sorted_by` to normalize the `within` parameter to an ordered list, then reuse the existing sort-and-group flow so it handles one or many grouping keys without materializing the entire dataset. Adjusted the post-group alignment to drop levels only when the grouped result is multi-indexed, keeping Pandas and Dask behavior in sync. Extended the operator schema and reference guide to document the list form, and added positive/negative multi-key tests for both PandasDataset and DaskDataset to cover the regression that affected rule CG0546.

### DIFF
--- a/cdisc_rules_engine/utilities/rule_processor.py
+++ b/cdisc_rules_engine/utilities/rule_processor.py
@@ -390,7 +390,7 @@ class RuleProcessor:
                 standard_version=standard_version,
                 standard_substandard=standard_substandard,
                 external_dictionaries=external_dictionaries,
-                ct_version=operation.get("ct_version"),
+                ct_version=operation.get("version"),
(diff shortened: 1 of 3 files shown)


gerrycampion (approved):

RamilCDISC (approved):
The PR updates the 'within' to a list while also preserving the functionality of it being treated as string. This ensures no other rules that uses within as string break. The PR was validated by:

2. Reviewing the PR updated logic in context with AC.
4. Ensuring all related testing has been updates.
5. Ensuring new relevant testing has been added. 
6. Manual testing by running with positive dataset in dev editor.
7. Manual testing by running negative dataset in dev editor.

Issue #890, opened 2024-11-05
Rule blocked: CORERULES-716 (SEND-282)


JozefAerts:
SEND-282: PPSPEC: When Multiple specimens are used, then PPSPEC should consist of values from (SPEC) delimited by ";"

My request: would it be possibility to have some functionality to split a variable value in parts by a delimiter (like ";") and then execute the test on each of these parts.

JozefAerts:
Also needed for SEND rule SEND315: "... Treatment values in EXTRT should be separated with a pipe (|)".
In this case, each of the values must then be present in an EXTRT of another record.

PR #1430, opened 2025-11-12
890: column should consist of delimited codelists

alexfurmenkov:
This PR adds support for https://github.com/cdisc-org/cdisc-rules-engine/issues/890.

Suggested rule:
```
Check:
  all:
    - name: $spec_codelist
      operator: not_contains_all
      value: $ppspec_value
Core:
  Id: CDISC.SENDIG.SEND282
  Status: Draft
  Version: "1"
Operations:
  - codelists:
      - SPEC
    level: term
    id: $spec_codelist
    operator: codelist_terms
    returntype: value
  - name: PPSPEC
    delimiter: ;
    id: $ppspec_value
    operator: split_by
Description: If multiple specimen types are used for a calculation (e.g., serum
  and urine for creatinine clearance), then this field should be populated with
  values from the (SPEC) Controlled Terminology codelist delimited by a
  semicolon.
Outcome:
  Message: The multiple specimens are not correctly separated by a semicolon
  Output Variables:
    - $ppspec_value
    - $spec_codelist
```

core.py command:
```
validate
-s
send
-v
1-0
-l
info
-dp
unit-test-coreid-SENDIG282.json
-lr
Rule.yml
-ct
sendct-2025-09-26
```

The idea is:

1. Pull SPEC codelist values from the provided CT package (codelist_terms operation)
2. Split the target column (PPSPEC in this case) by the desired delimiter (split_by operation)
3. For every row, check that the result of step 1 contains all items from step 2. The `Check` section should be interpreted as "Find all rows where `$spec_codelist` doesn't contain all of `$ppspec_value`"

### DIFF
--- a/resources/schema/Operator.md
+++ b/resources/schema/Operator.md
@@ -746,6 +746,22 @@ True if all values in `value` are contained within the variable `name`.
     - "Unplanned Treatment"
 ```
 
+The operator also supports lists:
+
+```yaml
+- name: "$spec_codelist"
+  operator: "contains_all"
+  value: "$ppspec_value"
+```
+
+Where:
+
+| $spec_codelist              |   $ppspec_value    |
+| :-------------------------- | :----------------: |
+| ["CODE1", "CODE2", "CODE3"] | ["CODE1", "CODE2"] |
+| ["CODE1", "CODE2", "CODE3"] | ["CODE2", "CODE3"] |
+| ["CODE1", "CODE2", "CODE3"] |     ["CODE1"]      |
+
 ### not_contains_all
 
 Complement of `contains_all`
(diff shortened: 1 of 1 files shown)


tests/resources/CoreIssue890/Rule.yml
```diff
@@ -0,0 +1,132 @@
...
+            Rule Identifier:
+              Id: "SEND282"
+              Version: "1"
+            Version: "5.0"
+        Version: "1.0"
+Check:
+  all:
+    - name: $ppspec_value
+      operator: not_contains_all
+      value: $spec_codelist
```
gerrycampion: These are swapped. Should be:
```yaml
- name: $spec_codelist
  operator: not_contains_all
  value: $ppspec_value
```
Probably want to update this in the PR description as well

cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -1048,16 +1048,26 @@ def non_empty_within_except_last_row(self, other_value: dict):
     @type_operator(FIELD_DATAFRAME)
     def contains_all(self, other_value: dict):
         target = self.replace_prefix(other_value.get("target"))
+        value_is_literal: bool = other_value.get("value_is_literal", False)
         comparator = other_value.get("comparator")
-        if isinstance(comparator, list):
-            # get column as array of values
-            values = flatten_list(self.value, comparator)
+        if self.is_column_of_iterables(
```
gerrycampion, line 1053: documentation should be updated for `contains_all` to explain that this will work with:
- a column of literals
- a column of lists
- a list of literals

(if this statement is true)

gerrycampion, line 1053: we should have unit tests added for each of these input types

gerrycampion (approved):

Issue #1194, opened 2025-06-02
Rule blocked: CORERULES-903


aniemes:
- Attach test dataset files that can be used to test the new feature
- Replace the "###" in the issue title with the primary JIRA ticket number for the rule. 
- Fill in the following information

**Links to related JIRA Tickets**
- https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-903
- https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-260

**Rule Information**
- **Standard**: SDTM
- **Rule ID**: CG0155
- **Rule ID**: CG0663
- **Rule Description**: Value length = 4 characters beginning with 'AP' and ending with 2 character SDTM domain

**Describe the problem**
The engine does not have the ability to take a part of the DOMAIN (all characters past the first 2) and compare it against some value.

**Describe the solution**
Having an operation that can extract part of the DOMAIN or dataset_names which can then be used to compare against domain_is_custom would be ideal.

**Proposed rule logic**
```yaml
Replace this with an example of how the new solution could be written as a rule
```

**Screenshots**
If applicable, add screenshots to help explain your problem and solution.

gerrycampion:
I think we could make use of the [Dataset Metadata Check](https://cdisc-org.github.io/conformance-rules-editor/#/rule_type?id=dataset-metadata-check) and expose information about the domain, like the AP-related domain.
Then we can use this variable to check against a list of domain names.

PR #1439, opened 2025-11-17
#1194 Add domain suffix extraction and standard domains operation for domain validation

RakeshBobba03:
Implements the ability to extract domain suffixes from AP-related domain names and compare them against valid SDTM domains. This adds a `domain_suffix` option to the `extract_metadata` operation that extracts characters 3-4 from domain names (e.g., "FA" from "APFA"), and introduces a new `standard_domains` operation that returns a sorted list of valid SDTM domain names from the CDISC Library metadata. These changes enable rule CG0663 to validate that Associated Persons (AP) Core Domains have a 4-character length beginning with 'AP' and ending with a valid 2-character SDTM domain.

### DIFF
--- a/cdisc_rules_engine/dataset_builders/dataset_metadata_define_dataset_builder.py
+++ b/cdisc_rules_engine/dataset_builders/dataset_metadata_define_dataset_builder.py
@@ -87,6 +87,15 @@ class DatasetMetadataDefineDatasetBuilder(BaseDatasetBuilder):
             return self.dataset_implementation(columns=define_col_order)
         return self.dataset_implementation.from_records(define_metadata)
 
+    def _ensure_required_columns(self, dataset_df, dataset_col_order):
+        if "dataset_size" not in dataset_df.columns:
+            dataset_df["dataset_size"] = None
+        if "is_ap" not in dataset_df.columns:
+            dataset_df["is_ap"] = False
+        if "ap_suffix" not in dataset_df.columns:
+            dataset_df["ap_suffix"] = ""
+        return self.dataset_implementation(dataset_df[dataset_col_order])
+
     def _get_dataset_dataframe(self):
         dataset_col_order = [
             "dataset_size",
@@ -130,11 +139,7 @@ class DatasetMetadataDefineDatasetBuilder(BaseDatasetBuilder):
                     "domain": "dataset_name",
                 }
                 dataset_df = datasets.rename(columns=data_col_mapping)
-                if "dataset_size" not in dataset_df.columns:
-                    dataset_df["dataset_size"] = None
-                if "is_ap" not in dataset_df.columns:
-                    dataset_df["is_ap"] = False
-                if "ap_suffix" not in dataset_df.columns:
-                    dataset_df["ap_suffix"] = ""
-                dataset_df = self.dataset_implementation(dataset_df[dataset_col_order])
+                dataset_df = self._ensure_required_columns(
+                    dataset_df, dataset_col_order
+                )
         return dataset_df


cdisc_rules_engine/operations/extract_metadata.py
```diff
@@ -9,6 +9,28 @@ def _execute_operation(self):
...
+
+        raw_metadata = self.data_service.get_raw_dataset_metadata(
+            dataset_name=self.params.dataset_path, datasets=self.params.datasets
+        )
+        if raw_metadata and raw_metadata.first_record:
+            domain = raw_metadata.first_record.get("DOMAIN", "")
+            if isinstance(domain, str) and len(domain) >= 4 and domain.startswith("AP"):
+                return domain[2:4]
+
+        return ""
```
gerrycampion: Can you move this logic to be a property, `ap_suffix` of `SDTMDatasetMetadataclass` instead? It would be good to add an `is_ap` property as well.
The top of that file has documentation with several different examples of dataset names. You'll want to add a column for the is_ap and ap_suffix and make sure the values are what you'd expect.

cdisc_rules_engine/operations/extract_metadata.py
```diff
@@ -9,6 +9,28 @@ def _execute_operation(self):
...
         return target_value
+
+    def _extract_domain_suffix(self, metadata: pd.DataFrame) -> str:
+        dataset_name = metadata.get("dataset_name", pd.Series())
+        if dataset_name.empty:
+            return ""
+
+        name = dataset_name.iloc[0]
+        if isinstance(name, str) and len(name) >= 4 and name.startswith("AP"):
+            return name[2:4]
```
gerrycampion: suffix should be derived from domain only, not from dataset name

cdisc_rules_engine/operations/extract_metadata.py
```diff
@@ -9,6 +9,28 @@ def _execute_operation(self):
...
+
+        raw_metadata = self.data_service.get_raw_dataset_metadata(
+            dataset_name=self.params.dataset_path, datasets=self.params.datasets
+        )
+        if raw_metadata and raw_metadata.first_record:
+            domain = raw_metadata.first_record.get("DOMAIN", "")
```
gerrycampion: domain should come from SDTMDatasetMetadata.domain

resources/schema/Operations.md
```diff
@@ -493,6 +515,7 @@ Returns the requested dataset level metadata value for the current dataset. Poss
 - dataset_location
 - dataset_name
 - dataset_label
+- domain_suffix
```
gerrycampion: I think `ap_suffix` or `ap_domain_suffix` would be a better name for this.

cdisc_rules_engine/operations/extract_metadata.py
```diff
@@ -9,6 +9,28 @@ def _execute_operation(self):
             dataset_name=self.params.dataset_path, datasets=self.params.datasets
         )
 
+        if self.params.target == "domain_suffix":
```
gerrycampion: For the rule to work, I think you'll need a param for `is_ap` and `domain` as well so that the length of the domain can be tested.

cdisc_rules_engine/operations/extract_metadata.py
gerrycampion, line 1: Can you also add the new vars (is_ap, ap_suffix) to the following rule types:
ContentsDefineDatasetBuilder
ContentMetadataDatasetBuilder
DatasetMetadataDefineDatasetBuilder
ValueCheckDatasetMetadataDatasetBuilder

cdisc_rules_engine/operations/standard_domains.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class StandardDomains(BaseOperation):
+    def _execute_operation(self):
+        standard_data: dict = self.library_metadata.standard_metadata
+        domains = standard_data.get("domains", set())
+        if isinstance(domains, set):
+            return sorted(list(domains))
+        elif isinstance(domains, (list, tuple)):
+            return sorted(list(domains))
```
gerrycampion: These cases should be combined.

tests/unit/test_operations/test_extract_metadata.py
```diff
@@ -41,3 +42,126 @@ def test_extract_metadata_get_dataset_name(
...
+        mock_service.get_raw_dataset_metadata = MagicMock(return_value=raw_metadata)
+    return mock_service
+
+
+@pytest.mark.parametrize(
+    "dataset_name, first_record, expected_suffix",
+    [
+        ("APFA", None, "FA"),
+        ("APXX", None, "XX"),
+        ("APLB", None, "LB"),
```
gerrycampion: these cases should have "" as expected_suffix since the domain can't be determined. This should fix itself after you make the mentioned changes above.

gerrycampion:
Also, can you provide a sample rule and test data to completely test CG0663. You can attach to this PR. (it will also help with your demo)

RakeshBobba03:
> Also, can you provide a sample rule and test data to completely test CG0663. You can attach to this PR. (it will also help with your demo)

Here are the dataset files and Rule file I used for testing:

[local-rule.yaml](https://github.com/user-attachments/files/23734639/local-rule.yaml)
[unit-test-coreid-CG0663-negative 3.xlsx](https://github.com/user-attachments/files/23734640/unit-test-coreid-CG0663-negative.3.xlsx)
[CG0663_TEST.xlsx](https://github.com/user-attachments/files/23734652/CG0663_TEST.xlsx)

cdisc_rules_engine/operations/extract_metadata.py
```diff
@@ -9,6 +9,12 @@ def _execute_operation(self):
             dataset_name=self.params.dataset_path, datasets=self.params.datasets
         )
 
+        if self.params.target in ("ap_suffix", "ap_domain_suffix", "domain_suffix"):
+            raw_metadata = self.data_service.get_raw_dataset_metadata(
+                dataset_name=self.params.dataset_path, datasets=self.params.datasets
+            )
+            return raw_metadata.ap_suffix
+
```
gerrycampion: Because you added ap_suffix to the metadata here: https://github.com/cdisc-org/cdisc-rules-engine/pull/1439/files#diff-c3aa3814fcb5bc0f31749463bdd48614d76cd6ea3459d64f9ce95f8e82b8595aR227
You no longer need this chunk of code.

cdisc_rules_engine/models/sdtm_dataset_metadata.py
```diff
@@ -12,21 +12,21 @@ class SDTMDatasetMetadata(DatasetMetadata):
...
+    | QSXX     | QS           | False   | QS     | None    | False |           |
+    | SUPPQS   | SUPPQS       | True    | None   | QS      | False |           |
+    | SUPPQSX  | SUPPQS       | True    | None   | QS      | False |           |
+    | SUPPQSXX | SUPPQS       | True    | None   | QS      | False |           |
+    | APQS     | APQS         | False   | APQS   | None    | True  | QS        |
+    | APQSX    | APQS         | False   | APQS   | None    | True  | QS        |
+    | APQSXX   | APQS         | False   | APQS   | None    | True  | QS        |
+    | SQAPQS   | SQAPQS       | True    | None   | APQS    | False |           |
+    | SQAPQSX  | SQAPQS       | True    | None   | APQS    | False |           |
+    | SQAPQSXX | SQAPQS       | True    | None   | APQS    | False |           |
```
gerrycampion: is_ap should be True for these AP supps. I don't know what the ap_suffix should be, I think it is fine to leave blank until we come across a rule that needs it.  Make sure you add tests for these too.

cdisc_rules_engine/models/sdtm_dataset_metadata.py
```diff
@@ -57,3 +57,24 @@ def unsplit_name(self) -> str:
...
+    @property
+    def is_ap(self) -> bool:
+        """
+        Returns true if domain starts with AP and is at least 4 characters long
+        """
+        return (
+            isinstance(self.domain, str)
+            and len(self.domain) >= 4
+            and self.domain.startswith("AP")
+        )
```
gerrycampion: Sorry, I didn't catch this earlier. For this test, we should move the logic from here: https://github.com/cdisc-org/cdisc-rules-engine/pull/1439/files#diff-c3aa3814fcb5bc0f31749463bdd48614d76cd6ea3459d64f9ce95f8e82b8595aL252-L256
into this file.
This means we will look for the existence of the  topic variable, APID in the first record instead of looking at the domain value. Rule CG0155 confirms we should do it this way because otherwise the rule would be self-referential and would never trigger an issue.

You'll probably need to refactor base_data_service a little bit to use the sdtm dataset metadata method instead of its own method.

For supps, I think you'll need to check the rdomain value with this existing logic, except use `== 4` instead of `>= 4`, because we are looking at domain names, not dataset names.

tests/unit/test_utilities/test_utils.py
```diff
@@ -45,6 +45,53 @@ def test_is_supp_dataset(mock_dataset, expected):
...
+    ({"first_record": {"DOMAIN": "APQS"}}, True),
+    ({"first_record": {"DOMAIN": "APFAMH"}}, True),
+    ({"first_record": {"DOMAIN": "AE"}}, False),
+    ({"first_record": {"DOMAIN": "LB"}}, False),
+    ({"first_record": {"DOMAIN": "AP"}}, False),
+    ({"first_record": {"DOMAIN": "APF"}}, False),
+    ({"first_record": None}, False),
+    ({"first_record": {}}, False),
+    ({}, False),
+]
```
gerrycampion, line 69: first record will need to include an APID variable for some of these tests to pass after the update to is_ap

gerrycampion (changes requested):

cdisc_rules_engine/operations/extract_metadata.py
gerrycampion, line 1: See my previous comment. This file still has modifications but it can be reverted completely

cdisc_rules_engine/services/data_services/base_data_service.py
```diff
@@ -243,38 +246,37 @@ def _handle_special_cases(
             if self._contains_topic_variable(dataset, dataset_metadata.domain, "OBJ"):
                 return FINDINGS_ABOUT
             return FINDINGS
-        if self._is_associated_persons(dataset):
+        if dataset_metadata.is_ap and not dataset_metadata.is_supp:
```
gerrycampion: I think you don't need `and not dataset_metadata.is_supp` here, because this condition is already handled by the line: `if self._contains_topic_variable(dataset, dataset_metadata.domain, "QNAM"):`

tests/unit/test_operations/test_extract_metadata.py
```diff
@@ -41,3 +42,132 @@ def test_extract_metadata_get_dataset_name(
...
+        {"STUDYID": ["TEST_STUDY"], "DOMAIN": ["AE"]}
+    )
+    operation_params.target = "ap_suffix"
+    cache = InMemoryCacheService.get_instance()
+    operation = ExtractMetadata(
+        operation_params, operation_params.dataframe, cache, mock_data_service
+    )
+    result = operation.execute()
+    assert operation_params.operation_id in result
+    assert all(item == "" for item in result[operation_params.operation_id])
```
gerrycampion: Would it make sense to combine these two sets of tests into a single method?
- "" can be the expected_suffix for the second set
- In both sets, operation_params.dataframe can be derived from the first_record param.

gerrycampion (changes requested):

PR #1440, opened 2025-11-17
1327: Fix dataset metadata handling in DatasetMetadataDefineDatasetBuilder

alexfurmenkov:


cdisc_rules_engine/dataset_builders/dataset_metadata_define_dataset_builder.py
```diff
@@ -93,19 +94,24 @@ def _get_dataset_dataframe(self):
...
-                datasets.data = (
-                    ds_metadata.data
-                    if datasets.data.empty
-                    else datasets.data.append(ds_metadata.data)
-                )
+                if ds_metadata:
+                    if datasets.data.empty:
+                        datasets.data = ds_metadata.data.copy()
+                    else:
+                        datasets.data = pd.concat(
```
RamilCDISC: using pd.concat will return a pandas dataframe. Here dataframe could be either pandas or dask.

RamilCDISC (changes requested):

cdisc_rules_engine/services/data_services/dummy_data_service.py
```diff
@@ -156,7 +156,22 @@ def __get_dataset_metadata(self, dataset_name: str, **kwargs) -> dict:
         return metadata_to_return
 
     def to_parquet(self, file_path: str) -> str:
```
SFJohnson24, line 158: @alexfurmenkov why are you updating the to_parquet method for dummy_data_service?  Dummy is for the TestRule Azure Function hosted endpoint for CORE and will never be in parquet format.  I was able to run CG0320 just fine without this codeblock?

alexfurmenkov: Otherwise it just doesn't work if ran with DASK

cdisc_rules_engine/dataset_builders/dataset_metadata_define_dataset_builder.py
```diff
@@ -55,9 +56,15 @@ def build(self):
...
-        matching_row = merged_cleaned[
+        matching_row: DatasetInterface = merged_cleaned[
             merged_cleaned["dataset_location"].str.lower() == dataset_filename
         ]
+        if matching_row.empty:
+            # when using DASK dataset_filename refers to temp parquet filename
+            matching_row: DatasetInterface = merged_cleaned[
+                merged_cleaned["dataset_domain"].str.upper()
+                == self.dataset_metadata.domain.upper()
+            ]
```
gerrycampion, line 67: I believe this is the purpose of [DatasetMetada.original_path](https://github.com/cdisc-org/cdisc-rules-engine/blob/0fa6d885e71ebe909d6b210cd5bd0f45ec8dbaaa/cdisc_rules_engine/models/dataset_metadata.py#L19)
Can you use this? Will this current line still work if we are dealing with a non-domain dataset?

alexfurmenkov: Makes sense. I've updated it to use self.dataset_metadata.original_path

alexfurmenkov: @gerrycampion wait... what's the purpose of `original_path`? To store the original path to the dataset file like `CDISC01/test/ae.xpt`?

I am asking because I didn't find any code that actually passes this parameter either to `DatasetMetadata` or `SDTMDatasetMetadata`.

So I think we need to do 2 things:
1. Update this code to try using original_path and fallback to the domain if original_path is null.
2. Update the data services to populate original_path.

alexfurmenkov: @SFJohnson24 I would be thankful if you could also hop in here

gerrycampion: https://github.com/cdisc-org/cdisc-rules-engine/blob/bde3df2c1cce87965dd0aa5f5dacfe0d6366adfb/scripts/run_validation.py#L168

alexfurmenkov: aah, I see now, thanks. Yeah, we can use `original_path` here for sure. I will push the fix for tests tomorrow morning.

cdisc_rules_engine/dataset_builders/dataset_metadata_define_dataset_builder.py
```diff
@@ -55,9 +56,21 @@ def build(self):
...
+            if self.dataset_metadata.original_path:
+                matching_row: DatasetInterface = merged_cleaned[
+                    merged_cleaned["dataset_location"].str.lower()
+                    == self.dataset_metadata.original_path.lower()
+                ]
+            else:
+                matching_row: DatasetInterface = merged_cleaned[
+                    merged_cleaned["dataset_domain"].str.upper()
+                    == self.dataset_metadata.domain.upper()
+                ]
```
gerrycampion: I still have an issue with falling back to dataset_domain (unless you actually mean domain). Wouldn't it be better to fall back to the name. See this chart:
https://github.com/cdisc-org/cdisc-rules-engine/blob/0afedf3f925e245aa6ec4e99b2b532ba526d420d/cdisc_rules_engine/models/sdtm_dataset_metadata.py#L14-L29

gerrycampion: Do we need this at all? In what scenario would this happen?

gerrycampion (approved):

RamilCDISC (approved):
The PR fixes the dataset metadata handling. The updates were validated by:

2. Reviewing the PR logic in accordance with the AC.
3. Ensuring all unit and regression testing pass
4. Ensuring all relevant testing is updated.
5. Ensuring compatibility for Pandas and DASK both.
6. Ensuring previous coding practices are followed.
7. Running manual testing with positive dataset
8. Running manual testing with negative dataset.

PR #1445, opened 2025-11-20
Errorhandling preprocessor

SFJohnson24:
This PR adds custom exceptions that get raised during several key parts of validation that were skipped by error coverage
- operation execution
- dataset builder build()
- preprocessing of relrec/supp/match datasets
Custom exceptions and messages with key metadata have been created to propagate up to validate single dataset which contains the try/except block that handles exceptions.  

the attached issue has a rule/dataset that fails with relrec merge.  Of note--i kept the caught stack trace as it is useful for development resolution of bugs but the added metadata context in the error messages should help rule authors with.  Also of note--I was uncertain whether to assign a rule skip or not for these---I though it would be worthwhile to skip the rule to continue execution despite these being 'execution errors'.

### DIFF
--- a/cdisc_rules_engine/utilities/dataset_preprocessor.py
+++ b/cdisc_rules_engine/utilities/dataset_preprocessor.py
@@ -50,6 +50,10 @@ class DatasetPreprocessor:
     def preprocess(  # noqa
         self, rule: dict, datasets: Iterable[SDTMDatasetMetadata]
     ) -> DatasetInterface:
+        """
+        Preprocesses the dataset by merging it with the
+        datasets from the provided rule.
+        """
         rule_datasets: List[dict] = rule.get("datasets")
         if not rule_datasets:
             return self._dataset  # nothing to preprocess
@@ -543,7 +547,9 @@ class DatasetPreprocessor:
                     f"Match keys: {match_keys}, "
                     f"Error: {str(e)}"
                 )
-        elif right_dataset_domain_name == "SUPP--":
+        elif right_dataset_domain_name.startswith(
+            "SUPP"
+        ) or right_dataset_domain_name.startswith("SQ"):
             try:
                 result: DatasetInterface = DataProcessor.merge_pivot_supp_dataset(
                     dataset_implementation=self._data_service.dataset_implementation,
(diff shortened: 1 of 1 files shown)


cdisc_rules_engine/utilities/dataset_preprocessor.py
```diff
@@ -49,10 +50,6 @@ def __init__(
     def preprocess(  # noqa
         self, rule: dict, datasets: Iterable[SDTMDatasetMetadata]
     ) -> DatasetInterface:
-        """
```
RamilCDISC, line 52: why are we removing the docstring here?

cdisc_rules_engine/utilities/dataset_preprocessor.py
```diff
@@ -457,7 +495,7 @@ def _update_dataset_with_merged_records(
         updated_data = pd.concat([remaining_records, merged_records], ignore_index=True)
         return self._dataset.__class__(data=updated_data)
 
-    def _merge_datasets(
+    def _merge_datasets(  # noqa
```
RamilCDISC, line 502: why are we adding a noqa here?

SFJohnson24: nesting things in try/except put it over the complexity limit for a function.  It makes sense to keep the merge logic together as it was so I added the noqa

cdisc_rules_engine/utilities/dataset_preprocessor.py
SFJohnson24: this was a mistake, I will restore it.

SFJohnson24:
I also looked into the ticket Gerry referenced today in review #1432  This logic does not interfere with that portion of the codebase

RamilCDISC (approved):
The PR adds exception handling to improve error handling. This helps skipping rules instead of just breaking the execution. The PR was validated by: 

1. Reviewing the PR code for any unwanted changes or comments.
2. Reviewing the PR Logic in context of AC.
3. Validating all unit and regression testing pass. 
4. Ensuring all related testing is updated.
5. Ensuring all the error message are clear and coherent.
6. Ensuring the new custom exceptions follow the pattern of previous custom exceptions.
7. Ensuring the exception is fully representative of code in the try block.
8. Ensuring the validation does not break where it should just skip the in-processing rule.

Issue #1017, opened 2025-02-11
CORERULES-9645 - is_column_of_iterables method only checks first row


ASL-rmarshall:
**Links to related JIRA Tickets**
- https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-9645

**Rule Information**
- **Standard**: USDM
- **Rule ID**: DDF00102
- **Rule Description**: A scheduled activity instance must only reference a timeline exit that is defined within the same schedule timeline as the scheduled activity instance.

**Describe the bug**
As shown in "Request.txt" attached below, this rule was originally implemented by using the `distinct` operator to retrieve a list of the distinct schedule timeline exit `id` values grouped by `parent_id`, which was merged onto the check dataset.  Error records were reported when the value on the `timelineExitId` variable in the check dataset `is_not_contained_by` the unique list of schedule timeline exit `id` values.  However, the rule failed with an error:
> Cannot perform 'and_' with a dtyped [bool] array and scalar of type [NoneType]

This was happening because the grouped list of unique values was only populated on records with matching group values - it was `None` for other records.  The `is_column_of_iterables` method is used to check if the comparison column contains iterables to determine the method used to do the "contains" comparison. However, as `is_column_of_iterables` only checks the first row, the comparison method used when the first row contains an iterable fails for any other row that contains `None`.

The same issue occurs if the `does_not_contain` operator is used.

Note that this rule has now been rewritten to use a table join instead, so this is no longer a rule-blocking issue.

**Error returned from Rule Engine**
```json
    "ScheduledActivityInstance": [
        {
            "executionStatus": "execution_error",
            "dataset": "ScheduledActivityInstance.xpt",
            "domain": "",
            "variables": [],
            "message": "rule execution error",
            "errors": [
                {
                    "dataset": "ScheduledActivityInstance.xpt",
                    "error": "An unknown exception has occurred",
                    "message": "Cannot perform 'and_' with a dtyped [bool] array and scalar of type [NoneType]"
                }
            ]
        }
    ],
```

**Expected behavior**
The `is_not_contained_by` and `does_not_contain` operators should run without error when comparing a variable value against a grouped list of unique values generated by the `distinct` operator.

**Test data**

[Request.txt](https://github.com/user-attachments/files/18766543/Request.txt)

Test data is also available in Excel format in the USDM > DDF00102 > positive > 01 > data folder of the CORE unitTesting SharePoint.

ASL-rmarshall:
[...]

PR #1446, opened 2025-11-20
#1017 Fix is_column_of_iterables to check all values instead of just first row

RakeshBobba03:
This PR fixes a bug in `is_column_of_iterables` where it only checked the first row (`iloc[0]`) to determine if a column contained iterables, which could cause misclassification when the first row was `None` but later rows were lists or sets. The fix now checks all values in the column, returning `True` only when all values are iterables (lists or sets), not `None`, and not `NaN`. This prevents incorrect behavior in containment operators (`contains`, `does_not_contain`, `is_contained_by`) when used with grouped `distinct` operations that produce columns with mixed types (e.g., `None` for groups with no matches, lists for groups with matches). The PR includes tests for the bug scenario (None-first-row cases) and updates the operator documentation in `Operator.md` to explain the iterable detection behavior.

### DIFF
--- a/resources/schema/Operator.md
+++ b/resources/schema/Operator.md
@@ -452,15 +452,26 @@ Date and time specific operations for comparing dates, validating date completen
 
 Date comparison. Compare `name` to `value`. Compares partial dates if `date_component` is specified.
 
+The `date_component` parameter accepts: `"year"`, `"month"`, `"day"`, `"hour"`, `"minute"`, `"second"`, `"microsecond"`, or `"auto"`.
+
+When `date_component: "auto"` is used, the operator automatically detects the precision of both dates and compares at the common (less precise) level.
+
+```yaml
+- name: "AESTDTC"
+  operator: "date_equal_to"
+  value: "RFSTDTC"
+  date_component: "auto"
+```
+
 ### date_not_equal_to
 
 Complement of `date_equal_to`
 
-Date comparison. Compare `name` to `value`. Compares partial dates if `date_component` is specified.
+Date comparison. Compare `name` to `value`. Compares partial dates if `date_component` is specified. Supports `date_component: "auto"`.
 
 ### date_greater_than
 
-Date comparison. Compare `name` to `value`. Compares partial dates if `date_component` is specified.
+Date comparison. Compare `name` to `value`. Compares partial dates if `date_component` is specified. Supports `date_component: "auto"`.
 
 > Year part of BRTHDTC > 2021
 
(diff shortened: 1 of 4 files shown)


resources/schema/Operator.md
```diff
@@ -299,6 +299,8 @@ Complement of `suffix_equal_to`
 
 Will return True if the value in `value` is contained within the collection/iterable in the target column, or if there's an exact match for non-iterable data.
 
+The operator determines if a column contains iterables by checking that **all** values in the column are iterables (lists or sets), not `None`, and not `NaN`. If all values are iterables, the operator performs row-by-row comparison. If the column has mixed types (e.g., `None` and lists), the column is not treated as iterables and the operator uses value vs all column values logic.
```
SFJohnson24: this is a bit verbose. A lot of the rule authors are non-technical and this maybe leans too technical so I would avoid the term iterables.  `The operator checks if every value in a column is a list or set. If yes, it compares row-by-row. If any value is blank or a different type (like a string or number), it compares each value against the entire column instead.`

resources/schema/Operator.md
```diff
@@ -311,6 +313,8 @@ Example:
 
 Complement of `contains`. Returns True when the value is NOT contained within the target collection.
 
+The operator uses the same iterable detection logic as `contains`: all values must be iterables for the column to be treated as iterables.
```
SFJohnson24: I am not sure if this is necessary since the docs already state it is the complement of contains.

resources/schema/Operator.md
```diff
@@ -638,6 +642,8 @@ Testing whether individual values or string parts belong to specific lists or se
 
 Value in `name` compared against a list in `value`. The list can have literal values or be a reference to a `$variable`.
 
+When the `value` parameter references a column variable (e.g., `$variable`), the operator determines if that column contains iterables by checking that **all** values in the column are iterables (lists or sets), not `None`, and not `NaN`. If all values are iterables, the operator performs row-by-row comparison. If the column has mixed types, the operator uses value vs all column values logic.
```
SFJohnson24: I dont know if this needs to be restated. I would just reference what you already wrote above.  'this operator behaves similarly to contains.  The key distinction: contains checks if comparator ∈ target, while is_contained_by checks if target ∈ comparator'

tests/unit/test_check_operators/test_containment_checks.py
```diff
@@ -417,3 +417,97 @@ def test_is_not_contained_by_case_insensitive(
         {"target": "target", "comparator": comparator}
     )
     assert result.equals(df.convert_to_series(expected_result))
+
```
SFJohnson24, line 420: these tests look good

PR #1447, opened 2025-11-22
1402 cg0546 support multi key param update (Execution error fix)

RakeshBobba03:
Fixed MultiIndex handling to properly detect when all index levels need to be removed. Instead of calling `droplevel()` on a MultiIndex with equal levels, the code now extracts the innermost level values and reconstructs a Series with the correct row index mapping.
Added checks to handle cases where `groupby().apply()` returns a DataFrame instead of a Series, extracting the first column to ensure consistent Series handling.
Updated `check_basic_sort_order` to enforce strict ordering by using `>=` for ascending checks and `<=` for descending checks. This correctly flags identical values (e.g., "DIAG", "DIAG") as violations, ensuring the operator detects missing sequence numbers as required by rules like CG0546.
Fixed index alignment issues that were masking errors. Both `basic_sort_check` and `date_overlap_check` are now reindexed to `sorted_df.index` before combining, ensuring all boolean Series are properly aligned and preventing `fill_value=True` from masking actual violations due to index mismatches.

### DIFF
--- a/tests/unit/test_check_operators/test_relationship_integrity_checks.py
+++ b/tests/unit/test_check_operators/test_relationship_integrity_checks.py
@@ -1255,6 +1255,124 @@ def test_target_is_sorted_by_multiple_within_not_sorted(dataset_class):
     )
 
 
+_COMPARATOR_SMSTDTC = [
+    {"name": "SMSTDTC", "sort_order": "ASC", "null_position": "last"}
+]
+_INVALID_INPUT_DF = {
+    "USUBJID": ["001", "002"],
+    "MIDS": ["A1", "A2"],
+    "SMSTDTC": ["2005-01-01", "2006-01-01"],
+}
+
+
+@pytest.mark.parametrize("dataset_class", [PandasDataset, DaskDataset])
+def test_target_is_sorted_by_and_complement_multi_within(dataset_class):
+    data = {
+        "USUBJID": ["001", "001", "003", "003"],
+        "MIDSTYPE": ["DIAGNOSIS", "DIAGNOSIS", "RELAPSE", "RELAPSE"],
+        "MIDS": ["DIAG1", "DIAG2", "RELAPSE2", "RELAPSE1"],
+        "SMSTDTC": ["2005-01-01", "2006-01-01", "2005-01-01", "2006-01-01"],
+    }
+    other_value = {
+        "target": "MIDS",
+        "within": ["USUBJID", "MIDSTYPE"],
+        "comparator": _COMPARATOR_SMSTDTC,
+    }
+    sorted_expected = [True, True, False, False]
+    df = dataset_class.from_dict(data)
+    dt = DataframeType({"value": df})
+    assert dt.target_is_sorted_by(other_value).equals(
+        df.convert_to_series(sorted_expected)
+    )
+    assert dt.target_is_not_sorted_by(other_value).equals(
+        df.convert_to_series([not x for x in sorted_expected])
+    )
+
+
+@pytest.mark.parametrize(
+    "other_value,exc_type,match",
+    [
+        (
+            {"target": "MIDS", "comparator": _COMPARATOR_SMSTDTC},
+            ValueError,
+            "within parameter is required",
+        ),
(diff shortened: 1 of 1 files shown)


tests/unit/test_check_operators/test_relationship_integrity_checks.py
gerrycampion, line 1: Why were all of these test results changed? I looked over the first sample and it was originally correct. I didn't check the rest, but please justify the changes. If you aren't sure of the intent behind these operators, we can discuss.

cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -1611,16 +1611,27 @@ def safe_compare(x, index):
...
+                is_sorted.iloc[sorted_indices[i]] = False
+                is_sorted.iloc[sorted_indices[i + 1]] = False
+            elif ascending:
+                if current >= next_val:
+                    is_sorted.iloc[sorted_indices[i]] = False
+                    is_sorted.iloc[sorted_indices[i + 1]] = False
+            else:
+                if current <= next_val:
+                    is_sorted.iloc[sorted_indices[i]] = False
+                    is_sorted.iloc[sorted_indices[i + 1]] = False
```
gerrycampion: There is a lot of branching logic here. All of these conditions result in the same 2 lines being executed. Those two lines should only appear once. You should combine all conditions into a **_single_** condition separated by `or/and`'s

cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -1643,6 +1654,41 @@ def check_date_overlaps(self, group, target, comparator):
 
         return is_sorted
 
+    def _process_grouped_result(
```
gerrycampion, line 1731: a more complex method like this requires some documentation of what it is doing

cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -1670,36 +1716,95 @@ def target_is_sorted_by(self, other_value: dict):
...
 
     @log_operator_execution
     @type_operator(FIELD_DATAFRAME)
     def target_is_not_sorted_by(self, other_value: dict):
-        return ~self.target_is_sorted_by(other_value)
+        target: str = self.replace_prefix(other_value.get("target"))
+        within_columns = self._normalize_grouping_columns(other_value.get("within"))
+        columns = other_value["comparator"]
+        result = pd.Series([False] * len(self.value), index=self.value.index)
+        is_pandas_dataset = isinstance(self.value, PandasDataset)
```
gerrycampion: this variable is only used in _process_grouped_result, so it should be created in that method instead of passed from this method.

cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -1643,6 +1654,41 @@ def check_date_overlaps(self, group, target, comparator):
...
+        grouped_result,
+        grouped_df,
+        within_columns,
+        sorted_df,
+        is_pandas_dataset,
+        check_func=None,
+    ):
+        if isinstance(grouped_result, pd.DataFrame):
+            grouped_result = grouped_result.iloc[:, 0]
+        if is_pandas_dataset and isinstance(grouped_result.index, pd.MultiIndex):
```
gerrycampion: whenever you have to check isintance(pandas), it is usually a sign that this should be handled by the [DatasetInterface](https://github.com/cdisc-org/cdisc-rules-engine/blob/f38215e4a067f5118c83e4795d6e94d6b18fd0b8/cdisc_rules_engine/models/dataset/dataset_interface.py)

gerrycampion (changes requested):

tests/unit/test_check_operators/test_relationship_integrity_checks.py
gerrycampion, line 1: I still disagree with the changes to these test results. I checked the first set and the original results look correct.

gerrycampion (approved):

PR #1449, opened 2025-11-24
1421: load variables ccode correctly

alexfurmenkov:


cdisc_rules_engine/rules_engine.py
```diff
@@ -335,8 +335,6 @@ def validate_rule(
         elif (
             rule.get("rule_type")
             == RuleTypes.VARIABLE_METADATA_CHECK_AGAINST_DEFINE.value
-            or rule.get("rule_type")
```
RamilCDISC, line 338: Could you please add details to the PR explaining why this check was removed and summarize your overall work?

RamilCDISC (changes requested):

cdisc_rules_engine/rules_engine.py
alexfurmenkov: After I fixed the dataset builder to correctly load `library_variable_ccode` the rule still wasn't working as expected, the equality checks just didn't work. 

I've started diving in and noticed that after calling `add_comparator_to_rule_conditions` on this rule it becomes corrupt:

Image

Pay attention to the conditions 2 and 3 - comparator values become `define_library_variable_ccode` whereas they should be `library_variable_ccode` and `define_variable_ccode`

cdisc_rules_engine/dataset_builders/base_dataset_builder.py
SFJohnson24, line 1: I am getting an error when I run Marcelina's data.
On line 159
```     def get_define_xml_variables_metadata(self) -> List[dict]:
        """
        Gets Define XML variables metadata.
        """
        define_xml_reader = DefineXMLReaderFactory.get_define_xml_reader(
            self.dataset_path, self.define_xml_path, self.data_service, self.cache
        )
        return define_xml_reader.extract_variables_metadata(
            domain_name=self.dataset_metadata.domain
        )
```  
extract_variables_metadata is being called with the None for domain for SUPPEC.  We need a check for self.dataset_metadata.is_supp() and if so call extract_variables_metadata with the rdomain property.  You could also do a check if domain exists, and if not, use rdomain.

the other issue I see would occur downstream from this: _get_domain_metadata() in base_define_xml__reader is looking only based on domain and grabbing the first that matches.  This will be an issue when differentiating Supp-- from Parents.  
image
I suspect passing the name and the domain/rdomain may be necessary and using the Name and the Domain property of the itemgroupdef to pick which metadata to grab would be best plan of attack

SFJohnson24 (changes requested):
see comment

cdisc_rules_engine/dataset_builders/base_dataset_builder.py
```diff
@@ -156,9 +156,13 @@ def get_define_xml_variables_metadata(self) -> List[dict]:
         define_xml_reader = DefineXMLReaderFactory.get_define_xml_reader(
             self.dataset_path, self.define_xml_path, self.data_service, self.cache
         )
-        return define_xml_reader.extract_variables_metadata(
-            domain_name=self.dataset_metadata.domain
-        )
+        # If domain is not set and this is a SUPP domain, use rdomain
+        domain = self.dataset_metadata.domain
+        if not domain and getattr(self.dataset_metadata, "is_supp", False):
```
SFJohnson24, line 161: this is the problem I described in my comment above.  "the other issue I see would occur downstream from this: _get_domain_metadata() in base_define_xml__reader is looking only based on domain and grabbing the first that matches. This will be an issue when differentiating Supp-- from Parents."
image
we need to update extract_variables_metadata to not only use domain but also name as if we use the rdomain--it will grab the parent as they both have domain set to the same thing.  There is a name property 'SUPP--' that will be the same as self.dataset_metadata.name.  This can be used in the case of supps to correctly retreive the supp data and not the parent dataset metadata

SFJohnson24 (changes requested):
see comment

Issue #1023, opened 2025-02-13
Ensure UTF Support in Report Generator


SFJohnson24:
The report generator currently outputs text in ASCII, which may cause character encoding issues when processing reports containing non-ASCII characters. We need to update the generator to support Unicode properly to ensure correct rendering of international characters.

A/C:
Ensure generated reports support UTF-8 encoding by default.
Test with datasets containing multilingual characters.
Verify correct rendering in different output formats (JSON, XLSX)

DmitryMK:
Waiting for #1022 to be merged to progress on this ticker.

gerrycampion:
@DmitryMK #1022 has been merged. Are you still planning on working on this ticket?

DmitryMK:
@gerrycampion yes, I will finish it. I wrote tests back in December so will just need to merge everything with the latest main branch. I will do it later this week.

PR #1456, opened 2025-11-26
1022 enhance utf handling

RakeshBobba03:
This PR adds UTF encoding support to the dataset readers to handle international characters and non-ASCII data. It adds an optional --encoding CLI parameter that propagates through the validation pipeline to all data readers and metadata readers. When encoding is not specified, the readers automatically detect encoding with fallbacks: JSON/NDJSON readers try UTF-8, UTF-16, and UTF-32 in sequence, while XPT readers try UTF-8, UTF-16, UTF-32, cp1252, and latin-1 to handle smart quotes and other Windows-1252 characters. This resolves UnicodeDecodeError issues when processing JSON files with international characters and XPT files containing non-UTF-8 characters from Excel exports. All readers (JSONReader, DatasetJSONReader, DatasetNDJSONReader, XPTReader) and metadata readers have been updated to support both explicit encoding specification and automatic detection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ chmod +x ./core
 
 2. Install dependencies:
    ```bash
-   pip install -r requirements.txt -r requirements-dev.txt
+   python -m pip install -r requirements-dev.txt
    ```
 
 3. Run validation:


tests/unit/test_dataset_json_reader.py
```diff
@@ -10,7 +10,9 @@ def test_from_file():
         f"{os.path.dirname(__file__)}/../resources/test_dataset.json"
     )
 
-    reader = DatasetJSONReader()
+    from cdisc_rules_engine.models.dataset.pandas_dataset import PandasDataset
```
RamilCDISC: Could you please move these to the top of the file among the other imports for consistency.

tests/unit/test_dataset_ndjson_reader.py
```diff
@@ -10,7 +10,9 @@ def test_from_file():
         f"{os.path.dirname(__file__)}/../resources/test_dataset.ndjson"
     )
 
-    reader = DatasetNDJSONReader()
+    from cdisc_rules_engine.models.dataset.pandas_dataset import PandasDataset
```
RamilCDISC: here too.

tests/unit/test_xpt_reader.py
```diff
@@ -10,7 +10,9 @@ def test_read():
     with open(test_dataset_path, "rb") as f:
         data = f.read()
 
-    reader = XPTReader()
+    from cdisc_rules_engine.models.dataset.pandas_dataset import PandasDataset
```
RamilCDISC: and here.

core.py
```diff
@@ -346,6 +346,16 @@ def _validate_no_arguments(logger) -> None:
         "If true, limits reported issues per dataset per rule."
     ),
 )
+@click.option(
+    "--encoding",
```
RamilCDISC, line 398: Could you please add the short form flag also for encoding for consistency. I think we should fix the options of encoding you can provide to this flag. This will ensure that only valid encoding names pass to internal engine functionalities. Also please update the readme.md for the new flag.

cdisc_rules_engine/services/data_readers/dataset_ndjson_reader.py
```diff
@@ -16,16 +16,39 @@
...
     def read_json_file(self, file_path: str) -> dict:
-        with open(file_path, "r") as file:
-            lines = file.readlines()
-        return json.loads(lines[0]), [json.loads(line) for line in lines[1:]]
+        if self.encoding:
+            with open(file_path, "r", encoding=self.encoding) as file:
+                lines = file.readlines()
+            return json.loads(lines[0]), [json.loads(line) for line in lines[1:]]
+
+        encodings_to_try = ["utf-8", "utf-16", "utf-32"]
```
DmitryMK: Should this be done by a user? I.e. a default encoding should be UTF-8 and if it fails a user gets a message and either updates the dataset to use UTF-8 or changes the encoding parameter value.

DmitryMK:
@RakeshBobba03 @SFJohnson24 @gerrycampion the current approach tries to read as UTF8/16/32 for JSON and UTF8/latin1/cp1252/UTF16/32 for XPT. 

Another reason to avoid it is that theoretically data can be encoded with one encoding, but will be read without errors by another encoding. For example: "Ã£" in latin1 is 'ã' in UTF-8, so if "UTF-8" is used first, it will work, but the result would be different if it was read with latin1. If we specify that default value is UTF-8 it will be more straight forward and clear for a user.

I think it is better to select an explicit approach, where default encoding is UTF8 and later a user can specify other encoding. It will impact [1023](https://github.com/cdisc-org/cdisc-rules-engine/issues/1023), because if one of the encodings worked, we will need to push it to the top level, so that later it can be used by the report (the report should use the same encoding as reader). 

I also think it can be important for a user to control which encoding is used, because when a regulatory agency is reading/checking the data, they need to understand what is the correct encoding. 

Do you think we can change it?

cdisc_rules_engine/services/data_readers/data_reader_factory.py
```diff
@@ -47,7 +53,17 @@ def get_service(self, name: str = None, **kwargs) -> DataReaderInterface:
         """
         service_name = name or self._default_service_name
         if service_name in self._reader_map:
-            return self._reader_map[service_name](self.dataset_implementation)
```
RamilCDISC, line 50: From what I see in the previous logic, self.dataset_implementation was consistently passed to all reader classes in _reader_map.
In the updated logic, however, the USDM reader is instantiated without dataset_implementation. Could you please clarify why dataset_implementation is not being passed here?

RakeshBobba03: My thinking was that the factory pattern already varies what parameters get passed based on what each reader actually needs. For example, `XPTReader`, `DatasetJSONReader`, and `DatasetNDJSONReader` get both `dataset_implementation` and `encoding` because they use both, while `ParquetReader` only gets `dataset_implementation` because that's what it needs (it doesn't accept `encoding` in its `__init__` since it hardcodes UTF-8).

In the case of `JSONReader`, it only reads raw JSON and returns a dictionary, and it never actually creates Dataset objects (that happens later in `USDMDataService`). Also, throughout the codebase, `JSONReader()` is instantiated with no parameters when called directly (like in `datasetjson_metadata_reader.py`, `usdm_data_service.py`, `dummy_data_service.py`, etc.), so I was trying to match that pattern in the factory.

That said, I can see the argument for consistency, it would make the factory code more uniform and predictable. And since `DataReaderInterface` has a default parameter, passing `dataset_implementation` wouldn't break anything, it just wouldn't be used. Looking at it now, I realize that if we want consistency, we could pass `dataset_implementation` to `JSONReader` just like we do for `ParquetReader` (which also gets only `dataset_implementation`, not `encoding`).

I'm a bit torn between keeping it minimal versus keeping it consistent. What are your thoughts on this? Do you think the consistency benefit outweighs the redundancy, or does the current approach make sense?

RamilCDISC: @gerrycampion Could you please add your opinion. What do you think would be better approach for the codebase?

gerrycampion:
> @RakeshBobba03 @SFJohnson24 @gerrycampion the current approach tries to read as UTF8/16/32 for JSON and UTF8/latin1/cp1252/UTF16/32 for XPT.
> 
> Another reason to avoid it is that theoretically data can be encoded with one encoding, but will be read without errors by another encoding. For example: "Ã£" in latin1 is 'ã' in UTF-8, so if "UTF-8" is used first, it will work, but the result would be different if it was read with latin1. If we specify that default value is UTF-8 it will be more straight forward and clear for a user.
> 
> I think it is better to select an explicit approach, where default encoding is UTF8 and later a user can specify other encoding. It will impact [1023](https://github.com/cdisc-org/cdisc-rules-engine/issues/1023), because if one of the encodings worked, we will need to push it to the top level, so that later it can be used by the report (the report should use the same encoding as reader).
> 
> I also think it can be important for a user to control which encoding is used, because when a regulatory agency is reading/checking the data, they need to understand what is the correct encoding.
> 
> Do you think we can change it?

@DmitryMK I agree this approach would make the most sense.
@RakeshBobba03 I see you've already implemented this approach

cdisc_rules_engine/services/data_readers/data_reader_factory.py
```diff
@@ -47,7 +53,17 @@ def get_service(self, name: str = None, **kwargs) -> DataReaderInterface:
...
+            reader_class = self._reader_map[service_name]
```
gerrycampion, line 57: To answer the question, I think the simplest solution is to just add this to the `DataReaderInterface` init params. The implementing classes can decide whether or not to use it. No need for the different conditions in the factory.

cdisc_rules_engine/services/data_readers/dataset_json_reader.py
```diff
@@ -15,14 +15,18 @@
 
 
 class DatasetJSONReader(DataReaderInterface):
+    def __init__(self, dataset_implementation=PandasDataset, encoding: str = None):
+        self.dataset_implementation = dataset_implementation
+        self.encoding = encoding
+
```
gerrycampion: Remove this since it will be in the `DataReaderInterface`

cdisc_rules_engine/services/data_readers/dataset_ndjson_reader.py
```diff
@@ -16,16 +16,30 @@
 
 
 class DatasetNDJSONReader(DataReaderInterface):
+    def __init__(self, dataset_implementation, encoding: str = None):
+        self.dataset_implementation = dataset_implementation
+        self.encoding = encoding
```
gerrycampion: Remove this since it will be in the `DataReaderInterface`

cdisc_rules_engine/services/data_readers/dataset_ndjson_reader.py
```diff
@@ -16,16 +16,30 @@
...
+
+    @property
+    def _encoding(self):
+        return self.encoding or "utf-8"
```
gerrycampion: remove this since the default should be set in `core.py` and passed from the factory to the `DataReaderInterface`

cdisc_rules_engine/services/data_readers/json_reader.py
```diff
@@ -6,11 +6,18 @@
 
 
 class JSONReader(DataReaderInterface):
-    def from_file(self, file_path):
+    def from_file(self, file_path, encoding: str = None):
         try:
-            with open(file_path, "rb") as fp:
-                json = load(fp)
-            return json
+            encoding = encoding or "utf-8"
```
gerrycampion: remove this since the default should be set in `core.py` and passed from the factory to the `DataReaderInterface`

cdisc_rules_engine/services/data_readers/xpt_reader.py
```diff
@@ -10,18 +10,29 @@
 
 
 class XPTReader(DataReaderInterface):
+    def __init__(self, dataset_implementation, encoding: str = None):
+        self.dataset_implementation = dataset_implementation
+        self.encoding = encoding
+
+    @property
+    def _encoding(self):
+        return self.encoding or "utf-8"
```
gerrycampion: remove this since the default should be set in `core.py` and passed from the factory to the `DataReaderInterface`

cdisc_rules_engine/services/data_readers/xpt_reader.py
```diff
@@ -10,18 +10,29 @@
...
+        data = pd.read_sas(file_path, format="xport", encoding=self._encoding)
         return PandasDataset(self._format_floats(data))
 
+    def _read_xpt_with_encoding(self, file_path: str, chunksize: int = None):
+        return pd.read_sas(file_path, chunksize=chunksize, encoding=self._encoding)
+
     def to_parquet(self, file_path: str) -> str:
         temp_file = tempfile.NamedTemporaryFile(delete=False, suffix=".parquet")
-        dataset = pd.read_sas(file_path, chunksize=20000, encoding="utf-8")
+        dataset = self._read_xpt_with_encoding(file_path, chunksize=20000)
```
gerrycampion: there is no need to add a new method here. just update the call to read_sas

core.py
```diff
@@ -346,6 +360,18 @@ def _validate_no_arguments(logger) -> None:
         "If true, limits reported issues per dataset per rule."
     ),
 )
+@click.option(
+    "-e",
+    "--encoding",
+    default=None,
```
gerrycampion: The help says this defaults to utf-8, so add the default "utf-8" here and remove it from the hardcoded locations in the rest of the code. Ensure there is also a default encoding value of utf-8 when core is called from the rule tester. maybe the datareaderfactory or datareaderinterface should have the default encoding, but we shouldn't have them at the data reader subclass level.

core.py
```diff
@@ -346,6 +360,18 @@ def _validate_no_arguments(logger) -> None:
...
+    required=False,
+    callback=validate_encoding,
+    help=(
+        "File encoding for reading datasets. "
+        "If not specified, defaults to UTF-8. "
```
gerrycampion: "UTF-8" or "utf-8"? I think either is valid, but it's good to be consistent.

README.md
```diff
@@ -148,6 +148,7 @@ Run `python core.py validate --help` to see the list of validation options.
                                   "[████████████████████████████--------]
                                   78%"is printed.
   -jcf, --jsonata-custom-functions Pair containing a variable name and a Path to directory containing a set of custom JSONata functions. Can be specified multiple times
+  -e, --encoding TEXT            File encoding for reading datasets. If not specified, defaults to UTF-8. Supported encodings: utf-8, utf-16, utf-32, cp1252, latin-1, etc.
```
gerrycampion: consistent capitalization again

tests/unit/test_dataset_json_reader.py
gerrycampion, line 1: would it be a lot of work to add tests for failing gracefully with the utf-8 encoding and passing with at least one other encoding?

cdisc_rules_engine/services/data_readers/dataset_json_reader.py
```diff
@@ -15,14 +15,18 @@
...
     def get_schema(self) -> dict:
         schema = JSONReader().from_file(
             os.path.join("resources", "schema", "dataset.schema.json")
         )
         return schema
 
     def read_json_file(self, file_path: str) -> dict:
-        return JSONReader().from_file(file_path)
+        return JSONReader().from_file(file_path, encoding=self.encoding)
```
gerrycampion: JSONReader should also be modified so that it uses the encoding from the interface. encoding here and other calls to from_file will need to be passed to the constructor, not the from_file

gerrycampion (changes requested):

cdisc_rules_engine/services/data_readers/dataset_json_reader.py
```diff
@@ -15,14 +15,15 @@
 
 
 class DatasetJSONReader(DataReaderInterface):
+
     def get_schema(self) -> dict:
-        schema = JSONReader().from_file(
+        schema = JSONReader(encoding=self.encoding).from_file(
```
RamilCDISC: These schema files are in our fixed resources folder. The encoding is fixed for them. If a user passes an encoding that is different than the encoding of the schema files in the repository. This may cause an error.

cdisc_rules_engine/services/data_readers/dataset_ndjson_reader.py
```diff
@@ -16,16 +16,23 @@
 
 
 class DatasetNDJSONReader(DataReaderInterface):
+
     def get_schema(self) -> dict:
-        schema = JSONReader().from_file(
+        schema = JSONReader(encoding=self.encoding).from_file(
```
RamilCDISC: same as above.

RamilCDISC (changes requested):

tests/unit/test_dataset_json_reader.py
RamilCDISC, line 1: I think we should at least have one test for each encoding mentioned in the readme file. Other than all looks good to me. @gerrycampion please let us know if your requested changes were satisfied and then i will add closing comments.

cdisc_rules_engine/interfaces/data_reader_interface.py
```diff
@@ -6,11 +6,13 @@ class DataReaderInterface:
     Interface for reading binary data from different file typs into pandas dataframes
     """
 
-    def __init__(self, dataset_implementation=PandasDataset):
+    def __init__(self, dataset_implementation=PandasDataset, encoding: str = "utf-8"):
```
gerrycampion: I think most instances of "utf-8" should come from the same constant, like `DEFAULT_ENCODING`, which will have "utf-8" as its value. This includes the references in core.py and the helper strings.

cdisc_rules_engine/services/data_services/local_data_service.py
```diff
@@ -53,13 +55,17 @@ def get_instance(
         config: ConfigInterface = None,
         **kwargs,
     ):
-        if cls._instance is None:
+        encoding = kwargs.get("encoding")
+        if cls._instance is None or (
+            encoding is not None and cls._instance.encoding != encoding
```
gerrycampion, line 65: how would this happen?

cdisc_rules_engine/services/data_services/dummy_data_service.py
```diff
@@ -177,17 +182,17 @@ def get_datasets(self) -> Iterable[SDTMDatasetMetadata]:
         return self.data
 
     @staticmethod
-    def get_data(dataset_paths: Sequence[str]):
-        json = JSONReader().from_file(dataset_paths[0])
+    def get_data(dataset_paths: Sequence[str], encoding: str = None):
```
gerrycampion: why are we using default value of None in some places like here, but utf-8 in others?
And do we have a way to get the cli's encoding in these staticmethods?

cdisc_rules_engine/services/datasetndjson_metadata_reader.py
```diff
@@ -15,23 +15,34 @@ class DatasetNDJSONMetadataReader:
...
+    def __init__(self, file_path: str, file_name: str, encoding: str = None):
         self._metadata_container = {}
         self._file_path = file_path
         self._first_record = None
         self._dataset_name = file_name.split(".")[0].upper()
+        self.encoding = encoding
+
+    @property
+    def _encoding(self):
+        return self.encoding or "utf-8"
```
gerrycampion: why is this inconsistent with the other readers?

cdisc_rules_engine/services/datasetxpt_metadata_reader.py
```diff
@@ -199,3 +198,22 @@ def _extract_adam_info(self, variable_names):
             "selection_algorithm": ad.selection_algorithm,
         }
         return adam_info_dict
+
+    @property
+    def _encoding(self):
+        return self.encoding or "utf-8"
```
gerrycampion: again, this seems inconsistent

cdisc_rules_engine/services/datasetxpt_metadata_reader.py
```diff
@@ -199,3 +198,22 @@ def _extract_adam_info(self, variable_names):
...
+
+    def _read_xport(self, **kwargs):
+        """Read XPT file with encoding, fallback to no encoding if TypeError."""
+        try:
+            return pyreadstat.read_xport(
+                self._file_path, encoding=self._encoding, **kwargs
+            )
+        except TypeError:
+            return pyreadstat.read_xport(self._file_path, **kwargs)
```
gerrycampion: Why are we doing a fallback in this reader, but not the others? Do we want this fallback? I'm not sure the data will be reliable if the encoding is different from the specified encoding.

tests/unit/test_dataset_ndjson_reader.py
```diff
@@ -10,7 +11,7 @@ def test_from_file():
         f"{os.path.dirname(__file__)}/../resources/test_dataset.ndjson"
     )
 
-    reader = DatasetNDJSONReader()
+    reader = DatasetNDJSONReader(PandasDataset)
```
gerrycampion: why did this need to be added? Isn't it already the default for the interface?

tests/unit/test_xpt_reader.py
```diff
@@ -10,7 +11,7 @@ def test_read():
     with open(test_dataset_path, "rb") as f:
         data = f.read()
 
-    reader = XPTReader()
+    reader = XPTReader(PandasDataset)
```
gerrycampion: same comment

cdisc_rules_engine/services/data_services/local_data_service.py
RakeshBobba03: When validation is run more than once in the same process with different --encoding values (e.g. first run with default, second with -e cp1252). Without this check, the singleton would keep the first run’s encoding and ignore the second. So the condition is there to reset the singleton when the requested encoding changes.

cdisc_rules_engine/interfaces/data_reader_interface.py
RakeshBobba03: Done. The interface now uses encoding: str = DEFAULT_ENCODING, and core.py uses DEFAULT_ENCODING for the default and help text.

cdisc_rules_engine/services/data_services/local_data_service.py
RakeshBobba03: Added a docstring on get_instance that describes this.

cdisc_rules_engine/services/data_services/dummy_data_service.py
RakeshBobba03: get_data and is_valid_data now use encoding: str = DEFAULT_ENCODING and encoding or DEFAULT_ENCODING when calling JSONReader. CLI encoding is passed via DataServiceFactory(encoding=args.encoding) → get_data(..., encoding=self.encoding).

cdisc_rules_engine/services/datasetndjson_metadata_reader.py
RakeshBobba03: The _encoding property was removed; we use self.encoding or DEFAULT_ENCODING inline, same as the other readers.

cdisc_rules_engine/services/datasetxpt_metadata_reader.py
RakeshBobba03: Same as above: _encoding removed; we use self.encoding or DEFAULT_ENCODING in _read_xport.

cdisc_rules_engine/services/datasetxpt_metadata_reader.py
RakeshBobba03: Agreed. The fallback was removed so we always pass the chosen encoding to pyreadstat.read_xport. Our minimum pyreadstat version supports encoding, and falling back could produce wrong data.

tests/unit/test_dataset_ndjson_reader.py
RakeshBobba03: Reverted. Kept DatasetNDJSONReader() and rely on the interface default.

tests/unit/test_xpt_reader.py
RakeshBobba03: Reverted. Kept XPTReader() and rely on the default.

tests/unit/test_dataset_json_reader.py
RakeshBobba03: Encoding tests already covered utf-8 (via the “wrong encoding” test) and cp1252. I added explicit tests for utf-16, utf-32, and latin-1 as a single parametrized test, so every encoding listed in the README (utf-8, utf-16, utf-32, cp1252, latin-1) now has at least one test.

cdisc_rules_engine/services/data_readers/dataset_json_reader.py
```diff
@@ -12,12 +12,13 @@
...
 
 from cdisc_rules_engine.services.data_readers.json_reader import JSONReader
+from cdisc_rules_engine.constants import DEFAULT_ENCODING
 
 
 class DatasetJSONReader(DataReaderInterface):
 
     def get_schema(self) -> dict:
-        schema = JSONReader(encoding="utf-8").from_file(
+        schema = JSONReader(encoding=DEFAULT_ENCODING).from_file(
```
gerrycampion: I know this may seem counterintuitive, but the schema loading here should actually be hardcoded independently to "utf-8", instead of being tied to our global const. Refer to Ramil's previous comment:
> These schema files are in our fixed resources folder. The encoding is fixed for them.

cdisc_rules_engine/services/data_readers/dataset_ndjson_reader.py
```diff
@@ -13,12 +13,13 @@
...
 
 from cdisc_rules_engine.services.data_readers.json_reader import JSONReader
+from cdisc_rules_engine.constants import DEFAULT_ENCODING
 
 
 class DatasetNDJSONReader(DataReaderInterface):
 
     def get_schema(self) -> dict:
-        schema = JSONReader(encoding="utf-8").from_file(
+        schema = JSONReader(encoding=DEFAULT_ENCODING).from_file(
```
gerrycampion: Same thing here....
I know this may seem counterintuitive, but the schema loading here should actually be hardcoded independently to "utf-8", instead of being tied to our global const. Refer to Ramil's previous comment:
> These schema files are in our fixed resources folder. The encoding is fixed for them.

cdisc_rules_engine/services/datasetxpt_metadata_reader.py
```diff
@@ -199,3 +201,14 @@ def _extract_adam_info(self, variable_names):
...
+    def _read_xport(self, **kwargs):
+        """Read XPT file using the configured encoding."""
+        encoding = self.encoding or DEFAULT_ENCODING
+        return pyreadstat.read_xport(self._file_path, encoding=encoding, **kwargs)
+
+    def _read_xport_with_encoding(self):
+        return self._read_xport(row_limit=self.row_limit)
+
+    def _read_xport_with_encoding_metadata_only(self):
+        return self._read_xport(metadataonly=True)
```
gerrycampion: I think it's overkill to have these 3 functions now, when we can do this inline with the calls.

gerrycampion (commented):
just a few more minor comments

gerrycampion (approved):

PR #1460, opened 2025-12-02
update all commands and docs

SFJohnson24:
This PR updates all the actions, readme for the jsonata in resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ Open source offering of the CDISC Rules Engine, a tool designed for validating c
 
 ## Quick start
 
+**Need help?** Jump to [Troubleshooting & Support](#troubleshooting--support)
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build_version.yml
```diff
@@ -134,6 +134,7 @@ jobs:
 #             --add-data="resources/templates:resources/templates" \
 #             --add-data="resources/schema:resources/schema" \
 #             --add-data="tests/resources/datasets:tests/resources/datasets"
```
ASL-rmarshall: Does this need a trailing \ if it's no longer the last `--add-data`?

.github/workflows/build_version.yml
```diff
@@ -37,7 +37,7 @@ jobs:
             --add-data="resources/cache:resources/cache" \
             --add-data="resources/templates:resources/templates" \
             --add-data="resources/schema:resources/schema" \
-            --add-data="tests/resources/datasets:tests/resources/datasets"
```
ASL-rmarshall, line 40: Is removal of test/resources/datasets deliberate?  It hasn't been removed from other places.
If it's retained, I assume a trailing \ will need to be added after it.

SFJohnson24: I think that might have been copilot--some of its suggestions are quite aggressive.  I am going to work on disabling it

PR #1461, opened 2025-12-02
new template, changed filename origin

SFJohnson24:
this PR updates the template with Nick's formatted version.  It also changes the dataset in dataset details to the name property over the filename as was requested in rule author meeting

### DIFF
--- a/host.json
+++ b/host.json
@@ -2,7 +2,7 @@
   "version": "2.0",
   "extensionBundle": {
     "id": "Microsoft.Azure.Functions.ExtensionBundle",
-    "version": "[2.*, 3.0.0)"
+    "version": "[4.0.0, 5.0.0)"
   },
   "functionTimeout": "18:00:00",
   "logger": {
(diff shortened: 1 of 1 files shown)


RamilCDISC (commented):
I am not sure if it this will fall under current ticket but because the new template was added in the connected issue. I observed that the conformance details tab in the report template shows none for Issue Limit per Rule. But the engine put 0 when limit is not specified in command line.

Please let me know if this will be fixed in a separate ticket or we can update it in the current PR?

SFJohnson24:
@RamilCDISC I updated the env for MAX_ERRORS_PER_RULE, updated our example and moved the parsing logic to the utils (as scripts are not put into our pypi package).  This should resolve the conformance details tab issue you noticed.

RamilCDISC (approved):
The PR updates the reporting to match the updated template. The changes were validated by: 


1. Reviewing the PR code for any unwanted code or comments.
3. Ensuring the changes are consistent with previous code.
4. Ensuring all unit and regression tests pass.
5. Ensuring all related testing is updated.
6. Running manual test locally and comparing the excel report with the template.
7. Running manual tests locally and ensuring json report is also updated accordingly.
8. Validating the Issue limit per dataset and per rule keys for default values.
9. Validating the issue limit per dataset and per rule keys for values set through CLI and env variables.

PR #1469, opened 2025-12-04
1017 ddf00102

RakeshBobba03:
Fixes Issue #1017  (CORERULES-9645): Updated `is_column_of_iterables` to check all non-null values instead of only the first row. When the first row is `None` (common after grouped distinct operations with LEFT joins), the method now correctly identifies columns containing iterables by examining all non-null values. Also fixed boolean column reading in Excel files by using nullable boolean dtype (`"boolean"`) and adding `true_values`/`false_values` parameters to handle `NaN` values. This fix ensures containment operators work correctly with columns containing `None` values and affects USDM rules using grouped distinct operations with containment operators.

### DIFF
--- a/cdisc_rules_engine/check_operators/dataframe_operators.py
+++ b/cdisc_rules_engine/check_operators/dataframe_operators.py
@@ -155,16 +155,7 @@ class DataframeType(BaseType):
     def is_column_of_iterables(self, column):
         if not self.value.is_series(column):
             return False
-        non_null_values = []
-        for val in column:
-            if val is None:
-                continue
-            try:
-                if pd.isna(val):
-                    continue
-            except (ValueError, TypeError):
-                pass
-            non_null_values.append(val)
+        non_null_values = column[column.notna()]
         return len(non_null_values) > 0 and all(
             isinstance(val, (list, set)) for val in non_null_values
         )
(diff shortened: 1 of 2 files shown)


cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -153,11 +153,20 @@ def get_comparator_data(self, comparator, value_is_literal: bool = False):
...
+        non_null_values = []
+        for val in column:
+            if val is None:
+                continue
+            try:
+                if pd.isna(val):
+                    continue
+            except (ValueError, TypeError):
+                pass
+            non_null_values.append(val)
```
gerrycampion: i believe you can replace this block with: 
`non_null_values = column[column.notnull()]`

it will be useful to take a tutorial to learn some pandas fundamentals

tests/unit/test_check_operators/test_containment_checks.py
```diff
@@ -511,3 +511,68 @@ def test_is_contained_by_with_none_in_comparison(
...
+    """Test pd.NA handling (pandas NA value)"""
+    import pandas as pd
+
+    test_cases = [
+        ([pd.NA], False),
+        ([None, pd.NA], False),
+        ([None, pd.NA, ["A"]], True),
+        ([pd.NA, ["A"]], True),
+        ([None, pd.NA, ["A"], ["B"]], True),
+    ]
```
gerrycampion: Sorry, one more thing. I think these test cases should be combined to a single function

ASL-rmarshall (approved):
General code review looks good.  Validation produced expected results.

gerrycampion (approved):

PR #1454, opened 2025-11-25
CG0562

SFJohnson24:
This PR resolves the issues currrently reported with CG0562.  Mainly: the authors need to filter only the date portion and not datetime for the record count.  I added regex handling to this operator to support this and the rule below can be found that implements this.  I also added logic for grouping based on a wildcard column to appropriately process the -- in the column name
[Datasets.json](https://github.com/user-attachments/files/23866748/Datasets.json)
[Datasets.xlsx](https://github.com/user-attachments/files/23866745/Datasets.xlsx)
[Rule_underscores.json](https://github.com/user-attachments/files/23866747/Rule_underscores.json)
[CORE-Report-2025-12-01T15-16-44.xlsx](https://github.com/user-attachments/files/23866744/CORE-Report-2025-12-01T15-16-44.xlsx)

This pull request adds support for applying regex transformations to grouping columns in the `record_count` operation, allowing users to group records based on extracted patterns (such as dates from datetime strings). The changes include updates to the operation logic, schema, documentation, and comprehensive unit tests to ensure correct behavior for regex-based grouping, including support for grouping aliases and filters.

**Record Count Operation Enhancements**

* Added support for a new `regex` parameter in the `record_count` operation, enabling transformation of grouping column values using a regex pattern before grouping. This allows, for example, grouping by just the date portion of a datetime string. (`cdisc_rules_engine/operations/record_count.py`, `cdisc_rules_engine/models/operation_params.py`, `cdisc_rules_engine/utilities/rule_processor.py`, `resources/schema/Operations.json`) [[1]](diffhunk://#diff-632d76e6b0ece29921b92a1f8d4f7e09edf5cb0941167b7200861038de99650cR26-R62) [[2]](diffhunk://#diff-632d76e6b0ece29921b92a1f8d4f7e09edf5cb0941167b7200861038de99650cR74-R146) [[3]](diffhunk://#diff-a8c3ea5072f0809ba706bc8b5ca1aa810d71a56b0bbb88bb49fc86cf6db89bb2R57) [[4]](diffhunk://#diff-7012335a387195f481fe543f82ef6df8ba075c2ef5ae09c1fbf0538c7c96e936R427) [[5]](diffhunk://#diff-a78091fe1253d28a9dd06546828bfb38d3106ea3315064c5187ffda66362ca4dR546-R548)

[...]

PR #1472, opened 2025-12-08
record count optimize

SFJohnson24:
in reviewing https://github.com/cdisc-org/cdisc-rules-engine/pull/1454 changes for sprint review--i noticed an optimization.  Instead of copying the frame and adding index to merge back to the original from the grouped, we can rely on the indexes in values().  This adds a memory optimization while maintaining the functionality.

This pull request refines the logic for handling regex-based grouping and counting in the `record_count.py` operation. The main improvements are focused on ensuring that regex grouping is only applied when necessary and simplifying the process of merging grouped counts back to the original dataframe.

Logic improvements for regex grouping:

* Updated the condition in `_execute_operation` to only perform regex grouping when a regex is provided and no pre-filtering has occurred, preventing unnecessary operations.

Simplification and efficiency in merging grouped counts:

* Refactored `_get_regex_grouped_counts` to streamline the process of merging grouped counts back to the original dataframe, removing the use of auxiliary `_idx` columns and simplifying the assignment of the `"size"` column.


cdisc_rules_engine/operations/record_count.py
```diff
@@ -102,19 +102,12 @@ def _get_regex_grouped_counts(self, dataframe, grouping_columns):
...
         )
-        original_with_idx["size"] = transformed_with_counts["size"].values
-        result = (
-            original_with_idx.drop(columns=["_idx"])
-            .groupby(grouping_columns, as_index=False, dropna=False)
-            .first()
-        )
+        result = dataframe[grouping_columns].copy()
+        result["size"] = transformed_with_counts["size"].values
+        result = result.groupby(grouping_columns, as_index=False, dropna=False).first()
```
RamilCDISC, line 110: The new logic breaks row alignments and i snot equivalent to the previous logic. This can cause silent bugs. Could you please mention if you think this is correct and that regex transformations will not reorder rows?

cdisc_rules_engine/operations/record_count.py
```diff
@@ -28,7 +28,7 @@ def _execute_operation(self) -> pd.Series:
                 if self.params.regex
                 else effective_grouping
             )
-            if self.params.regex:
+            if self.params.regex and not filtered:
```
RamilCDISC: filtered here would be either none or a dataframe. Using not on a dataframe raises ```
ValueError(
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```
This would need a different logic for handling here

cdisc_rules_engine/operations/record_count.py
SFJohnson24: @RamilCDISC why do you believe it will break row alignment?  df_for_grouping is transformed by the apply_regex_to_grouping_columns but this performs only element-wise operations on selected columns.  No rows are removed/added/reordered by this.  The left merge keep all rows in the original order.  I dont know where the row alignment not changing would be introduced

cdisc_rules_engine/operations/record_count.py
SFJohnson24: I corrected this--changed it to an explicit None check as I want to avoid doing regex on the original frame if filtering has taken place and the regex will be applied to it lower down

cdisc_rules_engine/operations/record_count.py
RamilCDISC: Sorry I actually wanted to tag line 105. The merge operation can change the ordering. The previous code had 'idx' which could be used for reorder to original. This may not be a big issue. Please let me know your thoughts.

SFJohnson24: @RamilCDISC 
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.merge.html
"left: use only keys from left frame, similar to a SQL left outer join; **preserve key order** "

It is a left merge so the key order is preserved.

PR #1482, opened 2025-12-15
1442: add filetype option to validate type of data

alexfurmenkov:


core.py
```diff
@@ -175,6 +184,12 @@ def _validate_no_arguments(logger) -> None:
     required=False,
     help=f"Path to directory containing data files ({VALIDATION_FORMATS_MESSAGE})",
 )
+@click.option(
+    "-filetype",
```
RamilCDISC: Could you please make this similar to other flags. full name with '--' and a short with '-'. You can use '-f' and '--filetype'

RamilCDISC (approved):
The PR adds a few flag that lets you control which files engine works on in the selected dataset folder. Not specifying this flag results in previous behaviour of engine. 

The validation was done by: 

1. Reviewing the PR for any unwanted code or logic. 
4. Ensuring all related testing is updated and new tests are added. 
5. Running engine locally for manual validation.
6. Verifying the edge case of empty directory. 
7. Verifying the edge case of multiple -ft flags.
8. Verifying edge case of invalid file type.

mhungria:
@alexfurmenkov , @RamilCDISC , thanks for adding the --filetype flag.

I'm testing the last release v0.14.1 on a linux box and so far it works fine.

Issue #1413, opened 2025-10-30
Allow update-cache to update from unauthenticated endpoints without an api key


gerrycampion:
Specifically, this is meant to support the [Membership Library 2026](https://github.com/orgs/cdisc-org/projects/16).
We need to continue to allow free access to Rules and will likely do this by exposing the rules api endpoint without the need of an api-ke. 

In the engine, we will need to allow update-cache to run without an api key and fetch data from any endpoints that don't require an api-key.  An information message should be printed to describe this. If any api-key is provided, the full update will proceed as always.

Acceptance Criteria

- Update cache runs without an API key
- Unauthorize request are log
- If an API key is provided the full update will proceed
- An information message is printed when running the restricted update

JozefAerts:
Hi @gerrycampion Is this the CDISC decision?
There has also been the discussion with Chris and others to e.g. limit the bandwith of monthly usage of the API for non-members, but these would still need to register for the API-key. If that is what will be done, I don't think we need to change anything here.

gerrycampion:
@JozefAerts We don't have a final decision on this detail yet. This is a placeholder for potential work.

PR #1484, opened 2025-12-15
Allow custom CDISC Library API URL (proxy / local cache support)

tomhub:
This PR adds support for configuring a custom CDISC Library API URL, enabling the use of local or private proxy/cache services (e.g. basic example https://github.com/tomhub/cdisc-proxy). This helps reduce load on the upstream CDISC Library API and limits repeated API calls in shared or CI environments.

**Key changes**
* Added --apiurl CLI option and CDISC_LIBRARY_API_URL environment variable.
* --apikey / CDISC_LIBRARY_API_KEY is now optional when using a custom API URL.
* Validation enforces that at least one of API key or custom API URL is provided:
* API key is required when using the default CDISC Library endpoint.
* Custom/proxy URLs may omit the API key if the proxy handles authentication.
* Updated cache update flow, config service, and data services to pass the API URL consistently.
* Improved README and .env example with proxy usage guidance.
* Added unit tests covering default vs custom URL validation logic.

Why this matters
* Enables local caching and proxying of CDISC Library requests.
* Reduces API throttling and repeated upstream calls.
* Improves usability in regulated, offline, or enterprise environments.

This should partially address issue #1413

NOTE: I think [tests/unit/test_cdisc_library_service.py] needs to be properly fixed.


cdisc_rules_engine/services/cdisc_library_service.py
```diff
@@ -16,13 +16,25 @@
...
+        
+        if not api_url:  # covers None and empty string
+            api_url = "https://api.library.cdisc.org/api"
+        
+        # Validation: If using the default CDISC Library URL, API key is required
+        if api_url == "https://api.library.cdisc.org/api" and not self._api_key:
+            raise ValueError(
+                "CDISC_LIBRARY_API_KEY is required when using the default CDISC Library API URL. "
+                "Either provide an API key or specify a custom CDISC_LIBRARY_API_URL which does not require an API key."
+            )
```
gerrycampion, line 30: This will go away with #1506 because a missing library api key will still be able to fetch the rules

tests/unit/test_cdisc_library_service.py
```diff
@@ -5,12 +5,34 @@
...
 from unittest.mock import MagicMock, patch
 
 from cdisc_rules_engine.config import config
 from cdisc_rules_engine.services.cdisc_library_service import CDISCLibraryService
 
 
+def test_library_service_requires_key_for_default_url():
+    """Test that API key is required when using default CDISC Library URL."""
+    with pytest.raises(ValueError, match="CDISC_LIBRARY_API_KEY is required"):
+        CDISCLibraryService("", "", MagicMock())
```
gerrycampion, line 18: This can be removed

core.py
```diff
@@ -562,15 +574,29 @@ def update_cache(
...
     custom_rule: str,
     remove_custom_rules: str,
     update_custom_rule: str,
     custom_standard: str,
     remove_custom_standard: str,
 ):
+    logger = logging.getLogger("validator")
+    
+    # Validation: Ensure at least one is provided when using default URL
+    effective_url = apiurl if apiurl else "https://library.cdisc.org/api"
```
gerrycampion, line 588: can this go into the click "default" value instead?

core.py
```diff
@@ -562,15 +574,29 @@ def update_cache(
...
+    if effective_url == "https://library.cdisc.org/api" and not apikey:
+        logger.error(
+            "CDISC_LIBRARY_API_KEY is required when using the default CDISC Library API URL.\n"
+            "Either provide --apikey or set CDISC_LIBRARY_API_KEY environment variable,\n"
+            "or specify a custom URL with --apiurl or CDISC_LIBRARY_API_URL environment variable,"
+            "which does or does not require specific API key for proxy access"
+        )
+        ctx.exit(2)
```
gerrycampion, line 596: this can be removed

env.example
```diff
@@ -1,4 +1,5 @@
 CDISC_LIBRARY_API_KEY=your_api_key_here
+CDISC_LIBRARY_API_URL=http://localhost:31415/api # smart proxy server will not use CDISC_LIBRARY_API_KEY, but will might need it's own key to query upstream server.
```
gerrycampion, line 2: typo "but will might need"

gerrycampion (changes requested):
primarily feedback where we will see merge conflicts with #1506 which will allow users to retrieve latest rules with the default library api, but no api key

gerrycampion:
@tomhub are you still working on this PR or do you think we should close this as it is no longer relevant?

PR #1490, opened 2025-12-16
manager shutdown

SFJohnson24:
This pull request refactors the `run_validation` function in `scripts/run_validation.py` to improve resource management and code readability. The main change is ensuring that the `CacheManager` is properly shut down after validation, even if an exception occurs, by wrapping the main logic in a `try/finally` block. Additionally, several lines are reformatted for better readability.

Resource management improvements:
* Wrapped the main logic of `run_validation` in a `try/finally` block to guarantee that `manager.shutdown()` is called, ensuring proper cleanup of resources. [[1]](diffhunk://#diff-4a7a926a43d976bc2fc0cfb78839df350e02f9e6191061859727d48c01101f66R131-R137) [[2]](diffhunk://#diff-4a7a926a43d976bc2fc0cfb78839df350e02f9e6191061859727d48c01101f66R209-R210)

### DIFF
--- a/scripts/run_validation.py
+++ b/scripts/run_validation.py
@@ -202,11 +202,15 @@ def run_validation(args: Validation_args):
         for reporting_service in reporting_services:
             reporting_service.write_report()
         print(f"Output: {args.output}")
-        engine_logger.info(" Report generated, Cleaning up intermediate files")
-        for file in created_files:
-            engine_logger.info(f"Deleting file {file}")
-            os.remove(file)
     finally:
+        if created_files:
+            engine_logger.info(" Report generated, Cleaning up intermediate files")
+            for file in created_files:
+                try:
+                    engine_logger.info(f"Deleting file {file}")
+                    os.remove(file)
+                except Exception as e:
+                    engine_logger.warning(f"Failed to delete {file}: {e}")
         manager.shutdown()
 
 


scripts/run_validation.py
```diff
@@ -128,77 +128,86 @@ def run_validation(args: Validation_args):
...
+            datasets, results, elapsed_time, args, data_service, dictionary_versions
+        )
+        reporting_services: List[BaseReport] = reporting_factory.get_report_services()
+        for reporting_service in reporting_services:
+            reporting_service.write_report()
+        print(f"Output: {args.output}")
+        engine_logger.info(" Report generated, Cleaning up intermediate files")
+        for file in created_files:
+            engine_logger.info(f"Deleting file {file}")
+            os.remove(file)
```
RamilCDISC: I think we should keep file parquet file deleting in final block because any exception that will occur before this part will leave the files undeleted.

SFJohnson24: moved, i also added it to a try/except to warn user if a file fails to delete

RamilCDISC (approved):
The PR refactors the run validation code to ensure that the pool processes are closed properly once the validation completes. The Validation was done by:

2. Ensuring all unit and regression testing pass.
3. Ensuring the refactored code fully covers the previous code. 
4. Running manual validations using CLI to ensure proper execution.
5. Validation using large datasets to ensure compatibility.

Issue #1274, opened 2025-07-29
Comparison in Reporting


SFJohnson24:
currently, for example with  CORE-000334. the output variables gives the two columns
<img width="1097" height="239" alt="Image" src="https://github.com/user-attachments/assets/91ea3366-d852-4a53-8b25-5c278bcde49f" />
it would be convenient to have some type of flag in the rule within the output variables section of the rule where it signals the reporting service to compare the two output variables.

It would need to process various different structures like objects nested in lists, lists, etc as we cannot expect the comparison to always be two lists.  The datastructures would need to be the same though for comparison.

mhungria:
Related to issue https://github.com/cdisc-org/cdisc-rules-engine/issues/1091.

RakeshBobba03:
From what Gerry suggested, I was thinking to do it this way:

Implement a new `minus` operation (in the Operations section, not a check operator) that computes set difference A ∖ B. The `minus` operation would:

- Use `name` and `value` to reference the two input variables
- Return elements in `name` that are not in `value`
- Use the same logic that currently produces "Extra in $expected_variables" in the reporting layer, moved into an Operation

For CORE-000334, the rule would look like:

```
Operations:
  - id: $expected_variables
    operator: expected_variables
  - id: $dataset_variables
    operator: get_column_order_from_dataset
  - id: $expected_minus_dataset
    name: $expected_variables
    operator: minus
    value: $dataset_variables
Check:
  all:
    - name: $expected_minus_dataset
      operator: non_empty
Outcome:
  Message: At least one expected variable is missing from dataset
  Output Variables:
    - $dataset_variables
    - $expected_variables
    - $expected_minus_dataset
```

The output in `$expected_minus_dataset` would match the original "Extra in $expected_variables" (variables expected but missing from the dataset). Edge cases: missing/empty/incorrect types for `name` and `value`.

@SFJohnson24  @mhungria  – does this approach look right before I implement it?

SFJohnson24:
seems good, I dont know if value being empty is an edge case.  listA - empty would just report all of listA as it is all missing from an empty value.   I would make sure to look at base_operation.py and see how various data types are handled.

mhungria:
Hi @RakeshBobba03 , @SFJohnson24 , I assume it is general 'minus' operator on two lists of objects of same datatype (lista - listb), right? I would consider preserving the order of items in lista in the resulting list.
Could it be generic including accepting lists with duplicate values?

PR #1491, opened 2025-12-16
#1274: Add inline `compared` syntax for N-way set-based output variable comparison

RakeshBobba03:
#1274: Implements explicit inline comparison syntax for output variables in rules via a new compared: dict block within Output Variables, enabling N-way set-based comparisons (len >= 2) where the first variable serves as baseline and subsequent variables are compared against it. The implementation flattens all variables (siblings + compared children) for UI display while isolating comparison logic to only variables within compared blocks, always uses set-based (order-independent) comparison. Reporting now shows formatted comparison summaries (missing/extra items) followed by raw variable lists, with Excel multi-line rendering support.

Attached are the Rule and dataset used for testing:

[CORE-000334.yaml](https://github.com/user-attachments/files/24203782/CORE-000334.yaml)
[unit-test-coreid-CG0016-negative.xlsx](https://github.com/user-attachments/files/24203783/unit-test-coreid-CG0016-negative.xlsx)


RakeshBobba03:
> I ran a validation using dev editor and updated the rule to have USUBJID also in output variables

Thanks for bringing this to my attention @RamilCDISC 

USUBJID is record-level data, not metadata. When it's included in Output Variables for a Variable Metadata Check rule, the rule operates on a metadata dataset, not the original data records. Since USUBJID is a data column and not part of the metadata structure, it doesn't exist in the metadata dataset. That's why it was showing "Not in dataset".

Sam and I agreed that the message "Not in dataset" is misleading in this context, since USUBJID actually exists in the original dataset, it's just not available in the metadata context for Variable Metadata Check rules. We decided to change the error message to "not available in metadata context" for Variable Metadata Check rules and Dataset Metadata Check rules (and their variants) to better reflect what's happening.

I updated the error message logic in `actions.py` to return "not available in metadata context" for metadata check rule types instead of "Not in dataset".

and

I updated the documentation in `Rule_Type.md` to clarify that output variables must match their respective rule types, and that for Variable Metadata Check and Dataset Metadata Check rules, variables not available in the metadata context will display "not available in metadata context" instead of the variable value.

resources/schema/Rule_Type.md
gerrycampion: This has nothing to do with Rule Types. It should be

cdisc_rules_engine/services/reporting/sdtm_report_data.py
gerrycampion: should most of this be done in the base_report_data instead of sdtm_report_data?

resources/schema/CORE-base.json
```diff
@@ -467,7 +467,25 @@
         },
         "Output Variables": {
```
gerrycampion: Maybe best to move this to it's own json file since it is getting more complex and needs a more complex markdown description.

gerrycampion:
Sorry to throw this in here, but after seeing this PR and giving this some thought, I don't think this should be a change in the reporting mechanism. I think we should just have a new operation like `minus`. Then we could have a rule like:

```yaml
Operations:
  - id: $expected_variables
    operator: expected_variables
  - id: $dataset_variables
    operator: get_column_order_from_dataset
  - id: $expected_minus_dataset
    name: $expected_variables
    operator: minus
    value: $dataset_variables
Check:
  all:
    - name: $expected_minus_dataset
      operator: non_empty
Outcome:
  Message: At least one expected variable is missing from dataset
  Output Variables:
    - $expected_variables
    - $dataset_variables
    - $expected_minus_dataset
```

Would this work?

gerrycampion:
> Sorry to throw this in here, but after seeing this PR and giving this some thought, I don't think this should be a change in the reporting mechanism. I think we should just have a new operation like `minus`. Then we could have a rule like:
> 
> Would this work?

btw if we go this route, it's probably easiest to just create a new branch

PR #1494, opened 2025-12-17
hotfix

SFJohnson24:
this PR fixes the error from #1445 where merged_domains was kept track of without accounting for USDM.  We now do a check for domain and if it is not there, we use name.  As USDM does not have domain, this prevents issues with merging datasets in preprocessor
to test: 
Any of these:
   DDF00103
   DDF00093
   DDF00092
   DDF00181
   DDF00127
   DDF00094
   DDF00048
   DDF00095
image
image

you can see the type.decode.GeographicScope for DDF00093 on the right which is this branch now correctly merges and we get the 4 results that are expected from the rule and the test data.  Right is the current engine and does not merge properly

### DIFF
--- a/cdisc_rules_engine/utilities/dataset_preprocessor.py
+++ b/cdisc_rules_engine/utilities/dataset_preprocessor.py
@@ -162,7 +162,9 @@ class DatasetPreprocessor:
                         right_dataset_domain_details=domain_details,
                         datasets=datasets,
                     )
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
The PR provides a hotfix for USDM bug. It covers the case of USDM which does not have a domain by utilizing the file name instead. All unit and regression testing pass. test_suite will be further updated to cover such cases in future.

PR #1506, opened 2026-01-01
Allow updating cache without apikey

nhaydel:
This PR is apart of the membership levers project.

We want to allow users who don't know their api key or don't have a valid api key to still pull rules from the library and update the cache. This PR just updates the cache populator to allow fetching rules only as well as handle any failed requests to the cdisc library api.

The expected outcomes are:
* Running the update_cache function **without** an api key should result in a log message saying only the rules will be updated. Only the `save_rules_locally` function should be run.

> [!NOTE]
> Since the ability to request rules from the production library api without an api key has not been released you will see 401 responses from the rules endpoint as well. This should be resolved once that functionality is released. If you want to ensure the rules are updated, you can use the dev environment.

* Running the update_cache function **with** an invalid api key should result in all failed requests being logged and the cache should not get updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,7 +319,7 @@ core.exe validate -rest -of -config -commands
 
 ### Updating the Cache (`update-cache`)
 
-Update locally stored cache data (An api-key can be provided through the environment variable - `CDISC_LIBRARY_API_KEY`. This is stored in the .env folder in the root directory, the API key does not need quotations around it.) When running a validation, CORE uses rules in the cache unless -lr is specified. Running the above command populates the cache with controlled terminology, rules, metadata, etc.
+Update locally stored cache data (An api-key can be provided through the environment variable - `CDISC_LIBRARY_API_KEY`. This is stored in the .env file in the root directory, the API key does not need quotations around it.) When running a validation, CORE uses rules in the cache unless -lr is specified. Running the above command populates the cache with controlled terminology, rules, metadata, etc.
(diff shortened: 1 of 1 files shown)


core.py
gerrycampion, line 1: i think you should add documentation describing the new behavior in this section: https://github.com/cdisc-org/cdisc-rules-engine/blob/allow-update-cache-without-apikey/README.md#updating-the-cache-update-cache

README.md
```diff
@@ -267,7 +267,9 @@ Then run normally: `core.exe validate -rest -of -config -commands
 
 ### Updating the Cache (`update-cache`)
 
-Update locally stored cache data (Requires an environment variable - `CDISC_LIBRARY_API_KEY`) This is stored in the .env folder in the root directory, the API key does not need quotations around it. When running a validation, CORE uses rules in the cache unless -lr is specified. Running the above command populates the cache with controlled terminology, rules, metadata, etc.
+Update locally stored cache data (An api-key can be provided through the environment variable - `CDISC_LIBRARY_API_KEY`. This is stored in the .env folder in the root directory, the API key does not need quotations around it.) When running a validation, CORE uses rules in the cache unless -lr is specified. Running the above command populates the cache with controlled terminology, rules, metadata, etc.
```
RamilCDISC: I think it would be .env file instead of .env folder. Could you please update that?

Issue #635, opened 2024-03-13
There should be documentation for Executability


gerrycampion:
1. The rule schema should be updated to include descriptions for Executability and each of the Executability options. This should work with intellisense.
2. The rule editor author guide should be updated to include a section for Executability.

SFJohnson24:
note: the second point is updated here: https://github.com/cdisc-org/conformance-rules-editor/blob/main/docs/_sidebar.md  you will need to create a new markdown file and add it to the sidebar which is how the https://cdisc-org.github.io/conformance-rules-editor/#/ reference guide is constructed

PR #1507, opened 2026-01-01
#635 Add documentation for executability

RakeshBobba03:
Issue #635 : Add documentation for Executability
Adds Executability documentation. Updates CORE-base.json to convert the Executability field from a simple enum to an anyOf pattern with markdownDescription for each option (Fully Executable, Not Executable, Partially Executable, Partially Executable - Possible Overreporting, Partially Executable - Possible Underreporting), enabling intellisense descriptions in the rule editor. Creates Executability.md in resources/schema/ with documentation for all Executability options, including examples and notes that Executability is informational metadata. Updates conformance-rules-editor/docs/executability.md to use the include pattern, pulling content from the cdisc-rules-engine repo to match the pattern used by other schema documentation files.

### DIFF
--- a/resources/schema/CORE-base.json
+++ b/resources/schema/CORE-base.json
@@ -400,29 +400,7 @@
       "type": "string"
     },
     "Executability": {
-      "anyOf": [
-        {
-          "const": "Fully Executable",
-          "title": "The rule can be fully executed and validated automatically by the rules engine"
-        },
-        {
-          "const": "Not Executable",
-          "title": "The rule cannot be executed automatically by the rules engine"
-        },
-        {
-          "const": "Partially Executable",
-          "title": "The rule can be partially executed, but may not capture all validation scenarios"
-        },
-        {
-          "const": "Partially Executable - Possible Overreporting",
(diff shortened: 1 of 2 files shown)


resources/schema/Executability.md
```diff
@@ -0,0 +1,35 @@
+## Executability
```
RamilCDISC: Could you please make its size  heading 1,  for consistency with pattern of other documentation.

RamilCDISC (changes requested):
I ran the editor locally with schema url pointing to core-base.json of this branch, when i hover over executability, documentation appears but the heading where all options are supposed to be visible is empty and shows only pipe symbols. Please see the attached screenshot. Please let me know i have missed something in local execution. 

Image

resources/schema/Executability.md
RakeshBobba03: I changed from markdownDescription to title for individual enum values and I think it should resolve the || issue, and also I updated the heading size. If you run into any issues, please let me know.

gerrycampion (approved):

Issue #791, opened 2024-08-14
define dataset variables for Dataset Metadata Check against Define XML


SFJohnson24:
currently for rules that are Dataset Metadata Check against Define XML, the ContentsDefineDatasetBuilder does not report define_dataset_variables.  Having this produced as one of the pieces of metadata would be useful for rule development

SFJohnson24:
see also: https://github.com/cdisc-org/cdisc-rules-engine/issues/784

mhungria:
I look forward to this enhancement.

gerrycampion:
I think we should instead use: `define_dataset_variable_order`
This will eliminate the need for #784 `get_column_order_from_define` operation

mhungria:
I think there may be confusion on the functionality I requested with get_column_order_from_define ("Returns list of dataset columns in order" as per the Define-XML document; i.e., list of ItemGroupDef/ItemRef ordered by ItemGroupDef/ItemRef/@OrderNumber. It should be similar to get_column_order_from_dataset ("Returns list of dataset columns in order" as included in the actual dataset).
Note the metadata variable define_variable_order_number (ItemGroupDef.ItemRef.OrderNumber) is only one value, not a list.

gerrycampion:
@mhungria, I think we are on the same page. My idea is that given the following XML:

```xml
<ItemGroupDef OID="IG.SE" Name="SE" Domain="SE">
    <ItemRef ItemOID="IT.SE.STUDYID" OrderNumber="1"/>
    <ItemRef ItemOID="IT.SE.DOMAIN" OrderNumber="2"/>
    <ItemRef ItemOID="IT.SE.USUBJID" OrderNumber="3"/>
    <ItemRef ItemOID="IT.SE.ETCD" OrderNumber="5"/>
    <ItemRef ItemOID="IT.SE.SESEQ" OrderNumber="4"/>
</ItemGroupDef>
```

`define_dataset_variable_order` will return a list with:
- STUDYID
- DOMAIN
- USUBJID
- SESEQ*
- ETCD*

*Note that ETCD and SESEQ are switched because `OrderNumber` attribute exists

given the following XML:
```xml
<ItemGroupDef OID="IG.SE" Name="SE" Domain="SE">
    <ItemRef ItemOID="IT.SE.STUDYID"/>
    <ItemRef ItemOID="IT.SE.DOMAIN"/>
    <ItemRef ItemOID="IT.SE.USUBJID"/>
[...]

PR #1508, opened 2026-01-02
Add define_dataset_variable_order with OrderNumber sorting support (#791)

RakeshBobba03:
#791 Implements `define_dataset_variable_order` metadata variable that returns dataset variables ordered by OrderNumber when present in Define XML, otherwise falls back to XML document order. This adds the new field to `extract_dataset_metadata()` and `extract_domain_metadata()` methods in BaseDefineXMLReader, updates all relevant dataset builders to include the field in their output, adds schema validation in MetaVariables.json, and includes comprehensive unit tests. The sorting logic is centralized in `_get_ordered_dataset_variables()` method following DRY principles.

### DIFF
--- a/tests/unit/test_define_xml_reader.py
+++ b/tests/unit/test_define_xml_reader.py
@@ -507,6 +507,8 @@ def test_extract_dataset_metadata_without_ordernumber(filename):
             dataset_metadata["define_dataset_variable_order"]
             == dataset_metadata["define_dataset_variables"]
         )
+
+
 class TestGetExtensibleCodelistMappings:
     @staticmethod
     def _make_codelist(name, coded_values_extended, alias_list):


resources/schema/rule/MetaVariables.json
gerrycampion, line 1: we need to add these values to relevant rule types in Rule_Type.md

cdisc_rules_engine/dataset_builders/dataset_metadata_define_dataset_builder.py
gerrycampion: Do we need to update test_dataset_metadata_define_dataset_builder?

cdisc_rules_engine/dataset_builders/dataset_metadata_define_dataset_builder.py
```diff
@@ -63,13 +70,44 @@ def _get_define_xml_dataframe(self):
...
+                    basic_metadata["define_dataset_key_sequence"] = None
+                    basic_metadata["define_dataset_has_no_data"] = None
+                    enriched_metadata.append(basic_metadata)
+            else:
+                basic_metadata["define_dataset_variables"] = None
+                basic_metadata["define_dataset_variable_order"] = None
+                basic_metadata["define_dataset_key_sequence"] = None
+                basic_metadata["define_dataset_has_no_data"] = None
+                enriched_metadata.append(basic_metadata)
+        return self.dataset_implementation.from_records(enriched_metadata)
```
gerrycampion, line 104: i don't quite understand why this is all needed. iow, is it possible to just use `get_define_xml_item_group_metadata_for_dataset` here instead?

gerrycampion: From slack: 
My [Rakesh] thought process was:

I need one row per dataset in the Define XML (from get_define_metadata()), with each row enriched with variables, variable order, key sequence, and has_no_data. So we still have to loop over that list and, for each dataset, get the enriched metadata. We could do that by calling get_define_xml_item_group_metadata_for_dataset in the loop (passing something like {"dataset_name": dataset_name} for each item). We’d still need the same try/except and fallback when extraction fails or define_dataset_name is missing.

The reason I kept the current approach is that the helper creates a new Define XML reader on every call, so we’d be re-fetching and re-parsing the Define file once per dataset. I wanted to avoid that, so I used a single reader and called extract_dataset_metadata(dataset_name) in the loop.

Do you think we should switch to using the helper for consistency with the rest of the codebase, or keep this version for the single-reader behavior? Happy to go either way based on your preference.

cdisc_rules_engine/dataset_builders/dataset_metadata_define_dataset_builder.py
```diff
@@ -63,13 +70,44 @@ def _get_define_xml_dataframe(self):
...
+                    basic_metadata["define_dataset_variables"] = None
+                    basic_metadata["define_dataset_variable_order"] = None
+                    basic_metadata["define_dataset_key_sequence"] = None
+                    basic_metadata["define_dataset_has_no_data"] = None
+                    enriched_metadata.append(basic_metadata)
+            else:
+                basic_metadata["define_dataset_variables"] = None
+                basic_metadata["define_dataset_variable_order"] = None
+                basic_metadata["define_dataset_key_sequence"] = None
+                basic_metadata["define_dataset_has_no_data"] = None
```
gerrycampion: instead of hardcoding these multiple times, can you create a method that takes `basic_metadata` and the `define_col_order ` and assigns None to any properties that are missing in `basic_metadata` but present in the `define_col_order`

gerrycampion: Sorry, another update to this - I noticed that we have this function: https://github.com/cdisc-org/cdisc-rules-engine/blob/406ed1714f4cc1b02b87b747d6117ba36f046916/cdisc_rules_engine/dataset_builders/dataset_metadata_define_dataset_builder.py#L112
This seems to be doing something similar to what you are doing in this block. Shouldn't we handle it in this similar way for _get_define_xml_dataframe?

RakeshBobba03: Added `_ensure_define_metadata_keys(metadata, col_order)` that sets `None` for any key in `col_order` that’s missing from `metadata`, and use it in both the except and else branches so we don’t hardcode the same four assignments. Same idea as `_ensure_required_columns` (ensure required fields exist), but for the dict case with a single default.

resources/schema/rule/MetaVariables.json
RakeshBobba03: Added `define_dataset_variable_order` from MetaVariables.json to the three rule types that expose it: **Dataset Contents Check against Define XML**, **Dataset Metadata Check against Define XML**, and **Define Item Group Metadata Check**. I didn’t add it to **Domain Presence Check against Define XML** because that rule type uses `read()` only, which doesn’t provide `define_dataset_variable_order`.

Issue #1197, opened 2025-06-03
Operation: domain_is_custom is currently not working for TIG standard.


eljanssens:
- Domain_is_custom works for CG0349 test data (product = sdtm, version = 3-4)
- Domain_is_custom doesn't work for TIG0514 test data (product = tig version = 1-0 substandard = sdtm)

Rule logic is the same so operation should work similarily.

[unit-test-coreid-CG0349-negative1_20250527.xlsx](https://github.com/user-attachments/files/20574371/unit-test-coreid-CG0349-negative1_20250527.xlsx)

[unit-test-coreid-TIG0514-negative.xlsx](https://github.com/user-attachments/files/20574382/unit-test-coreid-TIG0514-negative.xlsx)

SFJohnson24:
I have reverted PR 1197 @RakeshBobba03 @RamilCDISC 
The main piece of problematic logic being
```     def _check_domain_in_use_case(
        self, domain: str, use_cases: List[str], substandard: str
    ) -> bool:
        allowed_domains = self._get_allowed_domains_for_use_cases(
            use_cases, substandard
        )
        if domain in allowed_domains:
            return True
        if self._is_custom_domain(domain):
            return True
        return False
```
This logic is fundamentally incorrect and would have caused all custom domains to run for all TIG rules, bypassing any current rule scoping that was in place.
This PR nearly made it into a production release which would have made it a much bigger deal than it is presently.  Please avoid implementing catch-all solutions or broad exception handling that performs a workaround that bypass the core validation logic and please, reach out with technical questions before implementing solutions if you're unsure about the architectural implications of a change.  Gerry and I are resources and here to help--don't be afraid to reach out.

SFJohnson24:
see my comments here on this ticket: https://jira.cdisc.org/browse/CORERULES-10223

I spoke with @eljanssens regarding this and YP should have been skipped 'YP does have a topic variable, but it is a --PARMCD which is linked to trial design or study reference which are not in scope of this rule' so the class skip was technically correct for it. 
I am working on https://github.com/cdisc-org/cdisc-rules-engine/issues/1451 which will also need to ID custom domains for TIG-- @RakeshBobba03 we should discuss with Els and scope out a function--maybe at scrum tomorrow or atleast make sure that the _handle_special_cases operation is applicable to TIG as well as SDTM

SFJohnson24:
@eljanssens I added the --use-case parameter to the CLI.  
[...]

PR #1511, opened 2026-01-02
#1197 Fix domain_is_custom operation for TIG standards and refactor domain custom check logic

RakeshBobba03:
Issue #1197: This PR implements custom domain detection by introducing a LibraryMetadataContainer that stores standard metadata (including domain sets) and exposes an is_domain_custom method to determine if a domain exists in the standard, a new domain_is_custom operation for rule authors to check custom domains, updates to CDISCLibraryService to extract and cache domain names from standard metadata via get_standard_details, and integration in RuleProcessor to use custom domain detection for use case validation, ensuring custom domains are handled correctly when pulling variable metadata from the model hierarchy rather than the implementation guide.


cdisc_rules_engine/services/cdisc_library_service.py
```diff
@@ -679,5 +678,11 @@ def _extract_domain_names_from_tabulation_standard(
         domain_names: Set[str] = set()
         for cls in standard_data.get("classes", []):
             for dataset in cls.get("datasets", []):
-                domain_names.add(dataset.get("name"))
+                domain_name = dataset.get("name")
+                if domain_name:
+                    domain_names.add(domain_name)
+        for dataset in standard_data.get("datasets", []):
```
RamilCDISC, line 684: Could you please explain briefly why we are now dataset names too and adding them to domain names?

RakeshBobba03: I noticed that TIG standards can have datasets in two places: nested under classes (like SDTMIG/SENDIG) and at the top level of the standard_data structure. The first loop handles datasets in classes. The second loop captures top-level datasets that appear in TIG standards.

I saw this pattern in _get_therapeutic_ig_codelists (around line 528), where TIG standards with substandard 'sdtm' or 'send' process top-level datasets separately from class datasets. Without the second loop, we'd miss domain names for those top-level datasets in TIG standards, which would cause domain_is_custom to incorrectly classify them as custom domains.

tests/unit/test_operations/test_domain_is_custom.py
RamilCDISC, line 1: I have reviewed and done manual validation. All seems good. Before merging could you please add a case where you use DASK too for new added test cases.

RamilCDISC (changes requested):

cdisc_rules_engine/services/cdisc_library_service.py
RamilCDISC: Thank you for your explanation.

SFJohnson24:
Hey @RakeshBobba03 @RamilCDISC I was looking this over and it seems this is not completely correct.  The YP domain in the test data Els provided is custom and the PMVALU column for it is expected to be flagged as an issue.  Currently it is skipping it and saying it is out of scope for the rule.  I have not reopened the branch but I moved the ticket out of the closed column as it is not closed.

cdisc_rules_engine/services/cdisc_library_service.py
SFJohnson24: for the future @RakeshBobba03 model defines the datasets and standard defines the domains.  Your code conflated the two.
image
image
Figured it was worth pointing out--it took me awhile to get model and standard down in my mind

PR #1518, opened 2026-01-08
reorder readme, fix tests

SFJohnson24:
I was in the process of giving the readme a reorg for more logical flow for the ClI.  This PR includes that.

The changes to close down manager at the end of a validation causes issues as 2 cache_managers are created for each subsequent validation run with the single command.  Python's multiprocessing leaves behind global state from the first validation. The first validation thus affects the multiprocessing infrastructure of the second.  This resulted in a ```BrokenPipeError: [WinError 232] The pipe is being closed.``` the pipes from the first manager are closed, and the second manager can't properly establish new ones in the same process.  Splitting the tests splits the processes so this issue does not occur.

see: https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/20826120159

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Once downloaded, simply unzip the file and run the following command based on yo
 
 ## Command-line Interface
 
-**Note**: the following examples are applicable to the source code and have references to "`python core.py`". When using the executable version as described in the [Quick Start](#quick-start) above, instances of "`python cored.py`" should be replaced with "`.\core.exe`" (Windows) or "`./core`" (Linux/Mac). You can also run directly on the source code by following the [Cloning](#cloning) instructions.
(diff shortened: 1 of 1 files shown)


core.py
```diff
@@ -787,8 +787,8 @@ def list_ct(cache_path: str, subsets: tuple[str]):
 
 
 @click.command()
-def test_validate():
-    """**Release Test** validate command for executable."""
+def test_validate_json():
```
gerrycampion: are these methods similar enough that json/xpt selector can just be a cli arg?

README.md
gerrycampion, line 1: why the reordering in the readme? The intent of the current readme ordering is to cater to end-users first (release and cli), then developers (github clone), then contributors.

SFJohnson24: I would assume for the CLI they would want to set up a venv and clone the github code instead of installing dependencies globally?

core.py
SFJohnson24: I merged them with a file type argument.  re-running of test release is linked above in original PR description

README.md
gerrycampion: venv and installing dependencies also goes under devs. end users can just run the executable without that, right?

SFJohnson24: think i misunderstood this--rolled it back (fixed 1 spell mistake)

gerrycampion (approved):

PR #1524, opened 2026-01-09
run test suite and editor preview for all PRs--internal and external

SFJohnson24:
this PR grabs the PR reference in the CI and passes it to test suite and the selenium e2e testing.  It then uses that to checkout the code for the test suite/editor preview run.  I also allowed manual dispatch of these actions with PR reference #'s, you can leave this blank and it will run it on the current branch that is selected or give it a particular PR (internal or external)


.github/workflows/test_rule_editor_preview.yml
```diff
@@ -2,7 +2,17 @@ name: Run Tests on Preview Deployment
...
+      pr_ref:
+        description: "Git reference to checkout (for PRs)"
+        required: false
+        type: string
   workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "PR number to test (leave empty for current branch)"
+        required: false
+        type: string
```
gerrycampion, line 15: i think we should add this to test_unit as well

SFJohnson24: @gerrycampion https://github.com/cdisc-org/cdisc-rules-engine/pull/1519 this PR resolves the unit tests for outside PRs.  Our unit tests dont make any API calls but the config service wants one when it is initialized.  If we give the unit tests a fakekey, they run appropriately for all internal and external PRs

gerrycampion (approved):

Issue #1022, opened 2025-02-13
Enhance UTF Handling in JSON Dataset Reader


SFJohnson24:
The dataset reader currently assumes utf8 and ascii characters, which causes issues when parsing JSON files containing international characters.   Since JSON is inherently Unicode (UTF-8 by default), we need to ensure that the dataset reader properly handles UTF-8, UTF-16, and UTF-32 encoded data.

A/C:   The dataset reader should be able to parse and process datasets containing international characters without issues.
- Update the all dataset reader to explicitly handle Unicode encoding.
- Ensure proper decoding when reading files.
- Add test cases with non-ASCII characters (Chinese, Japanese, etc.).
- both Test and validate commands should work

Errors:
this is the error with validate 
<img width="1326" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM2ZiQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a5a744dababddbf9199c17b333b851436ec48141/image.png" alt="image.png" />
this is the error with test
<img width="1172" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM2piQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--064b32fbf7f6c0795dcee9eb5da9b127d4cea8d7/image.png" alt="image.png" />

DatasetJSON:
[ae_nonascii.zip](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMmZiQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--44b07d0f80fa06337841ddc9d4e8ffb101093202/ae_nonascii.zip)
Test Command JSON
[TestDatasets.zip](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM0hiQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6ff2a3cbedabb063c569cec0ea475cbc01fc302a/TestDatasets.zip)

code:
[...]

PR #1540, opened 2026-01-14
1174: json load fallback logic. update-cache encoding parameter.

alexfurmenkov:
added fallback logic while reading json. cache input file encoding can be specified with parameter.


SFJohnson24 (changes requested):
This looks good-- can you update the readme to reflect the new command and describe the fallbacks.  I would also make a not regarding the encodings we can currently run/expect as #1022 is still not completed so even if they can put any encoding into the cache, they will not be able to run them

PR #1547, opened 2026-01-20
Python

SFJohnson24:
this PR adds a python version check and a warning if users are not using 3.12.


cdisc_rules_engine/utilities/utils.py
```diff
@@ -481,3 +483,14 @@ def set_max_errors_per_rule(args):
...
+
+def python_version_check():
+    python_version = sys.version_info
+    if python_version.major != 3 or python_version.minor != 12:
+        logger.warning(
+            f"This tool is designed for the Python version outlined in the top of the readme."
+            f"You are using Python {python_version.major}.{python_version.minor}.{python_version.micro} "
+            f"You may experience unexpected errors or issues with the validation"
+        )
+    return
```
RamilCDISC: The AC requires that the validation exits if its not correct python version. With current code only warning will be logged but validation will continue.

SFJohnson24: i thought this through and alterred this--code now enforces a minimum version and warns if it is a newer version of python >3.12 as these are untested

gerrycampion (approved):

PR #1551, opened 2026-01-20
471 improved issue reporting

gerrycampion:
### Rules Report tab
Adds more status to make 4 possible statuses: EXECUTION_ERROR, ISSUE_REPORTED, SKIPPED, SUCCESS. ISSUE_REPORTED was made to distinguish between Rules that ran successfully and produced no issues (SUCCESS), vs rules that ran succussfully and produced issues (ISSUE_REPORTED). EXECUTION_ERROR was made to distinguish from normal SKIPPED status due to COLUMN_NOT_FOUND_IN_DATA, DOMAIN_NOT_FOUND, SCHEMA_VALIDATION_IS_OFF, OUTSIDE_SCOPE, vs other types of execution errors.

### Issue Details tab
Details rule execution errors and rule evaluation errors. These show up with a message starting with "rule execution error -" or "rule evaluation error -".

### Sample output changes
Sample vanilla test updates: https://github.com/cdisc-org/CORE_Test_Suite/pull/78/changes
Sample USDM test updates: https://github.com/cdisc-org/CORE_Test_Suite/pull/82/changes
You can also view the artifacts for these corresponding runs to see the excel reports
These will need to be merged for the test suite check to pass

### Notes

Preferably, this QOL PR would be merged first: https://github.com/cdisc-org/CORE_Test_Suite/pull/79
I will need to make changes in the rule editor as well here: https://github.com/cdisc-org/conformance-rules-editor/pull/197
I created a followup issue to resolve the Execution errors that are found in the current reports: https://github.com/cdisc-org/cdisc-rules-engine/issues/1578
Note that I did not create a column in the report for skip reason because the report is at the rule level. In order to have skip reasons, we would need a new tab with a matrix of rules x datasets to provide a reason for skip at each dataset within each rule.


gerrycampion:
> After the last push, the issues count in issue summary sheet has increased from 9 to 34. The PR does not seems to be updating something related to rule execution. The count should not in increase. Please let me know if i am not understanding something fully and this is correct behavior.

This is what I expect to happen. Before the last push, I was adding the new execution errors to the Issue Details tab, but wasn't adding these to the Issue Summary tab. Now they are also added to the Issue Summary tab. It covers these issues:
image

Note the changes in the PR expose errors that were previously treated as skips. These errors will need to be fixed or categorized intentionally as skips. That is what this upcoming Ticket #1578 is for.

PR #1556, opened 2026-01-22
TIG use case

SFJohnson24:
this PR along with the editor PR adds -use-case to scope rules for TIG.  This allows the custom domains to be ruled as in scope and have the rule fire for them.  
This resolves the issue with the custom domains running for TIG0514 from the original ticket.

Note: the YP class is out of scope so that rule will not fire but the two X- domains will


cdisc_rules_engine/services/reporting/sdtm_report_data.py
```diff
@@ -67,20 +67,22 @@ def __init__(
...
+                use_case=use_case,
             ),
             "Dataset Details": self.get_dataset_details_data(),
             "Issue Summary": self.get_summary_data(),
             "Issue Details": self.get_detailed_data(),
             "Rules Report": self.get_rules_report_data(),
         }
 
-    def get_conformance_details_data(
+    def get_conformance_details_data(  # noqa
```
RamilCDISC, line 85: Could you please mention the reason for skipping the lint check here now?

core.py
```diff
@@ -363,7 +375,7 @@ def _validate_no_arguments(logger) -> None:
     ),
 )
 @click.pass_context
-def validate(
+def validate(  # noqa
```
RamilCDISC, line 383: same as above.

cdisc_rules_engine/services/reporting/sdtm_report_data.py
SFJohnson24: ignoring complexity check to keep the conformance detail data as one function--it is not complex, it just needs to fill in each field of the conformance details

core.py
SFJohnson24: same as above--I wanted to keep validate as a single functional unit

Issue #1543, opened 2026-01-15
Update Cache Naming strategy


SFJohnson24:
currently our cache key naming strategy involves using the directory path of the dataset.  If that path is quite large, the key is too large and the dataset is not cached--we should consider a different, unique key name that will not have this result 

<img width="1894" height="778" alt="Image" src="https://github.com/user-attachments/assets/3f9ad1e6-1d10-4a71-9c9b-0747e50fbdd8" />

alexfurmenkov:
Hey @SFJohnson24, looks like the pic wasn't attached

SFJohnson24:
@alexfurmenkov thanks, reattached

alexfurmenkov:
Turned out it was a cache size issue and not key length

PR #1557, opened 2026-01-23
#1543 tests to check if cache key length influence caching behavior.

alexfurmenkov:


gerrycampion:
PR to prove it is not an actual issue

PR #1570, opened 2026-01-26
1516: Refactor target name handling to preserve order

alexfurmenkov:


SFJohnson24 (approved):
this PR passes high level review and currenctly maintains order of output variables from rules into report output-- currectly resolving attached issue

Issue #1348, opened 2025-09-19
Incorrect results with "not" as first Check condition


ASL-rmarshall:
If the first check condition is `not`, it is ignored - as in:
```yaml
Check:
  not:
   ...
```
Example rule with `not` first that reports incorrect, reversed results: [DDF00015 - does not work.yaml](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1J6Q3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--130f424337f0b3171b8c152339729cb0343eb757/DDF00015%20-%20does%20not%20work.txt)
Example rule with redundant `all` first that reports correct results: [DDF00015 - works.yaml](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1p6Q3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--98a0ef221dbf337a58fe469dd825f83afa1edb12/DDF00015%20-%20works.txt)

Test datasets:
- Positive: [regression-test-coreid-DDF00015-positive.json](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd2R6Q3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--67695130a54be131cd72ed4556c6c2efb4658ddd/regression-test-coreid-DDF00015-positive.txt)
- Negative: [regression-test-coreid-DDF00015-negative.json](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd2x6Q3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--dbc3683d1df51fd7bd84c5cdc2946118687b4594/regression-test-coreid-DDF00015-negative.txt)

It is sometimes easier to express a rule as a negation of the correct conditions, but this this currently not possible without inserting an unneeded `all` as the first condition.

[Note that these rules use the updated `codelist_terms` operation - #1326]

PR #1574, opened 2026-01-29
#1348 fix not as head condition provides proper AST

alexfurmenkov:


SFJohnson24 (approved):
this PR correctly resolves the issue where empty condition composites are returned when NOT conditions are the sole condition.  Rule and both datasets from original ticket now execute correctly with the Not rule and no longer require the all conditon

PR #1580, opened 2026-01-30
1329: Add DatasetIsCustom operation to determine if a dataset is custom

alexfurmenkov:


cdisc_rules_engine/models/sdtm_dataset_metadata.py
```diff
@@ -12,22 +12,27 @@ class SDTMDatasetMetadata(DatasetMetadata):
...
-    | SUPPQSXX | SUPPQS       | True    | None   | QS      | False |           |
-    | APQS     | APQS         | False   | APQS   | None    | True  | QS        |
-    | APQSX    | APQS         | False   | APQS   | None    | True  | QS        |
-    | APQSXX   | APQS         | False   | APQS   | None    | True  | QS        |
-    | SQAPQS   | SQAPQS       | True    | None   | APQS    | True  |           |
-    | SQAPQSX  | SQAPQS       | True    | None   | APQS    | True  |           |
-    | SQAPQSXX | SQAPQS       | True    | None   | APQS    | True  |           |
-    | RELREC   | RELREC       | False   | None   | None    | False |           |
-    """
+    | name     | unsplit_name | is_supp | domain | rdomain | is_ap | ap_suffix | dataset_is_custom | related_domain	| related_domain_is_custom |
```
SFJohnson24: Can we add this to the reference guide.  This would need a PR from rule editor.  https://github.com/cdisc-org/conformance-rules-editor/blob/main/docs/README.md
sidebar grabs these files: https://github.com/cdisc-org/conformance-rules-editor/blob/main/docs/_sidebar.md?plain=1 to create the reference guide.  It would be good to have this chart in there for end-users to see as well as reference from these operations

resources/schema/rule/Operations.md
```diff
@@ -458,6 +458,54 @@ Output
 true
 ```
 
+### related_domain_is_custom
+
+Checks whether the related domain (for example, the parent domain of a SUPP or RELREC dataset) is not present in the set of standard domains for the provided standard and version. This is useful for determining whether relationships point to non-standard or custom domains.
+
+Input
+
+Target Domain: SUPPEX
```
SFJohnson24, line 467: this notation with the target domain / standard / version strays from what readme formatting.  Can we remove it?

alexfurmenkov: I can remove these fields; however, in that case the description of my operations will deviate from the overall style, since these fields are used in the descriptions of the neighboring operations (such as domain_is_custom, standard_domains, extract_metadata, expected_variables, etc.).

SFJohnson24: I missed this, leave them then

cdisc_rules_engine/models/sdtm_dataset_metadata.py
alexfurmenkov: Hi @SFJohnson24, I added the table in Readme (https://github.com/cdisc-org/conformance-rules-editor/pull/291)

RamilCDISC: @SFJohnson24 if @alexfurmenkov PR fulfills what you wanted please let us know here too. I will do a final validation and approve then.

alexfurmenkov:
Hi @RamilCDISC ,
I checked your case. There were not EX and AE names in library datasets, because of dataset_is_custom is True.
image

SFJohnson24:
@alexfurmenkov good catch by @RamilCDISC  Custom datasets must be from general observation classes which are defined in the standard, not the model.  So the operation needs to pull from the IG standard not the model to determine if a dataset is custom.

alexfurmenkov:
During the implementation of the dataset_is_custom operation, we concluded that the dataset name can only be retrieved from model_metadata, since standard_metadata does not contain dataset names - it only contains domains.

If we were to take domains from standard_metadata, the check would effectively become identical to domain_is_custom. Additionally, we believe that the dataset name does not always match the domain name.

Therefore, in order to address this limitation, we would like to clarify where the dataset names from the library should be retrieved from.
image

gerrycampion:
@alexfurmenkov I agree with Sam.
My proposal in the attached ticket was to use `dataset_is_custom` and `related_domain_is_custom` (instead of `domain_is_custom`) to disambiguate things. For SDTM, in both cases, you'll need to refer to the list of domains in the IG, not the model.
- For the case of SDTM `dataset_is_custom`, you'll want to compare the DOMAIN value with the list of domains in the IG. If there is no DOMAIN value, the result is also false as it is not applicable. 
- For the case of SDTM `related_domain_is_custom`, you'll want to compare the `related_domain` attribute with the list of domains in the IG. Again, if there is no `related_domain` attribute, the result is false. To determine `related_domain`, you will need to derive this depending on the dataset type. See the chart and let me know if you have specific questions about this.
- For non-SDTM, I'm not sure how custom datasets work, so you'll need to speak with someone more familiar. However, it is not part of release 1.0, so not important at this time.

alexfurmenkov:
Hi @gerrycampion ,
So we are renaming the domain_is_custom operation to dataset_is_custom. That means dataset_is_custom will determine it based on the domain rather than the dataset name, comparing it with the domains from standard metadata. And we are also adding related_dataset_is_custom, correct?

gerrycampion:
> Hi @gerrycampion , So we are renaming the domain_is_custom operation to dataset_is_custom. That means dataset_is_custom will determine it based on the domain rather than the dataset name, comparing it with the domains from standard metadata. And we are also adding related_dataset_is_custom, correct?

After more thought, this really only applies to domains at the moment. I think we should continue to call it `domain_is_custom` and add an operation `related_domain_is_custom`. I've updated the table in the attached issue.

alexfurmenkov:
> > Hi @gerrycampion , So we are renaming the domain_is_custom operation to dataset_is_custom. That means dataset_is_custom will determine it based on the domain rather than the dataset name, comparing it with the domains from standard metadata. And we are also adding related_dataset_is_custom, correct?
> 
> After more thought, this really only applies to domains at the moment. I think we should continue to call it `domain_is_custom` and add an operation `related_domain_is_custom`. I've updated the table in the attached issue.

@gerrycampion ok, pls re-review

PR #1607, opened 2026-02-10
Update core.py

SFJohnson24:
this PR normalizes standard before it is ingested by engine via validation_args.  This causes downstream issues where the standard is not identified.


mhungria:
I recommend to adjust the core engine treatment of standard as it is defined in CT for Define-XML, Standard Name (case sensitive); i.e., the user should refer to SDTMIG, ADaMIG, etc. (see https://library.cdisc.org/browser/#/mdr/ct/2025-09-26/packages/define-xmlct-2025-09-26/codelists/C170452).

The core engine could provide a message to the users indicating that the standard indicated is not available in CT for any variations of case not in CT.

Note: As CDISC Standard developers and part of CDISC Standards implementation teams, we need to work on better integration across standards and systems.

eljanssens:
@mhungria since this is a CLI and not a CDISC standard, we are not going to enforce people to add CDISC CT correct commands. This will be challenging with for example ADaMIG. The engine will now handle the standards case insensitive which should solve your problem. As long as the standard wording is correct, the engine will work.

PR #1612, opened 2026-02-11
update expected json and workflow to resolve import issue

RamilCDISC:
The PR updates the expected result in end2end testing. The change is because of new executionStatus added to engine and updated rule. It also updates the workflow to pin setuptools version to <82 to avoid `ModuleNotFoundError: No module named 'pkg_resources'`


RamilCDISC:
a successful run can be seen [here](https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/21922551697/job/63306415837)

PR #1613, opened 2026-02-11
1274 Add minus operation for set difference to support CORE-000334 expected-variable checks

RakeshBobba03:
Issue #1274 asked for a way to compare output variables in validation reports. Rather than adding a special "compared" flag in reporting, we chose to add a `minus` operation so rules can compute set differences themselves and expose them as output variables. That gives the engine a single, consistent mechanism: rules produce comparison results via operations, and the existing output-variable reporting surfaces them.

This PR adds a `minus` operation that implements that set-difference logic. It takes `name` (minuend) and `value` (subtrahend) as references to other operations and returns elements in `name` that are not in `value`, keeping the order from `name`. When `value` is empty or missing, it returns all elements from `name`.

Implementation changes include a new `Minus` operation class with `_normalize_to_list` and `_set_difference_preserve_order` helpers; an optional `value` field on `OperationParams`; passing `value` from the rule processor into `OperationParams`; registering `minus` in the operations factory; and updating the Operations JSON schema to define and require the `value` property for the minus operator. Operations.md documents the operation with an example. Unit tests cover the helper logic, main behavior, both PandasDataset and DaskDataset, the empty-value case, and the missing-name-ref case.

Files used for testing:
[CORE-000334.yaml](https://github.com/user-attachments/files/25244049/CORE-000334.yaml)
[unit-test-coreid-CG0016-negative.xlsx](https://github.com/user-attachments/files/25244050/unit-test-coreid-CG0016-negative.xlsx)

### DIFF
--- a/resources/schema/rule/check_parameter.md
+++ b/resources/schema/rule/check_parameter.md
@@ -1,4 +1,3 @@
-
 ## Overview
 
 Check parameters are configuration elements that define how validation rules are applied within the CDISC rules engine. These parameters control the behavior, scope, and criteria for data validation checks across clinical trial datasets. Each parameter serves a specific purpose in customizing rule logic to ensure data integrity and compliance with CDISC standards.
@@ -266,6 +265,17 @@ Expected return type for operation results. Valid values are "code" (for NCI cod
 
 Either "submission" or "evaluation" for which dataset to check the variable_is_null from. Evaluation is the dataset constructed by
 the rule type while submission is the raw dataset submitted that is being evaluated.
+
+### subtract
+
+Reference to another operation result, used as the second operand in operations that take two inputs. For example, in the `minus` operation, `name` references the minuend (first list) and `subtract` references the subtrahend (second list); the operation returns elements in the first list that are not in the second.
+
+```yaml
+- id: $expected_minus_dataset
+  name: $expected_variables
+  operator: minus
+  subtract: $dataset_variables
+```
 
 ### term_code
 


cdisc_rules_engine/operations/operations_factory.py
RamilCDISC, line 1: Could you please set this file. it seems as there is no much changes but for some reason whole file is in diff. It is difficult to review like this.

RakeshBobba03: I made the update, it was a formatting issue

tests/unit/test_operations/test_minus.py
RamilCDISC, line 1: Could you please add test cases for duplicate value and scalars. Also for the case where value in the list is empty string ""

cdisc_rules_engine/operations/minus.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def _execute_operation(self):
+        name_ref = self.params.target  # minuend (list A)
+        value_ref = self.params.value  # subtrahend (list B)
+
+        if not name_ref or name_ref not in self.evaluation_dataset.columns:
+            return []
+        list_a = self.evaluation_dataset[name_ref].iloc[0]
+        if not value_ref or value_ref not in self.evaluation_dataset.columns:
+            # listA minus empty = all of listA (per Sam)
```
RamilCDISC: Please remove the name form comment. If you would like to highlight for Sam, you can send him a message.

RakeshBobba03: sorry, those were meant for me when making changes, must've missed to take them out. Taking them out ASAP

resources/schema/rule/Operations.md
```diff
@@ -750,6 +750,34 @@ Operations:
     operator: get_column_order_from_dataset
 ```
 
+### minus
+
+Computes set difference: elements in `name` that are not in `value`. Uses [set difference](<https://en.wikipedia.org/wiki/Set_(mathematics)#Set_difference>) semantics (A ∖ B). Preserves order from the first list. Both `name` and `value` must reference other operation results (e.g., `$expected_variables`, `$dataset_variables`).
+
+When `value` is empty or missing, returns all elements from `name`.
```
SFJohnson24: why is this line separate from the rest of the operation description?

resources/schema/rule/Operations.md
```diff
@@ -750,6 +750,34 @@ Operations:
...
+Operations:
+  - id: $expected_variables
+    operator: expected_variables
+  - id: $dataset_variables
+    operator: get_column_order_from_dataset
+  - id: $expected_minus_dataset
+    name: $expected_variables
+    operator: minus
+    value: $dataset_variables
+Check:
```
SFJohnson24: this example is a bit bloated.  Can we only show the operation yaml and add a line to the description that says simply. 'Can be computed and added to output variables to display missing elements in error results'

cdisc_rules_engine/models/operation_params.py
```diff
@@ -57,6 +57,7 @@ class OperationParams:
     regex: str = None
     returntype: str = None
     target: str = None
+    value: str = None
```
SFJohnson24: you will also need to make a PR to editor for the reference guide for the parameter you are adding to keep our docs up to date.  https://github.com/cdisc-org/conformance-rules-editor/blob/main/docs/check_parameter.md   this has all the parameters and a description of them.

resources/schema/rule/Operations.md
RakeshBobba03: This was by accident. Merged into a single paragraph: the “When subtract is empty or missing, returns all elements from name” wording is now part of the main minus description, and the whole section uses subtract instead of value.

resources/schema/rule/Operations.md
RakeshBobba03: Updated: the minus example now shows only the Operations YAML block, and the description ends with: “Can be computed and added to output variables to display missing elements in error results.”

cdisc_rules_engine/models/operation_params.py
RakeshBobba03: Done. docs/check_parameter.md in the conformance-rules-editor repo now documents the subtract operation parameter (with description and minus example). PR: https://github.com/cdisc-org/conformance-rules-editor/pull/293.

SFJohnson24: you still need to get rid of value since we are replacing it subtract

RakeshBobba03: Removed value from OperationParams, from the rule processor’s OperationParams construction, and from the Operations JSON schema.

SFJohnson24:
@RakeshBobba03 can you resolve the merge conflict.
check_parameter was also moved to engine.  https://github.com/cdisc-org/conformance-rules-editor/pull/293 I closed this--can you just move the text from it into https://github.com/cdisc-org/cdisc-rules-engine/blob/main/resources/schema/rule/check_parameter.md this file
we can get this merged then

Issue #1548, opened 2026-01-20
Recommendation to add "Global" to Sensitivity


mhungria:
Recommendation to add "Global" to Sensitivity:

We discussed this type of functionality during one of our prior office hours meetings and I mentioned it would be simple to do once we had sql added as functionality. I recall now that we agreed not to add a ticket yet; however, when discussed domain presence checks again during today's office hours, we said it could be evaluated to elevate the priority, so I am adding this ticket now.

Note it would allow to perform checks across the complete set included in a package and not repeat it when processing individual datasets. See https://github.com/cdisc-org/cdisc-rules-engine/issues/1330.
See also **Rule FB1101**.

gerrycampion:
For domain existence checks like FB1101, instead of using the not_exists operator, we would use a list of existing domains and a list of required domains and use the `contains_all` operator.

mhungria:
Yes, I agree. That's the reason I recommended to remove the rule type Domain Presence Check (See ticket #1330).

The recommendation on this ticket for Sensitivity for the error to be reported only once.

PR #1615, opened 2026-02-13
Fb0405

SFJohnson24:
This PR does a few things:
- revises domain presence check as the domain from define was getting grabbed and overriding domain from the dataset, if domain/dataset are not present but are referenced in define, their values are null.  
- add STUDY sensitivity and update docs.  This allows for a single output for a domain presence check (it would report for each dataset submitted previously as it was Dataset sensitivity).  Since a domain presence check is for the entire study, we want all errors for the study, reported only once.  This is what the sensitivity does.
- fix the report template as the CT was formatted to the right side of the cell
image

[Rule_underscores.json](https://github.com/user-attachments/files/25299928/Rule_underscores.json)
[CORE-Report-2026-02-13T14-04-32.xlsx](https://github.com/user-attachments/files/25299935/CORE-Report-2026-02-13T14-04-32.xlsx)

### DIFF
--- a/cdisc_rules_engine/models/actions.py
+++ b/cdisc_rules_engine/models/actions.py
@@ -68,7 +68,6 @@ class COREActions(BaseActions):
             error_object.domain = DomainPresenceValues.DOMAIN.value
             for error in error_object.errors:
                 error.dataset = DomainPresenceValues.DATASET.value
-                error.domain = DomainPresenceValues.DATASET.value
                 error.row = DomainPresenceValues.RECORD.value
         self.output_container.append(error_object.to_representation())
 
--- a/resources/schema/rule/Rule_Type.md
+++ b/resources/schema/rule/Rule_Type.md
@@ -192,8 +192,8 @@ all:
 
 One row per dataset defined in Define-XML:
 
-- `domain` - The domain if the dataset exists, empty otherwise
-- `filename` - The file name if dataset exists, empty otherwise
+- `domain` - The domain if the dataset exists, null otherwise
+- `filename` - The file name if dataset exists, null otherwise
 - `define_dataset_name`
 - `define_dataset_label`
 - `define_dataset_location`


cdisc_rules_engine/models/actions.py
```diff
@@ -65,8 +65,10 @@ def generate_dataset_error_objects(self, message: str, results: pd.Series):
         )
         if "domain presence" in self.rule.get("rule_type", "").lower():
             error_object.dataset = DomainPresenceValues.DATASET.value
+            error_object.domain = DomainPresenceValues.DOMAIN.value
             for error in error_object.errors:
                 error.dataset = DomainPresenceValues.DATASET.value
+                error.domain = DomainPresenceValues.DATASET.value
```
RamilCDISC: is this supposed to be DomainPresenceValues.DATASET.value or DomainPresenceValues.DOMAIN.value

resources/schema/rule/Rule_Type.md
```diff
@@ -192,8 +192,8 @@ all:
 
 One row per dataset defined in Define-XML:
 
-- `domain`
-- `filename` - The file name if dataset exists, null otherwise
+- `domain` - The domain if the dataset exists, empty otherwise
```
RamilCDISC: We are saying empty here but builder specifies None for this case. When None is json serialized it will be converted to null. Maybe we should say empty/null instead.

cdisc_rules_engine/models/actions.py
SFJohnson24: that is an artifact as error.domain doesnt exist.  error_object has domain--removed this.

resources/schema/rule/Rule_Type.md
SFJohnson24: I changed it back to null

RamilCDISC:
To update you in reference to #1548 @mhungria the functionality you described is covered by the new sensitivity "study" introduced in this PR. 
@SFJohnson24

RamilCDISC (approved):
The PR updates/fixes domain presence check and adds new sensitivity level. The validation was done by:

4. Ensuring the tests are updated for new changes.
5. Verifying the structure of the report.
6. Running manual validations using dev editor checking cases like.

- dataset present in the dataset file but missing in define.
- dataset present in define but missing in dataset file.
- dataset missing in both define and dataset file.
- multiple datasets missing in the dataset.
- verifying sensitivity set to STUDY provides only single output for domain presence check.
- verifying the report structure now is correct.

Issue #1616, opened 2024-02-15
For test data, better error message when the "Library" or "Datasets" tab, or the column headers within them are not found or malformed


chowsanthony:
Reproduce step: Name the test data spreadsheet tab "datasets" instead of "Datasets."

When the "Datasets" tab is not found or malformed, I see this error message, which I found a tad cryptic and not easy to troubleshoot:

```json
{
  "error": "KeyError",
  "message": "\"one or more datasets missing the following keys {'label'}\""
}
```

In contrast. when the "Library" tab is not found or malformed, I see this error message, which I found more informative:

```json
{
  "MI": [
    {
      "executionStatus": "execution_error",
      "domain": "MI",
      "variables": [],
      "message": "rule execution error",
      "errors": [
        {
          "error": "An unknown exception has occurred",
          "message": "Missing standard and version data"
        }
      ]
    }
  ]
}
```

Lastly, I hope the error message to include some coping mechanism, e.g.,, "Make sure there is a 'Datasets' tab, case-sensitive."

Another example: `Product` and `Version` within `Library` tab are also case sensitive and don't provide appropriate error messages.

SFJohnson24:
i know this ticket is from rule editor--we are decommissioning it but I do think there should be a more user-friendly error for the metadata readers and data reader classes

SFJohnson24:
I discussed with Rakesh this morning the A/C:
Some catches have been made for the excel reader already.
I also would like to see:

a. catches wrapped around the various reader classes in case an error occurs which returns an error that is user friendly.  This may need to be the metadata readers if they get called first in the validation script.  This may have already been handled partially with the encoding ticket but I think it would be good to have a catch surrounding our readers in-case the data is malformed that tells users 'your data file is incorrect/corrupted/etc and cant be read'

b. given the decommisioning of editor, i think we should add logic for the standard at the start of core.py.  we have the custom-standard flag which should override this.  We should have a list of constant standards supported from its file in the enum folder.  It should check if the standard.lower() is a supported standard https://github.com/cdisc-org/cdisc-rules-engine/blob/380f580d767e6dd100b0e2b1bfe12074410e1e52/core.py#L461

[...]

PR #1626, opened 2026-02-18
#1616: User-friendly errors for missing/wrong Datasets tab and columns, standard validation, and library metadata lookup failures

RakeshBobba03:
This PR improves error handling and messages for test data and standard/version usage (issue #1616 ).

**Datasets:** When the Excel test data workbook is missing the "Datasets" sheet or required column headers (e.g. "Filename", "Label"), the engine now raises a clear, user-facing error (e.g. `ExcelTestDataError` or `InvalidDatasetFormat`) that explains what is missing and that names are case-sensitive, instead of a generic `KeyError` such as "one or more datasets missing the following keys {'label'}". The API (TestRule) validates the incoming datasets payload and returns `BadRequestError` with guidance when required keys are missing. Reader calls in the validation script are wrapped so that malformed or unreadable data files produce a single, consistent error instead of raw exceptions.

**Standard and library metadata:** At the start of validation in `core.py`, the standard is checked against the supported list (`StandardTypes` enum); invalid values produce a clear error and point users to the `--custom-standard` flag. The `--custom-standard` flag correctly skips this check. When the standard/version combination has no library metadata (e.g. invalid Library tab or CLI arguments), the library lookup is wrapped in a dedicated catch that raises a specific error (`LibraryMetadataNotFoundError`) with a message that explains the invalid Standard/Version combination and hints at checking the Library tab or CLI arguments.

### DIFF
--- a/resources/schema/rule/Operator.md
+++ b/resources/schema/rule/Operator.md
@@ -992,7 +992,7 @@ Checking for consistent values across groups and validating that variables maint
 
 ### is_inconsistent_across_dataset
 
-Checks if a variable maintains consistent values within groups defined by one or more grouping variables. Groups records by specified value(s) and validates that the target variable maintains the same value within each unique combination of grouping variables.
+Checks if a variable maintains consistent values within groups defined by one or more grouping variables. Groups records by specified value(s) and validates that the target variable maintains the same value within each unique combination of grouping variables.  When inconsistency is detected within a group, the operator attempts to identify a majority value. If one value appears more frequently than all others, only the minority records (those not matching the majority value) are flagged. If no single majority exists — i.e., two or more values are tied for the highest frequency — all records in that group are flagged.
 
 Single grouping variable - true if the values of BGSTRESU differ within USUBJID:
 


TestRule/__init__.py
```diff
@@ -32,19 +46,38 @@ def validate_datasets_payload(datasets):
                 )
 
     if missing_keys:
-        raise KeyError(
-            f"one or more datasets missing the following keys {missing_keys}"
+        missing_list = sorted(missing_keys)
+        raise BadRequestError(
+            f"Test data is missing required dataset properties: {missing_list}. "
```
SFJohnson24: reporting the keys doesnt work--can we remove this logic.  if the dataset workbook is missing, it just says the label is missing. Since filename and label match with the corresponding dataset workbooks, it doesnt work exactly as implemented.  I do think checking and telling them about the Datasets works but the reporting logic with missing keys is not functioning correctly and could be skipped

TestRule/__init__.py
```diff
@@ -17,11 +24,18 @@ class BadRequestError(Exception):
     pass
 
 
+_DATASETS_TAB_GUIDANCE = (
+    "Make sure there is a 'Datasets' tab in your test data workbook (name is "
```
SFJohnson24: I am not wild about putting this at the top of the file since it is only needed by 1 function and only referenced once. This should be the error that on line 53.

cdisc_rules_engine/services/data_services/excel_data_service.py
```diff
@@ -199,23 +208,42 @@ def read_data(self, file_path: str) -> IOBase:
...
-                keep_default_na=False,
-            )
-        except TypeError as e:
-            logger.error(
-                f"Failed to read datasets from the Excel file at {self.dataset_path}. "
-                f"Ensure the file is in the correct format. "
-                f"Try opening and saving the file in Microsoft Excel. "
-                f"Error: {str(e)}"
-            )
+            with pd.ExcelFile(self.dataset_path) as xl:
```
SFJohnson24, line 210: if you give engine an excel sheet with no dataset tab, it does not make it into any of this logic.  https://github.com/cdisc-org/cdisc-rules-engine/blob/07357c45cf3b3dcff58d61c705fbb131db33d945/cdisc_rules_engine/utilities/utils.py#L487 this catches it.  
Image

SFJohnson24: if I give a dataset where the Datasets tab has a mismatch in the filename vs the sheet filename, it also gets caught by validate_dataset_files_exist

scripts/script_utils.py
SFJohnson24, line 126: I believe we will need some logic potentially here to address point D from my comment when incorrect CT pacakages are given

cdisc_rules_engine/services/data_services/excel_data_service.py
SFJohnson24: apologies--i did not have file extension.  will resume testing

TestRule/__init__.py
RakeshBobba03: The logic that reported which keys were missing is removed. We still check for required dataset properties and raise a single message that tells users to check the Datasets tab and column headers (case-sensitive), without listing key names.

TestRule/__init__.py
RakeshBobba03: The _DATASETS_TAB_GUIDANCE constant is removed. That text is now inlined in the single BadRequestError message in validate_datasets_payload (at the raise), so it’s only defined where it’s used.

scripts/script_utils.py
RakeshBobba03: Done. (1) In script_utils.get_library_metadata_from_cache, right after the CT loop, we now validate that every requested -ct package exists in the cache; if any are missing we raise CTPackageNotFoundError with the missing list, available packages, and an update-cache hint. (2) In run_single_rule_validation we added a check so that when codelists are requested but not found in the cache we raise CTPackageNotFoundError with a message to check the Library tab or codelist names (covers incorrect CT/standard/version from the editor). TestRule’s handle_exception returns 400 with the CT error message for both paths.

cdisc_rules_engine/exceptions/custom_exceptions.py
```diff
@@ -57,11 +79,35 @@ class InvalidDatasetFormat(EngineError):
     description = "Dataset data is malformed."
 
 
+INVALID_DATASET_FORMAT_REASON = (
```
SFJohnson24: both this and LIBRARY_METADATA_NOT_FOUND_HINT seem out of place in a file for exceptions.  I understand they are reused but they also seem unnecessary.  The exception is telling the user what went wrong, not how to fix it.  Telling the user their data at a path cant be read and also that the standard/version they entered failed to find library metadata is enough without these constants.

cdisc_rules_engine/services/data_services/excel_data_service.py
```diff
@@ -125,15 +130,19 @@ def get_raw_dataset_metadata(
...
-        datasets_worksheet = pd.read_excel(
-            self.dataset_path,
-            sheet_name=DATASETS_SHEET_NAME,
-            na_values=[""],
-            keep_default_na=False,
-        )
-        metadata = datasets_worksheet[
-            datasets_worksheet[DATASET_FILENAME_COLUMN] == dataset_name
-        ]
+        _worksheet = kwargs.get("_worksheet")
```
SFJohnson24: what is the purpose of this new worksheet argument and if/else handling?  I am assuming this is to prevent repeat excel reads?  We are already caching the result    @cached_dataset(DatasetTypes.RAW_METADATA.value)

scripts/run_validation.py
```diff
@@ -123,6 +132,37 @@ def initialize_logger(disabled, log_level):
         engine_logger.setLevel(log_level)
 
 
+def _get_datasets_or_raise(data_service):
```
SFJohnson24: this is a script helper and should be moved to ./script_utils.py

scripts/script_utils.py
```diff
@@ -117,6 +131,20 @@ def get_library_metadata_from_cache(args) -> LibraryMetadataContainer:  # noqa
...
+        if missing:
+            sorted_missing = sorted(
+                missing, key=lambda x: (x is None, str(x) if x is not None else "")
+            )
+            available = ", ".join(sorted(published_ct_packages)) or "(none)"
+            raise CTPackageNotFoundError(
+                f"{CT_PACKAGE_NOT_FOUND_PREFIX} in cache: "
+                f"{', '.join(str(c) for c in sorted_missing)}. "
+                f"Available packages: {available}. "
+                "Run 'core.py update-cache' to refresh the cache."
```
SFJohnson24: I entered an incorrect CT and landed in this error but updating the cache will not resolve my incorrect CT package.  Again, exceptions are for IDing what went wrong, can we please remove the hints for what the user needs to do to fix things.

TestRule/__init__.py
```diff
@@ -32,19 +42,46 @@ def validate_datasets_payload(datasets):
                 )
 
     if missing_keys:
-        raise KeyError(
-            f"one or more datasets missing the following keys {missing_keys}"
+        raise BadRequestError(
+            "Test data is missing required dataset properties. "
+            "This usually means the 'Datasets' sheet in your Excel file is missing "
```
SFJohnson24: 'test data is incorrect and missing required formatting'

cdisc_rules_engine/exceptions/custom_exceptions.py
```diff
@@ -57,11 +79,35 @@ class InvalidDatasetFormat(EngineError):
...
+
+
+CT_PACKAGE_NOT_FOUND_PREFIX = "Controlled terminology package(s) not found"
+
+
+class CTPackageNotFoundError(EngineError):
+    code = 400
+    description = (
+        f"{CT_PACKAGE_NOT_FOUND_PREFIX} in cache. "
+        "Check package names and run 'core.py update-cache' if needed."
```
SFJohnson24: again please remove hints.  telling them their CT is incorrect is enough

scripts/run_validation.py
```diff
@@ -259,8 +297,20 @@ def run_single_rule_validation(
     )
 
     ct_package_metadata = {}
+    codelists = codelists or []
     for codelist in codelists:
         ct_package_metadata[codelist] = cache.get(codelist)
+    if codelists:
```
SFJohnson24: Image
It seems I was mistaken about this.  Bad CT in testrule gets caught [here ](https://github.com/cdisc-org/cdisc-rules-engine/blob/fb8a8d5f123beaeba254e98b24604bcfb5a7f6a4/TestRule/__init__.py#L105) and errors out so this codeblock is never used.

scripts/script_utils.py
```diff
@@ -117,6 +131,20 @@ def get_library_metadata_from_cache(args) -> LibraryMetadataContainer:  # noqa
...
+            )
```
SFJohnson24: also: re this check--is there a reason we do this above the code below?  If we have a define, we get the CT packages from that so we would want to validate the case where the define has the wrong CT outlined in it.  I can give you a define.xml if you do not have one to test

cdisc_rules_engine/services/data_services/excel_data_service.py
RakeshBobba03: The `_worksheet` argument is there so we don’t re-read the same Excel sheet multiple times. In `get_datasets()` we parse the Datasets sheet once, then for each row we call `get_raw_dataset_metadata(dataset_name, _worksheet=worksheet)`. Without passing the worksheet, each of those calls would read the sheet again before the cache is populated. The `@cached_dataset` decorator helps when the same dataset is requested again later, but on the first pass we’d still do N reads of the same sheet. Passing the already-parsed DataFrame avoids that. I can remove it and rely only on the cache if you’d prefer simpler code and are okay with that first-pass cost.

SFJohnson24: good catch--missed that.  Could we clean up the implementation a little and utilize caching instead of an additional argument?  Something like:
```
@functools.lru_cache(maxsize=None)
def _get_datasets_worksheet(self) -> pd.DataFrame:
    return pd.read_excel(
        self.dataset_path,
        sheet_name=DATASETS_SHEET_NAME,
        na_values=[""],
        keep_default_na=False,
    )
    ```
    then call self._get_datasets_worksheet() inside get_raw_dataset_metadata.

File: scripts/run_validation.py [hunk 12]
```diff
@@ -160,21 +186,13 @@ def run_validation(args: Validation_args):
...
-            for dataset in datasets:
-                file_path = dataset.full_path
-                if file_path.endswith(".parquet"):
-                    continue
-                num_rows, new_file = data_service.to_parquet(file_path)
-                created_files.append(new_file)
-                dataset.full_path = new_file
-                dataset.record_count = num_rows
-                dataset.original_path = file_path
+        datasets = _get_datasets_or_raise(data_service)
```
SFJohnson24: I incorrectly said this should be in scripts--I apologize. Is there a reason this was changed and the raises were not just added to data_service.get_datasets() and instead function was made that is just get_datasets with a raise?

scripts/run_validation.py
SFJohnson24: re: this comment--what is the intention of this codeblock?  the CLI catches bad CT in get_library_metadata_from_cache() at line 165 in this file and editor never sends the request as it crashes when trying to populate the cache https://github.com/cdisc-org/cdisc-rules-engine/blob/fb8a8d5f123beaeba254e98b24604bcfb5a7f6a4/TestRule/__init__.py#L105 here

scripts/script_utils.py
SFJohnson24: I just tested and your logic in get_library_metadata_from_cache actually catches bad CT from the define--can we revert the changes below?

SFJohnson24 (changes requested):
This looks good--just a few housekeeping things:
- The logic in get_library_metadata_from_cache is quite robust and is getting the bad CT from the CLI and from define so we can revert some logic to keep our codebase clean.
- caching over a new argument--that way the caller doesnt need to remember to give a workbook, the optimization is included in the call to the function
- adding the raise to the original data_service function versus making a new function

scripts/run_validation.py
RakeshBobba03: There wasn’t a strong reason for the wrapper; it was just the approach I took at the time. Your approach is better. I’ve removed `_get_datasets_or_raise` and now call `data_service.get_datasets()` directly. The raise logic lives in the data services: `ExcelDataService.get_datasets()` and `LocalDataService.get_datasets()` both use try/except and raise the appropriate errors (`ExcelTestDataError` / `InvalidDatasetFormat`). No wrapper anymore.

PR #1627, opened 2026-02-18
Variable is null

SFJohnson24:
- this PR adds the new variable_is_empty metadata to the 2 relevant builders and updates test and docs. 

- I also had updated the variable_is_null operation .  No rule has the operation so I believe leaving the operation with the changes is correct.  It also updates the readme as the old operation was not well explained to end-users.

- I also changed the wildcard flag for CG0531 per this discussion and updated the readme https://github.com/cdisc-org/cdisc-rules-engine/discussions/1608

- I also moved the check_parameter file to engine in this PR in conjunction with https://github.com/cdisc-org/conformance-rules-editor/pull/292 to more easily update the documents on the operation/operator parameters as they are hosted in engine repo

to test: this is the logic for cg0015:
[Rule_underscores.json](https://github.com/user-attachments/files/25401996/Rule_underscores.json)

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug.yml
@@ -20,8 +20,6 @@ body:
         - FDA BR SDTM
         - FDA BR SDTMIG
         - FDA BR SEND
-        - FDA VR SDTMIG
-        - FDA VR SENDIG
         - Custom
   - type: input
     id: reference_rule_id
@@ -41,7 +39,14 @@ body:
       value: https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-
   - type: markdown
     attributes:
-      value: "In the next fields, please provide a [Minimal Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example)."
+      value: "\nIn the next fields, please provide a [Minimal Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example)."
+  - type: input
+    id: cli_command
+    validations:
+      required: true
+    attributes:
+      label: CLI Command Used (if applicable)
+      value: python core.py -s SDTM -v 3.4 -d /path/to/datasets
   - type: textarea
     id: rule
     validations:
--- a/.github/ISSUE_TEMPLATE/2-rule-feature.yml
+++ b/.github/ISSUE_TEMPLATE/2-rule-feature.yml
@@ -20,8 +20,6 @@ body:
         - FDA BR SDTM
         - FDA BR SDTMIG
         - FDA BR SEND
-        - FDA VR SDTMIG
-        - FDA VR SENDIG
         - Custom
   - type: input
     id: reference_rule_id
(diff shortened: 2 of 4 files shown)


resources/schema/rule/Operations.json
```diff
@@ -380,7 +380,7 @@
       "properties": {
         "operator": { "const": "variable_is_null" }
       },
-      "required": ["id", "operator"],
+      "required": ["id", "operator", "source"],
```
RamilCDISC: schema has source as required but the documentation mentions "evaluation" is default. if there is a default then it should not be required. Could we please reconcile these?

resources/schema/rule/Operations.md
```diff
@@ -1311,14 +1311,19 @@ Operations:
 
 ### variable_is_null
 
-Returns true if a variable is missing from the dataset or if all values within the variable are null or empty string. This operation first checks if the target variable exists in the dataset, and if it does exist, evaluates whether all its values are null or empty.
-The operation can work with both direct variable names and define metadata references (variables starting with "define_variable").
+Returns true if a variable is missing from the dataset or if all values within the variable are null or empty string. This operation first checks if the target variable exists in the dataset, and if it does exist, evaluates whether all its vaes are null or empty.
```
RamilCDISC: Could you please fix the typo 'vaes' in this line?

cdisc_rules_engine/operations/variable_is_null.py
```diff
@@ -3,19 +3,16 @@
 
 class VariableIsNull(BaseOperation):
     def _execute_operation(self):
-        # Always get the content dataframe. Similar to variable_exists check
-        dataframe = self.data_service.get_dataset(dataset_name=self.params.dataset_path)
-        if self.params.target.startswith("define_variable"):
```
RamilCDISC, line 8: The new changes does not have this check of starting with define_variable. Is this case handled by evaluation dataset branch now? Could you please explain briefly?

RamilCDISC (changes requested):

cdisc_rules_engine/operations/variable_is_null.py
SFJohnson24: if a define is contained, it will be a define rule type and those variable will be accessible via the evaluation dataset built by that define rule type.

cdisc_rules_engine/operations/variable_is_null.py
```diff
@@ -3,19 +3,16 @@
...
-            # Handle checks against define metadata
-            target_column = self.evaluation_dataset[self.params.target]
-            result = [
-                self._is_target_variable_null(dataframe, value)
-                for value in target_column
-            ]
-            return self.data_service.dataset_implementation().convert_to_series(result)
+        if self.params.source == "submission":
+            if self.params.level == "row":
+                raise ValueError("level: row may only be used with source: evaluation")
```
RamilCDISC, line 8: Could we please add a unit test covering this case ensuring ValueError is raised? We can cover other combinations too that can happen in real case.

SFJohnson24: this is done

RamilCDISC (approved):
The PR updates the wildcard and variable_is_null operation. The validation was done by: 

4. Ensuring new tests has been added.
5. Ensuring previous testing is updated where necessary.
6. Running manual validation using dev editor covering cases:

- using updated rule types that include variable_is_null
- ensuring edge case of spaces in empty column
- edge case of tabs in empty column
- edge case of empty string in the empty column
- when there is only one value in the empty column
- when all values are populated in empty column

PR #1628, opened 2026-02-18
library_variable_has_codelist new metadata

SFJohnson24:
this PR adds the new metadata requested from the parent issue to variable metadata with Library builders


resources/schema/rule/MetaVariables.json
```diff
@@ -133,6 +133,7 @@
     },
     { "const": "filename" },
     { "const": "library_variable_core" },
+    { "const": "library_variable_has_codelist" },
```
gerrycampion, line 142: this file is missing library_variable_ccode

cdisc_rules_engine/dataset_builders/define_variables_with_library_metadata.py
gerrycampion, line 1: missing updates to `test_define_variables_with_library_metadata`

SFJohnson24:
updated @gerrycampion

RamilCDISC:
@SFJohnson24 do you think it would be a good idea to add 1 or 2 unit tests focused on extraction behavior from _links. Currently the updated tests verify propagation of column only.

SFJohnson24:
> @SFJohnson24 do you think it would be a good idea to add 1 or 2 unit tests focused on extraction behavior from _links. Currently the updated tests verify propagation of column only.

this would likely be done in test_sdtm_utils and not the builder.  @RamilCDISC  Do you want to make some coverage for the _link extraction using get_variables_metadata_from_standard in a separate PR?

PR #1640, opened 2026-02-26
1558 csv data reader

alexfurmenkov:
Added support for data in .csv format. 
Implemented params reading from .env files - based on .env in data folder.

### DIFF
--- a/core.py
+++ b/core.py
@@ -384,6 +384,7 @@ def load_custom_dotenv_from_data_options(ctx, param, value):
         "CDISC TIG Use Case for scoping a TIG Validation."
         "Any of INDH, PROD, NONCLIN, or ANALYSIS."
     ),
+    envvar="USE_CASE",
 )
 @click.option(
     "-ct",
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ This will show the list of validation options.
   -uc, --use-case TEXT            Use Case for TIG Validation
                                   "INDH", "PROD", "NONCLIN", or "ANALYSIS"
                                   [required for TIG]
+                                  USE_CASE environment variable can be used to pass value.
   -ct, --controlled-terminology-package TEXT
                                   Controlled terminology package to validate
                                   against, can provide more than one


cdisc_rules_engine/services/data_readers/csv_reader.py
```diff
@@ -0,0 +1,37 @@
...
+class CSVReader(DataReaderInterface):
+    def read(self, data):
+        """
+        Function for reading data from a specific file type and returning a
+        pandas dataframe of the data.
+        """
+        raise NotImplementedError
+
+    def from_file(self, file_path):
+        with open(file_path, "r", encoding=self.encoding) as fp:
```
RamilCDISC: We will need encoding + general error handling wrapping similar to json reader here.

cdisc_rules_engine/services/csv_metadata_reader.py
```diff
@@ -0,0 +1,120 @@
...
+    def __data_meta(self):
+        first_row_df = pd.read_csv(self.file_path, encoding=self.encoding, nrows=1)
+
+        if not first_row_df.empty:
+            first_record = first_row_df.iloc[0].fillna("").astype(str).to_dict()
+        else:
+            first_record = {}
+
+        with open(self.file_path, encoding=self.encoding) as f:
+            dataset_length = sum(1 for _ in f) - 1  # subtract header
```
RamilCDISC: This can result in dataset_length = -1 for empty csv files. For those cases I guess it should be zero. So we should clamp it to zero.

cdisc_rules_engine/exceptions/custom_exceptions.py
```diff
@@ -62,6 +62,11 @@ class InvalidJSONFormat(EngineError):
     description = "JSON data is malformed."
 
 
+class InvalidCSVFormat(EngineError):
```
RamilCDISC: I think better naming for this would be InvalidCSVFile instead of format because this error is for when csv is malformed or encoding failed which makes it invalid csv file. What do you think?

core.py
RamilCDISC, line 1: I think for csv files with the route of dataset-path -dp flag we  will need some validation that user has not given metadata csv. The --data path handles this but for -dp path it just processes the file directly. it would be better to have a check here.

cdisc_rules_engine/services/data_readers/csv_reader.py
```diff
@@ -0,0 +1,50 @@
...
+        for chunk in dataset:
+            num_rows += len(chunk)
+
+            if not created:
+                chunk.to_parquet(temp_file.name, engine="fastparquet")
+                created = True
+            else:
+                chunk.to_parquet(temp_file.name, engine="fastparquet", append=True)
+
+        return num_rows, temp_file.name
```
RamilCDISC, line 64: This function always returns a file path, but file is only written when csv is not empty. An empty csv can cause to return path to a file that is not yet created. It may cause some downstream errors.

cdisc_rules_engine/services/data_services/local_data_service.py
RamilCDISC: Please update this to add csv also here as we will be supporting csv too now.

core.py
```diff
@@ -104,6 +105,44 @@ def cli():
...
+    import pandas as pd
+
+    paths = [Path(p) for p in dataset_paths]
+    tables_path = next((p for p in paths if p.name.lower() == "tables.csv"), None)
+
+    dataset_files = [
+        p for p in paths if p.name.lower() not in ("tables.csv", "variables.csv")
+    ]
+
+    if not tables_path:
```
SFJohnson24: this should return an error and cancel execution if the 2 metadata files are not found.  It should also ensure that only 1 of each is present.

core.py
```diff
@@ -104,6 +105,44 @@ def cli():
...
+        return [str(p) for p in dataset_files if p.suffix.lower() == ".csv"]
+
+    tables_df = pd.read_csv(tables_path, encoding=encoding)
+
+    if "Filename" not in tables_df.columns:
+        return [str(p) for p in dataset_files if p.suffix.lower() == ".csv"]
```
SFJohnson24: I also dont think we should be falling back if the tables.csv is incorrect-- it seems this will just cause downstream problems?  We should be throwing an InvalidCSVFile error and telling the user their metadata files is malformed, halting execution

core.py
alexfurmenkov: what do you mean by that? How can it be possible for two files with same name to sit in same directory?

and if we use -dp parameter, for examlpe

-dp datasets/dm.csv
-dp datasets_other/lb.csv

should we check that both of the directories have `tables.csv` or how we should cover this scenario? as I remember, `variables.csv` are not neccesary(if not required by the rule), but should we also check it?

cdisc_rules_engine/services/data_readers/csv_reader.py
alexfurmenkov: added empty parquet file creation in case when csv was empty. 
or should we raise value error in this scenario? should I also fix it in xpt reader?

core.py
alexfurmenkov: I think we can treat this situation like multiple -d flags. 
so in sutuation when we provide

> -dp datasets/dm.csv
> -dp datasets_other/lb.csv

we should expect at least this files to exist:

> datasets_1/dm.csv
> datasets_1/tables.csv
> datasets_1/variables.csv

and

> datasets_2/lb.csv
> datasets_2/variables.csv

cdisc_rules_engine/services/data_services/local_data_service.py
```diff
@@ -202,11 +206,16 @@ def read_metadata(
...
+                f"Please provide dataset files in SAS V5 XPT, "
+                f"Dataset-JSON (JSON or NDJSON), CSV, or Excel (XLSX) format."
             )
 
         contents_metadata = _metadata_reader_map[file_extension](
-            file_metadata["path"], file_name, encoding=self.encoding
+            file_metadata["path"],
+            file_name,
+            encoding=self.encoding,
+            variables_csv_path=self.variables_csv_path,
```
SFJohnson24, line 217: this line is causing as error in the test suite.
TypeError: DatasetXPTMetadataReader.__init__() got an unexpected keyword argument 'variables_csv_path'
this should be a conditional where if it is the csv reader, it passes the csv files to it,otherwise just the others

alexfurmenkov:
@SFJohnson24 
> .env validation argument

I've added `load_custom_dotenv` function to be called when `--dotenv-path` param is passed and did not remove original     `load_dotenv` call but modified to use `dotenv_path` parameter. Can you please clarify what kind of validation is expected? 
something like this?

```Python  
  elif dataset_path:
        if not Path(dotenv_path).exists():
            load_dotenv()
        else:
            load_dotenv(dotenv_path)
        ...
```

core.py
```diff
@@ -372,7 +483,13 @@ def _validate_no_arguments(logger) -> None:
         "is printed."
     ),
 )
-@click.option("-dxp", "--define-xml-path", required=False, help="Path to Define-XML")
+@click.option(
+    "-dxp",
+    "--define-xml-path",
+    required=False,
+    help="Path to Define-XML",
+    envvar="DEFINE",
```
SFJohnson24: this is "DEFINE_XML" in the .env

core.py
```diff
@@ -425,11 +542,26 @@ def _validate_no_arguments(logger) -> None:
         f"Supported encodings: utf-8, utf-16, utf-32, cp1252, latin-1, etc."
     ),
 )
+@click.option(
+    "-vcp",
+    "--variables-csv-path",
+    required=False,
+    help="Path to variables.csv",
+    envvar="VARIABLES_CSV",
```
SFJohnson24: there is no VARIABLES_CSV in the .env

core.py
```diff
@@ -425,11 +542,26 @@ def _validate_no_arguments(logger) -> None:
...
+)
+@click.option(
+    "-tcp",
+    "--tables-csv-path",
+    required=False,
+    help="Path to tables.csv",
+    envvar="TABLES_CSV",
```
SFJohnson24: there is no TABLES_CSV in the .env

core.py
```diff
@@ -255,20 +358,23 @@ def _validate_no_arguments(logger) -> None:
     required=True,
     default=None,
     help="CDISC standard to validate against",
+    envvar="STANDARD",
```
SFJohnson24: this is PRODUCT

core.py
```diff
@@ -289,6 +395,7 @@ def _validate_no_arguments(logger) -> None:
         "Controlled terminology package to validate against, "
         "can provide more than one"
     ),
+    envvar="CONTROLLED_TERMINOLOGY_PACKAGE",
```
SFJohnson24: this is CT

core.py
```diff
@@ -217,11 +309,20 @@ def _validate_no_arguments(logger) -> None:
...
+    required=False,
+    help="Path to .env file containing custom dotenv values",
+    callback=load_custom_dotenv,
+)
 @click.option(
     "-d",
     "--data",
     required=False,
     help=f"Path to directory containing data files ({VALIDATION_FORMATS_MESSAGE})",
+    envvar="DATA_DIR",
```
SFJohnson24: there is no DATA_DIR .env

core.py
```diff
@@ -236,6 +337,8 @@ def _validate_no_arguments(logger) -> None:
     required=False,
     multiple=True,
     help=f"Absolute path to dataset file ({VALIDATION_FORMATS_MESSAGE})",
+    envvar="DATASET_PATH",
```
SFJohnson24: there is no DATASET_PATH

core.py
```diff
@@ -320,6 +427,7 @@ def _validate_no_arguments(logger) -> None:
     "--define-version",
     type=click.Choice(["2-1", "2-0", "2.0", "2.1"]),
     help="Define-XML version used for validation",
+    envvar="DEFINE_VERSION",
```
SFJohnson24: there is no DEFINE_VERSION

core.py
```diff
@@ -338,12 +446,14 @@ def _validate_no_arguments(logger) -> None:
     "-r",
     multiple=True,
     help="Specify rule core ID ex. CORE-000001. Can be specified multiple times",
+    envvar="RULES",
```
SFJohnson24: there is no RULES

core.py
```diff
@@ -338,12 +446,14 @@ def _validate_no_arguments(logger) -> None:
...
 )
 @click.option(
     "--exclude-rules",
     "-er",
     multiple=True,
     help="Specify rule core ID to exclude, ex. CORE-000001. Can be specified multiple times",
+    envvar="EXCLUDE_RULES",
```
SFJohnson24: there is EXCLUDE_RULES

core.py
```diff
@@ -352,6 +462,7 @@ def _validate_no_arguments(logger) -> None:
     type=click.Path(exists=True, readable=True, resolve_path=True),
     help="Path to directory containing local rules.",
     multiple=True,
+    envvar="LOCAL_RULES",
```
SFJohnson24: no LOCAL_RULES

core.py
SFJohnson24: it is also not a path, it is just the name of the define located in the same directory as the .env DEFINE_XML=define_negative1.xml
The path will need to be constructed

core.py
```diff
@@ -731,6 +867,7 @@ def update_cache(
     required=False,
     help="Rule ID to get rule for.",
     multiple=True,
+    envvar="RULEID",
```
SFJohnson24: no RULEID

SFJohnson24 (changes requested):
noticed a lot of env params that dont exist:
```
PRODUCT=TIG
VERSION=1-0
SUBSTANDARD=SDTM
CT=sdtmct2020-12-18
DEFINE_XML=define_negative1.xml
```
above is a complete .env file with all possible params from the cdisc-open-rules repository.  You can hop on the `rules` branch and see what it looks like if you want

core.py
alexfurmenkov: It can be either just a filename (when define.xml is in the same directory) or a full/relative path when it's located elsewhere. This parameter allows overriding the default behavior, where the constructor looks for define.xml in the dataset folder

core.py
```diff
@@ -255,20 +356,23 @@ def _validate_no_arguments(logger) -> None:
...
     help="CDISC standard to validate against",
+    envvar="PRODUCT",
 )
 @click.option(
     "-v",
     "--version",
     required=True,
     default=None,
     help="Standard version to validate against",
+    envvar="VERSION",
```
RamilCDISC, line 367: All else looks fine to me. Could you please adda test where we check the .env interoperability. You can cover all the flags in few tests.

RamilCDISC:
@SFJohnson24 Could you please confirm if you agree with the changes made on your request. I have done manual validation and as well as code review. All seems good to me. Once you give a heads up i will add final comment and approve.

env.example
```diff
@@ -2,3 +2,5 @@ CDISC_LIBRARY_API_KEY=your_api_key_here
 DATASET_SIZE_THRESHOLD=10485760  # max dataset size in bytes to force dask implementation
 MAX_REPORT_ROWS = 10  # integer for maximum number of issues per excel sheet (plus headers) in result report.  Defaults to 10000.
 MAX_ERRORS_PER_RULE = (10, True)  # Tuple for maximum number of errors to report per rule during a validation run. Also has a per dataset flag described as second bool value in readme. example value 
+DEFINE_XML
+DATA_DIR
```
SFJohnson24, line 6: what is DATA_DIR?  Is this an artifact?

SFJohnson24: should this example file also have the other variables the .env can handle?

core.py
SFJohnson24, line 388: I tested this on TIG and it we will need envvar="USE_CASE" for this standard.  I updated the open rules to reflect USE_CASE as an .env variable

SFJohnson24:
i pushed the 2 lines up just to expedite this getting merged--looks good @alexfurmenkov

SFJohnson24 (approved):

PR #1644, opened 2026-02-27
1182: Add custom exceptions for unsupported XPT format and enhance error handling in XPTReader

alexfurmenkov:


cdisc_rules_engine/services/data_readers/xpt_reader.py
```diff
@@ -7,22 +7,30 @@
...
 
     def read(self, data):
-        df = pd.read_sas(BytesIO(data), format="xport", encoding=self.encoding)
+        df = self._read_sas(BytesIO(data), format="xport")
         df = self._format_floats(df)
         return df
 
     def _read_pandas(self, file_path):
-        data = pd.read_sas(file_path, format="xport", encoding=self.encoding)
+        data = self._read_sas(file_path, format="xport")
```
SFJohnson24, line 31: was the removal of encoding purposeful?  It is necessary functionality from https://github.com/cdisc-org/cdisc-rules-engine/pull/1456

alexfurmenkov: @SFJohnson24 check out the `self._read_sas` method (defined on line 14):

`return pd.read_sas(source, encoding=self.encoding, **kwargs)`

alexfurmenkov:
@RamilCDISC, I execute with ae.xpt file (v8 version from the zip attached in ticket) and set -l error . 
image

alexfurmenkov:
When using pd.read_sas, the argument format="xport" indicates that we expect a SAS Transport V5 file. Therefore, if there is a version mismatch while reading the SAS file, the error "Header record is not an XPORT file." is raised.

At the moment, we catch all possible errors during read_sas. In the next commit, I will narrow this down to catching only this specific error: "Header record is not an XPORT file."

I can also add unit tests for attempts to read SAS files of different versions (V5 and V8). If you have examples of test data for other versions as well, please share them and I will include them in the tests.

RamilCDISC:
@alexfurmenkov The v8 xpt test data is attached in the connected issue. You can take from there and v5 xpt are available in the test resources. i have verified, the engine does show error. 

What we can decide on is do we want for such a file skip execution for all rules or show RULE EXECUTION ERROR in the status column of the report. @SFJohnson24 What do you think?

SFJohnson24:
@RamilCDISC I think skipping will take more of a lift. If a user tries to submit bad data--the execution should be halted until they format in v5.

RamilCDISC:
@SFJohnson24 Thanks for your response. @alexfurmenkov the PR is then correct. Are you still planning to narrow down the error catch as you mentioned in your previous comment, os I should perform final validation and add closing comments.

alexfurmenkov:
Hi @RamilCDISC 
I have finished the planned changes. Please feel free to review.

Issue #1442, opened 2025-11-18
Request to add a CLI option to filter the filetype of files to process


mhungria:
**Request to add a CLI option to filter the file type of files to process from a given folder _(with -d option only)_**.
Note: discussed at the CORE meeting today 2025-11-18.

The current CLI functionality provides an error message if there is more than one file type of the file types accepted for validation. It indicates that there are multiple formats in the folder.

**
A CLI option like the one below, would address that issue:
--filetype "json"
**

Or maybe, one more generic:
--subset_files "*.json" (one can add an regex to subset both filenames and filetypes).

**Update on 2025-12-10** after a call with @alexfurmenkov : 

**1)** This is the error from an example CLI execution when more that one file types are in a folder requested for validation:

(venv) marcelinahungria@Marcelinas-MBP cdisc-rules-engine % python3 core.py validate -s sdtmig -v 3-4 -d /Users/marcelinahungria/Documents/W10X1/Clients/Internal/DIcG_git/DIcGweb/data/examples/sdtm --output /Users/marcelinahungria/Documents/W10X1/Clients/Internal/DIcG_git/DIcGweb/data/examples/sdtm/tempstats/val_reports/core_val_data_examples_sdtm_20251210 --output-format XLSX -lr /Users/marcelinahungria/Documents/W10X1/core_env/Rules/DraftRules20250825subset -l error

**[ERROR 2025-12-11 11:08:46,716 - core.py:283] - Argument --data contains more than one allowed file format (NDJSON, XPT, JSON).**
(venv) marcelinahungria@Marcelinas-MBP cdisc-rules-engine % 

Folder content (source data included in the Dataset-JSON v1.1 release, partial content below:
marcelinahungria@Marcelinas-MBP ~ % cd /Users/marcelinahungria/Documents/W10X1/Clients/Internal/DIcG_git/DIcGweb/data/examples/sdtm
marcelinahungria@Marcelinas-MBP sdtm % ls -la
total 55512
drwxr-xr-x@ 104 marcelinahungria  staff     3328 Oct 22 16:27 .
drwxr-xr-x    8 marcelinahungria  staff      256 Oct 23 14:38 ..
-rw-r--r--@   1 marcelinahungria  staff    14340 Dec 10 15:58 .DS_Store
-rw-r--r--    1 marcelinahungria  staff  6825094 Nov 13  2024 acrf.pdf
[...]

PR #1645, opened 2026-03-02
#1442 fix dataset filtering when -dp is provided

alexfurmenkov:


README.md
```diff
@@ -241,7 +241,8 @@ This will show the list of validation options.
                                   "[████████████████████████████--------]
                                   78%"is printed.
   -jcf, --jsonata-custom-functions Pair containing a variable name and a Path to directory containing a set of custom JSONata functions. Can be specified multiple times
-  -e, --encoding TEXT            File encoding for reading datasets. If not specified, defaults to utf-8. Supported encodings: utf-8, utf-16, utf-32, cp1252, latin-1, etc.
+  -e, --encoding TEXT             File encoding for reading datasets. If not specified, defaults to utf-8. Supported encodings: utf-8, utf-16, utf-32, cp1252, latin-1, etc.
+  -ft, --filetype TEXT            File extension to filter datasets. Has higher priority then --dataset-path parameter.
```
RamilCDISC: A small typo. I believe it should be higher priority **'than'** instead of **'then'** --dataset-path parameter

RamilCDISC (changes requested):
In cases when -dp points to a file that is not mentioned by -ft then the message should state the reason why no files were processed.

Right now if I mention only one dataset file using -dp and use -ft to point to a different type of file. The message is unclear. The excel file mentioned with -dp flag is valid and does exist in the system. The reason it was not used because -ft flag was specified to use only json files. 
```
python core.py validate -s sdtmig -v 3.4 -dp tests/resources/report_test_data/test_datasets.xlsx -ft json
[ERROR 2026-03-02 14:29:57,900 - core.py:184] - No valid dataset files provided.
Supported formats: SAS V5 XPT, Dataset-JSON (JSON or NDJSON), or Excel (XLSX)
```

SFJohnson24 (approved):
PR correctly addresses issue

Issue #1248, opened 2025-07-17
Operation "define_variable_metadata" doesn't work for Excel data


HajimeShimizu:
define_variable_metadata  doesn't function when data is in Excel format. (It works when data is in XPT format).

- Internally, file path to define.xml is set as "define.xml" (just file name, without any additional path). I confirmed it in Log file.
- Path to define.xml file is generated from path to dataset.
- In Excel format reader, path to dataset is just file name (ex. ae.xpt.  not like "testdata/ae.xpt")

Some adjustment is required.

gerrycampion:
@HajimeShimizu We are not able to reproduce this behavior. Can you please provide more information, like the commandline command, rule, test data, and error output?

HajimeShimizu:
Thanks for reply. Please give me some time. I'll collect relevant information and/or files.

HajimeShimizu:
Hi, I checked the issue. Hope this helps us.

Version CORE v0.14.0
Test data is in excel file.
Define.xml is provided. (v2.1)

Command：
core.py validate -s sdtmig -v 3-2 -o report -of JSON -lr test/sample.yml -cs -dxp test/define.xml -dp test/data.xlsx

Note;
(1) Excel file and define.xml is stored in the same folder. Both path is specified in the command.
(2) When I specify .xpt file it works.

Error Message：
```
[ERROR 2026-03-03 15:58:12,152 - console_logger.py:50] - error in operation define_variable_metadata: [Errno 2] No such file or directory: 'define.xml'
Traceback (most recent call last):
  File "....\core\cdisc_rules_engine\operations\base_operation.py", line 72, in execute
    result = self._execute_operation()
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\core\cdisc_rules_engine\operations\define_variable_metadata.py", line 35, in _execute_operation
    define_contents = self.data_service.get_define_xml_contents(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\core\cdisc_rules_engine\services\data_services\base_data_service.py", line 79, in inner
    dataset = func(*args, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^
  File "...\core\cdisc_rules_engine\services\data_services\excel_data_service.py", line 193, in get_define_xml_contents
    with open(dataset_name, "rb") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1649, opened 2026-03-06
#1248 Specify define xml path in operations

alexfurmenkov:
Added define xml path parameter to operation.
Found an issue in ExcelDataService ruining define.xml loading, because we use one path to look for both data and define.xml. Also, when we use ExcelDataService, dataset metadata doesn't contain real path to the *.xlsx file.


tests/QARegressionTests/test_Issues/test_CoreIssue1248.py
```diff
@@ -0,0 +1,54 @@
...
+            "3-2",
+            "-of",
+            "JSON",
+            "-lr",
+            "tests/resources/CoreIssue1248/sample.yml",
+            "-cs",
+            "-dxp",
+            "tests/resources/CoreIssue1248/define_subfolder/define.xml",
+            "-dp",
+            "tests/resources/CoreIssue1248/relrec.json",
```
RamilCDISC: Better would be to use excel dataset here instead of json. The PR updates are for excel dataset. Using json will not test the updated changes.

tests/QARegressionTests/test_Issues/test_CoreIssue1248.py
RamilCDISC, line 1: we should add a test with no -dxp flag supplied to verify the case of getting define from dataset path also.

cdisc_rules_engine/services/data_services/excel_data_service.py
```diff
@@ -152,7 +152,8 @@ def get_raw_dataset_metadata(
                 os.path.getmtime(self.dataset_path)
             ).isoformat(),
             filename=dataset_name,
-            full_path=dataset_name,
+            full_path=self.dataset_path,
+            # full_path=dataset_name,
```
RamilCDISC: Could you please remove this old commented out code?

cdisc_rules_engine/operations/define_variable_metadata.py
```diff
@@ -32,8 +33,15 @@ def _execute_operation(self):
                     ...
                 }
         """
+        define_path = (
+            self.params.define_xml_path
+            if self.params.define_xml_path
+            else os.path.join(self.params.directory_path, DEFINE_XML_FILE_NAME)
+        )
+        if not os.path.exists(define_path):
+            raise FileNotFoundError("Define XML file %s not found", define_path)
```
RamilCDISC: This string formatting is incorrect. Use `f"Define XML file {define_path} not found"` instead.

cdisc_rules_engine/services/data_services/excel_data_service.py
alexfurmenkov: I’ll revert it back to `full_path=dataset_name` for now.
Ideally, it should be `full_path=self.dataset_path` so the name matches and the report shows the actual path.

However, changing this will break the current code because `get_dataset `expects a dataset name, while all service clients are currently sending `full_path`.

Possible solutions:
1. Detect the service type and send the filename field instead of `full_path`. This would work but doesn’t feel very elegant.
2. Change `get_dataset `to accept a dataset metadata object instead of a path parameter. Then each service can decide what field from the metadata it needs to retrieve the dataset.

tests/QARegressionTests/test_Issues/test_CoreIssue1248.py
alexfurmenkov: This will not work without the `-dxp` flag until changes are made to `ExcelDataService` related to [comment](https://github.com/cdisc-org/cdisc-rules-engine/pull/1649#discussion_r2918399957)

or changes to `self.rule_processor.perform_rule_operations` call should be made, so `dataset_metadata.full_path` containing in our case `relrec.xpt` value will pass something else

cdisc_rules_engine/services/data_services/excel_data_service.py
RamilCDISC: The AC in the attached issue requires solution of the bug in excel dataset use path. i think we should resolve the issue in this PR before merge. You can discuss with @SFJohnson24 and @gerrycampion to decide what solution to use.

cdisc_rules_engine/operations/define_variable_metadata.py
```diff
@@ -32,8 +33,15 @@ def _execute_operation(self):
...
+            if self.params.define_xml_path
+            else os.path.join(self.params.directory_path, DEFINE_XML_FILE_NAME)
+        )
+        if not os.path.exists(define_path):
+            raise FileNotFoundError(f"Define XML file {define_path} not found")
         define_contents = self.data_service.get_define_xml_contents(
-            dataset_name=os.path.join(self.params.directory_path, DEFINE_XML_FILE_NAME)
+            dataset_name=define_path
         )
         define_reader = DefineXMLReaderFactory.from_file_contents(define_contents)
```
gerrycampion: This chunk should be moved to a BaseOperation function so that it can be reused

RamilCDISC (approved):
The PR resolves the bug related to reading of define xml when using excel dataset. The PR was validated by:

1. Review the PR for any unwanted code or comments.
2. Review the PR in accordance  with AC.
3. Ensure all testing pass.
4. Ensure relevant testing is updated.
5. Validate manually using CLI to ensure: 

- The same command and data from AC does not give an error now.
- The case when define and excel are in same dataset with -dxp flag
- The case when define and excel are not in same dataset with dxp filag
- The above two cases repeated without mentioning the -dxp flag
- Ensuring other formats still work fine

PR #1653, opened 2026-03-06
Resolve rule execution errors

gerrycampion:
- fix dy operation to return a SKIPPED DomainNotFoundError when DM missing
- fix domain wildcard replacement to use domain instead of dataset unsplit name
- simplify args for rule_processor.perform_rule_operations to use DatasetMetadata instead of individual args
- reraise Operations DomainNotFoundError and KeyError to trickle up as SKIPS
- Better handling of empty datasets and treat EMPTY_DATASET as a SKIP
- Change ignored data files from warning to info as this is a pretty normal occurrence in a standard study package
- Find dataset, class, and dataset variables from the model if dataset is in the model but not in the ig
- Fix some instances of confusion between domain and dataset
- Move sdtm-related utils from utils to sdtm_utils
- Move function from utils to dataset_metadata property
- replace extract_file_name_from_path_string with built-in os.path.basename
- fix distinct's call to data service referenced datasets
- fix relrec merge on columns with differing datatypes
- fix relrec merge to trigger skip when it results in no records
- clarified SDTMDatasetMetadata wildcard_replacement variable and fixed wildcard replacement instances
- fix merge on define xml dataset metadata

To compare the Execution Errors before and after, see these reports:
[Report before](https://github.com/user-attachments/files/25986528/CORE-Report-2026-03-13T18-23-26.xlsx)
[Report after](https://github.com/user-attachments/files/25986479/CORE-Report-2026-03-13T18-04-03.xlsx)

[CORE Test Suite Updates](https://github.com/cdisc-org/CORE_Test_Suite/pull/90)

### DIFF
--- a/cdisc_rules_engine/utilities/data_processor.py
+++ b/cdisc_rules_engine/utilities/data_processor.py
@@ -17,6 +17,7 @@ from cdisc_rules_engine.services.data_services import (
     DataServiceFactory,
     DummyDataService,
 )
+from cdisc_rules_engine.exceptions.custom_exceptions import PreprocessingError
 from cdisc_rules_engine.utilities.utils import (
     search_in_list_of_dicts,
 )
@@ -339,7 +340,7 @@ class DataProcessor:
             validation_keys.append(idvar_for_qnam)
             grouped = qnam_check.groupby(validation_keys).size()
             if (grouped > 1).any():
-                raise ValueError(
+                raise PreprocessingError(
                     f"Multiple records with the same QNAM '{qnam}' match a single parent record"
                 )
         return result_dataset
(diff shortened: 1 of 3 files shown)


cdisc_rules_engine/operations/distinct.py
```diff
@@ -70,9 +70,9 @@ def get_existing_column_names(group):
 
     def _get_referenced_datasets(self):
         referenced_datasets = {}
-        for dataset_meta in self.data_service.data:
-            dataset = self.data_service.get_dataset(dataset_meta.filename)
-            referenced_datasets[dataset_meta.name] = dataset
+        for dataset_metadata in self.data_service.get_datasets():
+            dataset = self.data_service.get_dataset(dataset_metadata.name)
```
SFJohnson24: I believe this needs to stay as 
dataset = self.data_service.get_dataset(dataset_meta.filename)

Looking at the interface--i think it is another case of name and filename getting mixed.  It needs the file extension, otherwise it returns an empty dataframe.  I tested this on cg0370 and was not getting the correct result as it was not finding the LB dataset referenced in the CO dataset (was returning nothing for the distinct operation)

we should change all the get_datasets parameters in the interface and excel/dummy/local/usdm data services to dataset_path to avoid this

SFJohnson24 (changes requested):
Nice changes to resolve several issues, organize code, and use native functionality/metadata properties over unneeded functions.
PR preserves relrec merge functionality--tested cg0602
Correctly resolves execution error vs. skip status from parent issue

Only found one issue that needs to be addressed.  Please see comment

cdisc_rules_engine/operations/distinct.py
gerrycampion: Fixed this and renamed dataset_metadata.dataset_name to data_service_identifier to make it a bit clearer. I will do more refactoring and get_datasets parameter renaming in a follow-up pr

cdisc_rules_engine/dataset_builders/base_dataset_builder.py
```diff
@@ -160,17 +160,14 @@ def get_define_xml_variables_metadata(self) -> List[dict]:
...
-        if not domain:
-            return []
-        return define_xml_reader.extract_variables_metadata(domain_name=domain)
+        domain = (
+            self.dataset_metadata.domain
+            or self.dataset_metadata.rdomain
+            or self.dataset_metadata.unsplit_name
+        )
+        return define_xml_reader.extract_variables_metadata(
+            domain_name=domain, name=self.dataset_metadata.name
```
RamilCDISC, line 171: I might be misunderstanding this but before this looked like a lookup only using domain and now we also require dataset name. Could that cause split dataset lookups to fail if in define xml is keyed by base domain?

gerrycampion: This was to fix an issue with CORE-001081 (see before-report) where define metadata was not being extracted for RELREC. Here is what I would expect to see for Name/Domain when doing an itemgroupdef lookup.

| Name | Domain|
| --- | --- |
| AE | AE |
| QSPH | QS |
| RELREC  | |
| SUPPDM | DM |

I added additional fixes now so that RELREC only needs a name and not domain.  This corresponds with what is in the Define-xml specification:

Image

RamilCDISC: Thanks for the clarification. Do you think we should a test for these define-xml lookup cases. So the expected matching behavior is clearly covered?

gerrycampion: Done

Issue #1654, opened 2026-03-06
Regex parameter is not working for operator is_inconsistent_across_dataset


marisawyckmans:
### Feature Description

Regex parameter is described in the reference guide, and can be used in combination with for example does_not_equal_string_part or is_not_unique_set operator.

Currently, it doesn't work in combination with is_inconsistent_across_dataset operator. We want to use it in a custom rule to extract date part from --DTC and check that date component is consistent within a subject-visit combination. In below example, the WEEK1 records should not generate output. Currently, they are giving output because time component is different. Using the regex parameter, we can extract date part only and ignore time component.

<img width="505" height="432" alt="Image" src="https://github.com/user-attachments/assets/76a448c8-51e3-4334-bb51-9b2f5d58a8f8" />

Thanks in advance for checking if regex parameter can also be allowed for the is_inconsistent_across_dataset operator

PR #1658, opened 2026-03-10
#1654 added regex to is_inconsistent_across_dataset operation

alexfurmenkov:


tests/unit/test_check_operators/test_value_set_checks.py
```diff
@@ -244,6 +244,47 @@ def test_is_inconsistent_across_dataset(
     assert result.equals(df.convert_to_series(expected_result))
 
 
+def test_is_inconsistent_across_dataset_regex():
```
RamilCDISC: could you please cover multiple cases such as regex missing, edge cases around regex shape etc.

cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -1162,6 +1165,17 @@ def is_inconsistent_across_dataset(self, other_value):
                     grouping_cols.append(col)
         df_check = self.value[grouping_cols + [target]].copy()
         df_check = df_check.fillna("_NaN_")
+        if regex:
+            pattern = re.compile(regex)
+            if pattern.groups == 0:
+                regex = f"({regex})"
+            extracted = df_check[target].astype(str).str.extract(regex, expand=False)
```
RamilCDISC: The documentation mentions that only first capture group will be used but here whole regex is applied, which will return multiple columns if there are more than 1 groups. Could you please reconcile these? Also could you then cover the case of multiple groups and invalid regex in tests.

alexfurmenkov:
@RamilCDISC hi, regarding last message, can you please re-run your test but with the fixed rule? I've made a mistake providing you examle rule. according to documentation,operation should look like that:

```YAML
- name: "BGSTRESU"
  operator: is_inconsistent_across_dataset
  value: "USUBJID"
```

 or with multiple grouping values
```YAML
- name: "--STRESU"
  operator: is_inconsistent_across_dataset
  value:
    - "--TESTCD"
    - "--CAT"
    - "--SCAT"
    - "--SPEC"
    - "--METHOD" 
```
here is updated rule 
[Rule.yml](https://github.com/user-attachments/files/26121853/Rule.yml)
and the report after running the rule
[CORE-Report-2026-03-19T17-34-49.json](https://github.com/user-attachments/files/26121893/CORE-Report-2026-03-19T17-34-49.json)

cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -1162,6 +1166,21 @@ def is_inconsistent_across_dataset(self, other_value):
                     grouping_cols.append(col)
         df_check = self.value[grouping_cols + [target]].copy()
         df_check = df_check.fillna("_NaN_")
+        if regex:
+            try:
+                pattern = re.compile(regex)
+            except re.error:
+                logger.error("Invalid regex: %s. It will be ignored.", regex)
+                regex, pattern = r".*", re.compile(r".*")
```
gerrycampion: I think an invalid regex should trickle up as an execution error so that the rule author has to fix it.

PR #1663, opened 2026-03-11
whitespace column header removal

SFJohnson24:
To prevent issues like the one experienced in the attached issue, I added logic to strip column headers of their whitespace.  Since we glean so much metadata--from column prefixes to domain from correctly IDing these columns, removing whitespace from them when reading the files is important
this was added to xlsx and json readers.  since xpt are made with tooling, padded column names are not a concern. NDJSON and Dataset-JSON readers are also unaffected as their column names are derived from schema-validated "name" fields rather than raw file headers.

### DIFF
--- a/.github/workflows/check-schema-markdown.yml
+++ b/.github/workflows/check-schema-markdown.yml
@@ -11,6 +11,7 @@ on:
       - main
 permissions:
   contents: read
+  pull-requests: write
 
 jobs:
   check-schema-markdown:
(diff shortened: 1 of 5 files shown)


RamilCDISC (approved):
The PR updates the excel and json dataset readers to handle whitespaces in headers. The updated code was validated by: 

4. Ensuring relevant testing is updated.
5.  Verifying the cases covered in new tests.
6. The passing of test using the white spaces json and excel file validates the PR.

PR #1666, opened 2026-03-13
688: update tests

alexfurmenkov:


alexfurmenkov:
The tests from test_Issues are now running, but some tests that cannot be quickly modified to work are being skipped.

.github/workflows/test-unit.yml
```diff
@@ -17,9 +17,11 @@ jobs:
...
         run: |
           pip install -r requirements-dev.txt
+          pip install -e .
       - name: Running Tests
         env:
           CDISC_LIBRARY_API_KEY: fakekey12341234
         run: |
           python -m pytest tests/unit/ --cov=cdisc_rules_engine --cov-fail-under=75
           python -m pytest tests/QARegressionTests/
+          pytest -m regression tests/QARegressionTests/test_Issues
```
RamilCDISC: Could you please update the `pytest -m regression tests/QARegressionTests/test_Issues` to `python -m pytest -m regression tests/QARegressionTests/test_Issues` for consistency

RamilCDISC:
One reason for not adding these tests as part of CI before was that these tests are slow. The time for CI-unit tests will increase a lot. If @gerrycampion  and @SFJohnson24  say the time increase is acceptable the PR is good for QA validation. My initial intent was that developers can run these if they want to test locally.

PR #1670, opened 2026-03-16
error handling and removed if conditional

SFJohnson24:
This PR removes the If conditional for the pivot logic that is needed for other supp domains to collapse, previously limited to SUPPDM only. After pivoting QNAM values into columns, process_supp produces one sparse row per original QNAM record.  The groupby/agg collapse is required for all SUPP domains to produce a single wide row per unique key.

I tested this on cg0371, fb0101, FB5801, and CG0531 for their various supp conditions.

  I also added logic for when common keys (contained in both parent and supp) have 0 alignment--a preprocessingerror is thrown to prevent silent failing, and raise the issue in the report.


SFJohnson24:
@RamilCDISC sorry I was doing A/B testing and forgot to remove the if statement.  removed it

cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -480,3 +486,19 @@ def column_metadata_equal_to_define_and_library(
...
+    ):
+        for key in common_keys:
+            left_values = set(left_dataset[key].dropna().unique())
+            right_values = set(right_dataset[key].dropna().unique())
+            if left_values and right_values and left_values.isdisjoint(right_values):
+                raise PreprocessingError(
+                    f"SUPP merge key '{key}' has no overlapping values between "
+                    f"parent dataset and SUPP dataset. "
+                    f"Parent values: {left_values}, SUPP values: {right_values}."
+                )
```
gerrycampion: 1. I don't think this is the way to sufficiently check this. For example, you could have common keys of STUDYID, USUBJID with these records:

Parent Dataset
| STUDYID | USUBJID  |
| ------- | -------- |
| STUDY1  | SUBJECT1 |
| STUDY2  | SUBJECT2 |


SUPP Dataset
| STUDYID | USUBJID  |
| ------- | -------- |
| STUDY1  | SUBJECT2 |
| STUDY2  | SUBJECT1 |

This method would validate there are overlaps even though none of the records will merge.

2. What are we trying to accomplish by having this check? We already have CG0371 which should partially check for proper supp merges. I'm afraid that putting a validation like this will interfere with rules like that.

cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -351,14 +358,13 @@ def process_supp(supp_dataset):
...
-            group_cols = [c for c in supp_dataset.columns if c not in excluded_columns]
-            supp_dataset = PandasDataset(
-                supp_dataset.data.groupby(group_cols, dropna=False, as_index=False).agg(
-                    lambda x: (x.dropna().iloc[0] if not x.dropna().empty else pd.NA)
-                )
+        excluded_columns = list(supp_dataset["QNAM"].unique()) + columns_to_drop
+        group_cols = [c for c in supp_dataset.columns if c not in excluded_columns]
+        supp_dataset = PandasDataset(
+            supp_dataset.data.groupby(group_cols, dropna=False, as_index=False).agg(
+                lambda x: (x.dropna().iloc[0] if not x.dropna().empty else pd.NA)
```
gerrycampion, line 366: please add a unit test for a test that failed with the previous logic but passes with this new logic

cdisc_rules_engine/utilities/data_processor.py
SFJohnson24: While validating the if condition could be removed, I saw the test data for FB0101 had no overlap of the merge keys so the merge completed but the Supp data was not added to the parent.  I should have checked if there was a conformance rule for this case.

cdisc_rules_engine/utilities/data_processor.py
SFJohnson24: added  test_data_processor_suppae_multiple_qnams.  It is an AE dataset that wouldnt be pivoted with the past If logic.  It has multiple qnam so it makes sure the pivot occurs and 1 row is returned (previously would be 2 rows)

cdisc_rules_engine/utilities/data_processor.py
gerrycampion: So if there are no overlap of merge keys, will FB0101 give an issue for each of the AE records as expected?

SFJohnson24: the rule is still in draft but you would likely get an execution error/skip depending on the logic.  The supp wont be merged with the parent so AETRTEM will be abscent from the analysis dataset-- this would cause an error / a skip if logic is added to the rule checking if AETRTEM exists

PR #1671, opened 2026-03-19
make library_column_order a shared parent class as a move toward consistency

gerrycampion:
made library_column_order a shared parent class for expected_variables, permissible_variables, required_variables, and get_dataset_filtered_variables. Achieved by adding optional filter params for library_column_order. 
This is to reduce inconsistent handling of variable_metadata and wildcard replacement.

### DIFF
--- a/cdisc_rules_engine/operations/library_column_order.py
+++ b/cdisc_rules_engine/operations/library_column_order.py
@@ -27,7 +27,7 @@ class LibraryColumnOrder(BaseOperation):
 
         # Filter variables based on the specified criteria
 
-        if self.params.key_name is not None and self.params.key_value is not None:
+        if self.params.key_name:
             variables_metadata = [
                 var
                 for var in variables_metadata


cdisc_rules_engine/operations/library_column_order.py
```diff
@@ -16,13 +15,28 @@ def _execute_operation(self):
...
-            var["name"].replace("--", self.params.domain) for var in variables_metadata
-        ]
-        return list(OrderedDict.fromkeys(variable_names_list))
+        # Filter variables based on the specified criteria
+
+        if self.params.key_name and self.params.key_value:
+            variables_metadata = [
+                var
+                for var in variables_metadata
+                if var.get(self.params.key_name) == self.params.key_value
```
RamilCDISC, line 34: Previously for here is `core `was missing the code would fall back to "PERMISSIBLE" but here it would become None. Is this behavior change intended?

cdisc_rules_engine/operations/library_column_order.py
```diff
@@ -16,13 +15,28 @@ def _execute_operation(self):
...
         variables_metadata: List[dict] = self._get_variables_metadata_from_standard()
 
-        # create a list of variable names in accordance to the "ordinal" key
-        variable_names_list = [
-            var["name"].replace("--", self.params.domain) for var in variables_metadata
-        ]
-        return list(OrderedDict.fromkeys(variable_names_list))
+        # Filter variables based on the specified criteria
+
+        if self.params.key_name and self.params.key_value:
```
RamilCDISC: This might not be a big issue for our case but if any chance key-value be 0 or False then this comparison will not work as intended. Could the better approach be: `if self.params.key_name is not None and self.params.key_value: is not None` ?

If you think these values are not possible here we can skip the change.

cdisc_rules_engine/operations/library_column_order.py
gerrycampion: This is the reason for the updates in sdtm_utilities. Does that resolve your concern?

cdisc_rules_engine/operations/library_column_order.py
gerrycampion: The values can be an empty string. I changed to only check truthiness of key_name. This way, if key_name is a non-empty string, we can do the filter. If key_value is missing, it will be treated as a filter for empty string.

PR #1673, opened 2026-03-25
1305: Fix type error in row number assignment to handle NaN values

### LINKED ISSUE #1305: Value Check against Define XML Variable fires Error
CORE: v0.12.0

Rule Type "Value Check against Define XML Variable" fires following error when variable(s) are not in dataset, but found in define.xml.

cdisc_open_rules_engine/model/actions.py
ValueError: cannot convert float NAN to integer

(1) This is not seen in v0.10.0
(2) We want to execute check like "variables in define.xml is not found in actual dataset". In such case, users want to merge this kind of incomplete combination of define and dataset

alexfurmenkov:


RamilCDISC (commented):
Could you please add a description. I looked at connected issue and the title of the PR. I see you found a problem and did a fix. If you could add a small description of observed problem and what fix you applied for it, that would make understanding the PR easier for review.

alexfurmenkov:
Hi @RamilCDISC, 
The file that Hajime shared in the task did not reproduce the issue. I started trying to reproduce the problem myself, and I was able to do so - if I set data_contents_long_df['variable_name'] = '' in the debugger, the issue described by Hajime is reproduced.

RamilCDISC (approved):
The PR adds a null check for robustness. The connected ticket issue was not reproduced. The PR enhances a check by adding a null check on top.

SFJohnson24 (approved):

Issue #1652, opened 2026-03-06
wildcard is not working for grouping_variables


marisawyckmans:
### Standard

FDA BR SDTMIG

### Reference Rule ID(s)

FB3001

### Conformance Rule ID(s) (if published in CORE)

CORE-000699

### JIRA Ticket

https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-9309

### CLI Command Used (if applicable)

NAP

### Rule YAML

```YAML
Authorities:
  - Organization: FDA
    Standards:
      - Name: SDTMIG
        References:
          - Citations:
              - Cited Guidance: Standard units should be consistent within the same assessment
                  (having the same --TESTCD, --CAT, --SCAT, --SPEC, --METHOD
                  values).
                Document: FDA
                Section: FDAB030
            Origin: FDA Business Rules
            Rule Identifier:
              Id: FB3001
              Version: '1'
            Version: '1.5'
        Version: '3.2'
      - Name: SDTMIG
        References:
          - Citations:
              - Cited Guidance: Standard units should be consistent within the same assessment
                  (having the same --TESTCD, --CAT, --SCAT, --SPEC, --METHOD
                  values).
                Document: FDA
                Section: FDAB030
            Origin: FDA Business Rules
            Rule Identifier:
              Id: FB3001
              Version: '1'
            Version: '1.5'
        Version: '3.3'
      - Name: SDTMIG
        References:
          - Citations:
              - Cited Guidance: Standard units should be consistent within the same assessment
                  (having the same --TESTCD, --CAT, --SCAT, --SPEC, --METHOD
                  values).
                Document: FDA
                Section: FDAB030
            Origin: FDA Business Rules
            Rule Identifier:
              Id: FB3001
              Version: '1'
            Version: '1.5'
        Version: '3.4'
      - Name: SENDIG
        References:
          - Citations:
              - Cited Guidance: Standard units should be consistent within the same assessment
                  (having the same --TESTCD, --CAT, --SCAT, --SPEC, --METHOD
                  values).
                Document: FDA
                Section: FDAB030
            Origin: FDA Business Rules
            Rule Identifier:
              Id: FB3001
              Version: '1'
            Version: '1.5'
        Version: '3.0'
      - Name: SENDIG
        References:
          - Citations:
              - Cited Guidance: Standard units should be consistent within the same assessment
                  (having the same --TESTCD, --CAT, --SCAT, --SPEC, --METHOD
                  values).
[...]

PR #1674, opened 2026-03-25
#1652 wildcard replacement with domain on grouping_variables

alexfurmenkov:
added wildcard replacement in `generate_targeted_error_object`  for group sensitivity.

### DIFF
--- a/tests/unit/test_actions.py
+++ b/tests/unit/test_actions.py
@@ -106,16 +106,153 @@ def test_targeted_error_object_with_dataset_sensitivity():
 
 
 @pytest.mark.parametrize(
-    "grouping_variables, wildcard_replacement, expected_column",
+    "grouping_variables, expected_column, df, expected_groups",
     [
-        # Original case: explicit column name, no wildcard
-        (["SETCD"], None, "SETCD"),
-        # Wildcard case: "--PARMCD" replaced with domain "TX" → "TXPARMCD"
-        (["--PARMCD", "SETCD"], "TX", "SETCD"),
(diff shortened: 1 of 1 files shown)


tests/unit/test_actions.py
```diff
@@ -105,11 +105,22 @@ def test_targeted_error_object_with_dataset_sensitivity():
...
 
 
-def test_targeted_error_object_with_group_sensitivity():
+@pytest.mark.parametrize(
+    "grouping_variables, wildcard_replacement, expected_column",
+    [
+        # Original case: explicit column name, no wildcard
+        (["SETCD"], None, "SETCD"),
+        # Wildcard case: "--PARMCD" replaced with domain "TX" → "TXPARMCD"
+        (["--PARMCD", "SETCD"], "TX", "SETCD"),
```
RamilCDISC: The test case is still covering SETCD not '--PARMCD' so wildcard replacement code is not being tested fully. Could you please add a case where expected is TXPARMCD.

SFJohnson24: agree with Ramil, would like a test where it is grouped only on --PARMCD
would also like to see TXPARMCD varied within SETCD for the test where you group on both to show that the grouping actually is contributing

RamilCDISC (approved):
The PR adds wild card replacement with domain on grouping variables. The updates were validated by: 

3. Reviewing the updated tests for coverage. 
4. Ensuring all tests pass.
5. Running manual validation on using dev editor with positive dataset
6. Running manual validation on dev editor with negative dataset.

Issue #1660, opened 2026-03-11
SUPP dataset merge by numeric variable


HajimeShimizu:
### Feature Description

CORE: 0.14.0

SUPP merge doesn't work in following case:
(1) Data is in Excel Format
(2) Merge by --SEQ variable (numeric)

Numeric data has ".0" in pandas dataframe in this case.
```
  STUDYID DOMAIN USUBJID SPDEVID AESEQ  ... 
0    TEST     AE    A001    None     1.0  ...   
1    TEST     AE    A001    None     2.0  ... 
```
On the other hand, IDVARVAL variable is character type.  
```
  STUDYID RDOMAIN USUBJID IDVAR IDVARVAL  ... 
0    TEST     AE    A001    AESEQ     1  ...   
1    TEST     AE    A001    AESEQ     2  ... 
```

"def merge_pivot_supp_dataset" handles this merge ([cdisc_rules_engine](https://github.com/cdisc-org/cdisc-rules-engine/tree/main/cdisc_rules_engine)/[utilities](https://github.com/cdisc-org/cdisc-rules-engine/tree/main/cdisc_rules_engine/utilities)/data_processor.py)

I see the following, and this code recognize "1" vs "1.0" is unequal.

```
            if not is_blank:
                common_keys.append(dynamic_key)
                current_supp = right_dataset.rename(columns={"IDVARVAL": dynamic_key})
                current_supp = current_supp.drop(columns=["IDVAR"])
                left_dataset[dynamic_key] = left_dataset[dynamic_key].astype(str)
                current_supp[dynamic_key] = current_supp[dynamic_key].astype(str)

```

Maybe normalization of numeric variable data resolve this issue. Create a new function like this...
```
        def normalize_numeric_key(x):
            if pd.isna(x):
                return x
            try:
                f = float(x)
                if f.is_integer():
                    return str(int(f))
                return str(f)
            except (ValueError, TypeError):
                return str(x)
```
... and apply for left dataset
```
            if not is_blank:
                common_keys.append(dynamic_key)
                current_supp = right_dataset.rename(columns={"IDVARVAL": dynamic_key})
                current_supp = current_supp.drop(columns=["IDVAR"])
                left_dataset[dynamic_key] = left_dataset[dynamic_key].apply(normalize_numeric_key)
                current_supp[dynamic_key] = current_supp[dynamic_key].astype(str)
```

gerrycampion:
[...]

PR #1676, opened 2026-03-27
#1660 added numeric fields normalization for supp datasets

alexfurmenkov:


cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -45,6 +45,16 @@ def convert_float_merge_keys(series: pd.Series) -> pd.Series:
         """
         return series.astype(float, errors="ignore").astype(str)
 
+    @staticmethod
+    def normalize_numeric_key(series: pd.Series) -> pd.Series:
```
RamilCDISC: In connected ticket @SFJohnson24 recommended moving this function in utils so both check operators and preprocessing logic can access the same function. Could you please update the code accordingly?

alexfurmenkov: should we keep both functions? since normalize_numeric_key applied to series and custom_str_conversion used per value? and use cases are a bit different...

RamilCDISC: @SFJohnson24 what is your recommendation for this please?

SFJohnson24: I still thinking moving custom_str_conversion to utils and utilizing that makes the most sense.  Custom_str_conversion already handles the exact problem this feature describes--normalizing numeric string representations before comparison.  I think we should harmonize the 2 as a single, reusable utils function instead of introducing a duplicate implementation.

RamilCDISC: @SFJohnson24 The PR looks good to me now. Could you please confirm if your requested change is fully addressed. I will do a final validaiton and approve after that.

SFJohnson24 (approved):
This PR correctly meets AC--moves string conversion to general util; does numeric check--merges on string key and restores original datatype to key column afterwards.

PR #1679, opened 2026-04-02
1198: Add variable_max_size to variables metadata for enhanced data insights

alexfurmenkov:


cdisc_rules_engine/dataset_builders/variables_metadata_dataset_builder.py
RamilCDISC, line 1: Could you please add testing for this new meta variable covering edge cases?

cdisc_rules_engine/dataset_builders/variables_metadata_dataset_builder.py
```diff
@@ -12,7 +13,44 @@ def build(self):
...
-        return self.data_service.get_variables_metadata(
+        # Get basic variable metadata
+        variables_metadata = self.data_service.get_variables_metadata(
             dataset_name=self.dataset_path, datasets=self.datasets, drop_duplicates=True
         )
+
+        # Check if the rule requires variable_max_size
+        if (
+            self.rule
+            and self.rule.get("output_variables")
```
SFJohnson24: i think we want to change the conditions for this--a rule using variable_max_size may not put it in the output variables and cause issue downstream.
We would want to iterate through conditions, looking for variable_max_size--this shouldnt be too costly since most rules dont have too many conditions.  We will not know the shape of conditions though--could be any/all, nested so we will need recursion

alexfurmenkov: @SFJohnson24 Could you check the new logic?

alexfurmenkov:
Hi, It's the rule and the negative dataset.
[1198.zip](https://github.com/user-attachments/files/26671254/1198.zip)

RamilCDISC:
I have completed my review. But before approval we need to update the documentation to add this new variable. Once updated I will approve and final comment.

PR #1680, opened 2026-04-03
1316: New rule - value against library metadata

### LINKED ISSUE #1316: Rule blocked: SEND296
For SEND296, where we need to raise an error when a dataset **value is not in the CDISC-CT** (according to the **CDISC Library** - the define.xml should not be involved) for variables with non-extensible codelists, we may need a new "rule type" e.g. named "Value Check against Library Metadata" which is similar to the existing "Value Check against Define XML Variable".

alexfurmenkov:
dataset builder for new rule type.

test dataset with joined library metadata

[1316_full.xlsx](https://github.com/user-attachments/files/26460805/1316_full.xlsx)

and filtered dataset with VISIT/VISITNUM/VISITDY data (maybe it's more representative since it has library metadata filled)
[1316_example.xlsx](https://github.com/user-attachments/files/26460804/1316_example.xlsx)


resources/schema/rule/Rule_Type.md
```diff
@@ -27,6 +27,7 @@ The combination of data in each rule type.
 |                       | [Value Check with Variable Metadata](#value-check-with-variable-metadata)                                                             |
 |                       | [Value Check against Define XML Variable](#value-check-against-define-xml-variable)                                                   |
 |                       | [Value Check against Define XML VLM](#value-check-against-define-xml-vlm)                                                             |
+|                       | [Value Check against Library Metadata](#value-check-against-library-metadata)                                                         |
```
gerrycampion, line 30: You need to also add this rule type to `Rule_Type.json`. This is why the merge script is not finding anything.

gerrycampion (changes requested):

cdisc_rules_engine/dataset_builders/contents_library_variables_dataset_builder.py
```diff
@@ -0,0 +1,33 @@
...
+        "library_variable_ccode",
+        ...,
+        """
+        # get dataset contents and convert it from wide to long
+        data_contents_long_df = ValuesDatasetBuilder.build(self)
+        variables_metadata = self.get_library_variables_metadata()
+        # merge dataset contents with library variable metadata
+        merged = data_contents_long_df.merge(
+            variables_metadata.data,
+            how="outer",
```
SFJohnson24, line 29: is there a reason this is an outer join versus a left join?  The rules mentioned want to check if values are not present in the library.  I dont think we want all the library metadata that isnt associated with a variable?

SFJohnson24 (changes requested):
I think we just want to change that join to left and this is good to go

SFJohnson24:
@alexfurmenkov I was mistaken--this should be an outer join to allow checks when it exists in library but not as a value in the dataset.  Apologies

SFJohnson24 (approved):
This PR correctly implements the new rule type and builder as well as all the requisite additions--readme, schema, testing.

PR #1687, opened 2026-04-09
1681: added @cached decorator to get_dataset

### LINKED ISSUE #1681: get_dataset() is not cached.
the call to builder.get_dataset() is not cached--this can be quite costly for memory as we are concatinating split datasets and building analysis datasets with build() and not caching the result, forcing each thread to build their respective datasets repeatedly

alexfurmenkov:
added caching to BaseDatasetBuilder.get_dataset function. added new fields to BaseDatasetBuilder to support caching without 
added corresponging unit tests and a fixture to clean cache between tests


cdisc_rules_engine/dataset_builders/base_dataset_builder.py
```diff
@@ -69,6 +79,7 @@ def build_split_datasets(self, dataset_name, **kwargs) -> DatasetInterface:
         finally:
             self.dataset_path = original_path
 
+    @cached("get_dataset")
     def get_dataset(self, **kwargs):
```
RamilCDISC, line 83: The get dataset still accepts kwargs which are then passed to concat_split_dataset function. Either pass kwargs to cached instance or could we please have a test to confirm kwargs cannot affect returned dataset

RamilCDISC (changes requested):

cdisc_rules_engine/dataset_builders/base_dataset_builder.py
alexfurmenkov: Just to clarify - should kwargs affect the resulting dataset in this case?

Currently, kwargs are passed to concat_split_datasets and used in DatasetBuilder.build function, but they are not included in the cache key (the decorator only considers args). I've added test to check that kwargs are ignored. 

I've noticed, that we can't pass any kwargs to this function -- it will cause issues later in `JsonSchemaCheckDatasetBuilder` -- it uses same kwargs for `dataset_implementation.from_records(filtered, **kwargs)`

SFJohnson24: kwargs doesnt appear to have any impact on get_dataset, nor is it used--we can probably remove it. I think it is an artifact from when we had both test and validate commands that was not removed

SFJohnson24: > I've noticed, that we can't pass any kwargs to this function -- it will cause issues later in JsonSchemaCheckDatasetBuilder -- it uses same kwargs for dataset_implementation.from_records(filtered, **kwargs)

I dont believe the split dataset logic is ever reached with these checks so I do not believe this will be an issue.  @gerrycampion can you confirm?

SFJohnson24: drop_duplicates is an artifact and It is no longer used at all-- @alexfurmenkov can you delete drop_duplicates from the concat_split_datasets logic (as well as all the calls to get get_variables_metadata() as an aside)
we can remove kwargs from get_dataset/concat_split_dataset.

Caching does appear that is will work fine as kwargs is not used--we can eliminate the test showing kwargs does not impact get_dataset as kwargs is not used at all

SFJohnson24 (changes requested):
see comment; otherwise looks good

SFJohnson24 (approved):
this PR correctly resolves the caching issue with get_dataset()'
it also removes artifacts from the test/validate separate command which has been resolved recently through other PRs.

PR #1690, opened 2026-04-10
691: Tests for CoW behavior in pandas

### LINKED ISSUE #691: rules_engine.execute_rules does a deepcopy of the dataset
Refer to the line here:
https://github.com/cdisc-org/cdisc-rules-engine/blob/cdb8cdbc6fd089c6e9d4ad6df3ae522d79f60621/cdisc_rules_engine/rules_engine.py#L313
If the dataset is large, this is an expensive operation for cpu, disk and ram. We should find a better way to do this.

alexfurmenkov:


cdisc_rules_engine/services/cache/in_memory_cache_service.py
```diff
@@ -66,7 +66,11 @@ def add_dataset(self, cache_key, data):
             self.dataset_cache[cache_key] = data
 
     def get_dataset(self, cache_key):
-        return self.dataset_cache.get(cache_key, None)
+        cached = self.dataset_cache.get(cache_key)
+        # Adding DaskDataset will cause downstream issues since Dask does not support copy-on-write.
+        if type(cached) is PandasDataset:
+            cached.data = cached.data.copy(deep=False)
+        return cached
```
RamilCDISC, line 74: Here the same cached wrapper is returned. Pandas CoW protects separate pandas objects sharing same underlying data. Here the change to cached.data mutates the wrapper but in end returns the same object.

tests/unit/test_services/test_cache/test_immutable_cache.py
RamilCDISC, line 1: This test only checks the local CacheService created in this test not the production InMemoryCacheService or PandasDataset. The changes will not be tested properly.

RamilCDISC (changes requested):

cdisc_rules_engine/services/cache/in_memory_cache_service.py
```diff
@@ -82,7 +85,10 @@ def add_batch(
             self.add(prefix + cache_key, item)
 
     def get(self, cache_key):
-        return self.cache.get(cache_key, None)
+        cached = self.cache.get(cache_key)
+        if type(cached) is PandasDataset:
+            return PandasDataset(cached.data.copy(deep=False))
+        return cached
 
     def get_all(self, cache_keys: List[str]):
```
RamilCDISC, line 95: The other get functions like get-all and get_all_by_prefix etc still return directly which can mutate the cached dataset. We need to cover these too as for get() function.

cdisc_rules_engine/rules_engine.py
```diff
@@ -62,6 +64,8 @@
 from cdisc_rules_engine.models.sdtm_dataset_metadata import SDTMDatasetMetadata
 from cdisc_rules_engine.enums.sensitivity import Sensitivity
 
+pd.options.mode.copy_on_write = True
```
RamilCDISC, line 64: This will turn pandas CoW process wide so any one integrating rules engine in their workflow will have this enabled. We should document it by adding to the repository readme.

tests/unit/test_services/test_cache/test_immutable_cache.py
RamilCDISC, line 1: Could you please add a little more testing that includes operations which will affect the length of the dataset for example add drop rows. The engine will perform these operations on the dataset. Confirming proper caching will be helpful.

RamilCDISC (changes requested):

cdisc_rules_engine/services/cache/in_memory_cache_service.py
```diff
@@ -82,27 +85,30 @@ def add_batch(
...
-                items.append(self.cache[key])
+                items.append(self.get(key))
         return items
 
     def dataset_keys(self):
         return self.dataset_cache.keys()
 
     def filter_cache(self, prefix: str) -> dict:
-        return {k: self.cache[k] for k in self.cache.keys() if k.startswith(prefix)}
+        return {k: self.cache.get(k) for k in self.cache.keys() if k.startswith(prefix)}
```
RamilCDISC: This will still return raw cached object for PandasDataset. Please update this too and I think we will be good to merge.

RamilCDISC (changes requested):

RamilCDISC (approved):
The PR removes the expensive deepcopy() requirements on PandasDataset by utilizing pandas CoW. The validation was done by: 

3. Reviewing the updated test to confirm coverage and cases.
4. Ensuring all unit, regression and integration testing pass.
5. Ensured the pandas CoW is turned on before cache service use in normal Validation path. 
6. Validated that all access to cached dataset will prevent mutation.
7. Reviewed the added test and requested changes so they cover the operation on dataset that may change the size length of the dataset.
8. Ensured the CoW behavior is documented because it turns it on process wide. 
9. Validated the Dask path is unaffected. 
10. Validated against any chance of regression.

SFJohnson24 (approved):
I had a few changes that I merged.  @RamilCDISC can you rereview my changes just to get second eyes on it.  Everything @alexfurmenkov did looked good.

I did notice a few things that I implemented:
clear_all — Replaces asizeof.asizeof with cust_asizeof so the rebuilt cache uses the same safe size function as everywhere else.
add_dataset guard — Skips caching datasets larger than max_dataset_cache_size to match add()'s behavior and avoid a ValueError crash if too large.
Unlocked iteration — Snapshots the cache keys under the lock before iterating so a concurrent write can't cause a RuntimeError.

RamilCDISC:
@SFJohnson24  All looks good.

PR #1695, opened 2026-04-15
1675: fix colliding file path

### LINKED ISSUE #1675: Colliding file path
When I just cloned the repository on Windows, I got a colliding path error due to case-sensitive duplicated filenames in the whodrug directory: https://github.com/cdisc-org/cdisc-rules-engine/blob/38c069fc112bce2aa32e7ed98c3349be38c0a7f5/tests/resources/dictionaries/whodrug/dd.txt

Specifically, both dd.txt and ina.txt have lower and upper case variants and overwrite each other. Is it possible to remove one of the case-variants?

alexfurmenkov:


RamilCDISC (approved):
The PR removes the duplicate whodrug dictionary file. There were two files one with small letters name and other with capital letter names. Then cloning the files were treated as one and would overwrite each other. This resolves that.

SFJohnson24 (approved):

PR #1700, opened 2026-04-22
798: Action to check engine against published rules

### LINKED ISSUE #798: All Rules Cron Job Git Action
create a cron job that runs all rules using git actions
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

alexfurmenkov:
workflow to run validation for all published rules, comparing with benchmark validation results

https://github.com/cdisc-org/cdisc-open-rules/pull/22 with changes to allow validation from this repo


.github/workflows/validate-published-rules.yml
```diff
@@ -0,0 +1,207 @@
...
+#   3. Installs engine Python dependencies
+#   4. Iterates every Published/ rule from cdisc-open-rules
+#   5. Runs the engine against each test case
+#   6. Compares output with committed results.csv baseline
+#   7. Publishes a Markdown report to Job Summary and as an artifact
+# ==============================================================================
+name: Validate Published Rules
+
+on:
+  pull_request:
```
gerrycampion: We don't want this to run on pull_request because it takes too long

.github/workflows/validate-published-rules.yml
```diff
@@ -0,0 +1,207 @@
...
+  workflow_dispatch:
+    inputs:
+      rules_ref:
+        description: 'Branch/tag/SHA of cdisc-open-rules to validate against'
+        required: false
+        default: 'main'
+      engine_ref:
+        description: 'Branch/tag/SHA of cdisc-rules-engine to use'
+        required: false
+        default: 'main'
```
gerrycampion: This shouldn't be an input. The engine branch should be selected as normal from the run workflow dropdown.

.github/workflows/validate-published-rules.yml
```diff
@@ -0,0 +1,207 @@
...
+
+    steps:
+      # -----------------------------------------------------------------------
+      # 1. Checkout cdisc-rules-engine
+      # -----------------------------------------------------------------------
+      - name: Checkout cdisc-rules-engine
+        uses: actions/checkout@v6
+        with:
+          repository: cdisc-org/cdisc-rules-engine
+          ref: ${{ inputs.engine_ref || github.sha }}
```
gerrycampion: remove `ref` for engine checkout. It should use whatever the default is.

.github/workflows/validate-published-rules.yml
```diff
@@ -0,0 +1,207 @@
...
+            validation_report.md
+          if-no-files-found: warn
+
+      # -----------------------------------------------------------------------
+      # 7. Write report to GitHub Actions Job Summary
+      # -----------------------------------------------------------------------
+      - name: Write report to workflow summary
+        if: always()
+        run: |
+          [ -f validation_report.md ] && cat validation_report.md >> $GITHUB_STEP_SUMMARY || true
```
gerrycampion: I see this is giving an error because it is too big. I think we should have 2 reports:
- One with a summary table that gets sent to GITHUB_STEP_SUMMARY and the artifacts. I think we could have columns like:

| Rule        | Type     | Number | Execution | Comparison | Reference | Match |
| ----------- | -------- | ------ | --------- | ---------- | --------- | ----- |
| CORE-000114 | positive | 01     | ❌        |            |           |       |
| CORE-000114 | negative | 01     | ✅        | 10         | 10        | ❌    |
| CORE-000114 | negative | 02     | ✅        | 10         | 10        | ✅    |
 
- another detailed report with the diffs, like the one you are producing, but it should only go to the artifact instead of GITHUB_STEP_SUMMARY. I don't think it needs all of the file references and I think it should only show details for Execution or Match failures.

gerrycampion (commented):
This is looking really nice, but I have some comments

.github/workflows/validate-published-rules.yml
```diff
@@ -0,0 +1,269 @@
...
+            ('CASE_NUM',    str(d['num'])),
+            ('EXEC_OK',     '\u2705' if d['exec'] else '\u274c'),
+            ('EXPECTED',    str(d.get('expected', ''))),
+            ('GOT',         str(d.get('got', ''))),
+            ('MATCH',       '\u2705' if d.get('match') else '\u274c'),
+            ('DIFF_FILE',   str(d.get('diff', ''))),
+            ('STDERR_FILE', str(d.get('stderr', ''))),
+          ]:
+            print(k + '=' + shlex.quote(v))
+          PYEOF
```
gerrycampion: This should just go in a python file in `/scripts`. In fact, I think this entire `Run validation for all Published rules` job after the first call to run_validation.sh should go in a python file for easier maintenance. I would copy this utility to simplify markdown table creation: https://github.com/cdisc-org/CORE_Test_Suite/blob/main/scripts/markdown_utils.py

gerrycampion (commented):
The summary and detailed outputs look awesome. I have one comment, sorry I didn't pickup on this earlier. Hopefully it is an easy conversion for ai.

.github/workflows/validate-published-rules.yml
```diff
@@ -0,0 +1,128 @@
...
+      # -----------------------------------------------------------------------
+      # 2. Checkout cdisc-open-rules (rules + test data + helper scripts)
+      # -----------------------------------------------------------------------
+      - name: Checkout cdisc-open-rules
+        uses: actions/checkout@v6
+        with:
+          repository: cdisc-org/cdisc-open-rules
+          ref: ${{ inputs.rules_ref || 'rules_2' }}
+          path: open-rules
+          # If cdisc-open-rules is private, add a PAT secret:
```
RamilCDISC: I think instead of leaving a comment like this we should either use it or totally remove to basedon the requirement.

.github/workflows/validate-published-rules.yml
```diff
@@ -0,0 +1,128 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      # -----------------------------------------------------------------------
+      # 2. Checkout cdisc-open-rules (rules + test data + helper scripts)
+      # -----------------------------------------------------------------------
+      - name: Checkout cdisc-open-rules
+        uses: actions/checkout@v6
+        with:
+          repository: cdisc-org/cdisc-open-rules
+          ref: ${{ inputs.rules_ref || 'rules_2' }}
```
RamilCDISC: here the system will always fallback to using branch rules_2 when it is auto triggered because of push to main. Input.rules_ref is set on manual override. Please let me know if I am misunderstanding the logic.

scripts/validate_published_rules.py
```diff
@@ -0,0 +1,373 @@
...
+    parse results, and write the two report files.
+
+    Returns True if any rule failed, False if all passed.
+    """
+    published_dir = os.path.join(rules_root, "Published")
+    scripts_dir = os.path.join(rules_root, ".github", "scripts")
+
+    if not os.path.isdir(published_dir):
+        print(f"WARNING: Published/ not found under {rules_root}", file=sys.stderr)
+        return False
```
RamilCDISC: Do you think it would be better to fail the test when published rule folder is missing, instead of printing. a warning and silently passing the test? Silently passing might create a false positive here.

RamilCDISC (changes requested):

RamilCDISC:
Thanks for the updates. The PR seems ready to me for merge now,  this s this PR depends on changes in [PR](https://github.com/cdisc-org/cdisc-open-rules/pull/22) we will wait for it to be merged before approving this one.

scripts/validate_published_rules.py
```diff
@@ -0,0 +1,373 @@
...
+    for a rule that produced no per-case JSONL output.
+    """
+    exec_ok = rule_exit == 0
+    row = {
+        "Rule": rule_id,
+        "Type": "\u2014",
+        "Number": "\u2014",
+        "Execution": CHECKMARK if exec_ok else CROSS,
+        "Expected": "\u2014",
+        "Got": "\u2014",
```
gerrycampion: We are a bit inconsistent with naming and it can be a little confusing.
This applies to both the summary table and detailed report. I think we should name these "Expected" and "Actual". This also applies to the diff results which has "committed" and "generated". We should use "Expected" and "Actual" here as well.

gerrycampion (changes requested):

SFJohnson24:
made this PR to remove the .csv file extension in the results.csv as requested  https://github.com/cdisc-org/cdisc-open-rules/pull/38

PR #1705, opened 2026-04-23
702: Add regex support for target sorting in target_is_sorted_by operator

### LINKED ISSUE #702: Rule blocked: CORERULES-246
- Attach test dataset files that can be used to test the new feature
- Replace the "###" in the issue title with the primary JIRA ticket number for the rule. 

- https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-
- https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-

**Rule Information**
- **Standard**:
- **Rule ID**:
- **Rule Description**:

**Describe the problem**
A clear and concise description of the problem.

**Describe the solution**
1) Need functionality to compare 2 numeric vectors.
2) Need functionality to populate a vector based on suffix on strings.

**Proposed rule logic**
```yaml
Replace this with an example of how the new solution could be written as a rule
```

**Screenshots**

alexfurmenkov:


cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -1816,37 +1849,55 @@ def target_is_sorted_by(self, other_value: dict):
...
+            if target_regex:
+                working_df = self.value[selected_columns].copy()
+                # Create a temporary column with extracted regex values
+                working_df[f"{target}_extracted"] = self._extract_regex_group(
+                    working_df[target], target_regex
+                )
+                target_for_sorting = f"{target}_extracted"
+                # Sort by within columns only, preserve original order within groups
+                sorted_df = working_df.sort_values(
+                    by=within_columns,
```
RamilCDISC: here we would need to sort by within_columns and extracted target not preserve the order for regex branch.

resources/schema/rule-merged/Operator.json
```diff
@@ -850,7 +850,7 @@
...
         }
       },
       "required": ["operator", "value", "within"],
```
RamilCDISC, line 856: I think we should add the new regex property here.

RamilCDISC (changes requested):

resources/schema/rule-merged/Operator.json
alexfurmenkov: It’s not very visible here, but I actually added information about the regex. It’s easier to view it in the editor.

RamilCDISC (changes requested):
I executed the rule CG0546 in dev editor. I used the dataset from folder CG0545 in sharepiont as there was no dataset for CG0546. I made a change and added the suffix to the MIDSTYPE column records in the SM dataset. The updated dataset.is attached. 

I get the following error 

```
{
  "SM": [
    {
      "executionStatus": "execution error",
      "dataset": "SM",
      "domain": "SM",
      "variables": [],
      "message": "rule evaluation error - operation failed",
      "errors": [
        {
          "dataset": "SM",
          "error": "Error occurred during operation execution",
          "message": "Failed to execute rule operation. Operation: record_count, Target: None, Domain: SM, Error: single positional indexer is out-of-bounds"
        }
      ]
    }
  ],
  "TM": [
    {
      "executionStatus": "skipped",
      "dataset": "TM",
      "domain": "TM",
      "variables": [],
      "message": "Rule skipped - doesn't apply to domain for rule id=CDISC.SDTMIG.CG0546, dataset=TM",
      "errors": [
        {
          "dataset": "TM",
          "error": "Outside scope",
          "message": "Rule skipped - doesn't apply to domain for rule id=CDISC.SDTMIG.CG0546, dataset=TM"
        }
      ]
    }
  ]
}
```

[unit-test-coreid-CG0545-negative 1.xlsx](https://github.com/user-attachments/files/27495916/unit-test-coreid-CG0545-negative.1.xlsx)

Please let me know if I updated the dataset incorrectly.

gerrycampion (commented):
The unit tests look good to me, by still waiting on @alexfurmenkov to address @RamilCDISC's test results

alexfurmenkov:
Hi @RamilCDISC ,
The initial test dataset unit-test-coreid-CG0545-negative.xlsx contained errors that prevented rule CG0546 from executing (status SKIPPED): the MIDSTYPE field in the SM dataset had values "DIAGNOSIS1" and "DIAGNOSIS2" instead of "DIAGNOSIS" (as in TM), which blocked dataset merging, and the MIDS field contained "DIAG" without trailing digits, failing the regex validation. After fixing these basic issues (changed MIDSTYPE to "DIAGNOSIS" and MIDS to "DIAG1"/"DIAG2"), two correct datasets were created: a positive test with proper order (DIAG1→2005-10, DIAG2→2007-10) and a negative test with violated order (DIAG2→2005-10, DIAG1→2007-10).

When testing the target_is_sorted_by operator in isolation, it worked correctly: for positive data it returned `[True, True]` (data is valid), for negative data it returned `[False, False]` (data is invalid). However, during full validation through the engine, a **complete logical inversion** was discovered: the positive dataset with correct data returned status "ISSUE REPORTED" with 2 errors (should be SUCCESS with 0 errors), while the negative dataset with incorrect order returned status "SUCCESS" with 0 errors (should be ISSUE REPORTED with 2 errors). This means the rule works completely backwards: valid data is marked as invalid (false positives), and invalid data is passed through (false negatives).

The problem is that the rule uses the operator's result directly for error generation: when `target_is_sorted_by` returns `True` (which semantically means "data is properly sorted"), the engine generates an error, and when it returns `False` (data is improperly sorted) - it does not generate an error. 

The quickest fix is to use the `target_is_NOT_sorted_by` operator instead of `target_is_sorted_by` in the rule, which inverts the logic: the operator will return `True` when order is violated, and the rule will correctly generate errors only for incorrect data. 

I attach correct datasets
[unit-test-coreid-CG0545-negative-FIXED.xlsx](https://github.com/user-attachments/files/27717504/unit-test-coreid-CG0545-negative-FIXED.xlsx)
[unit-test-coreid-CG0545-positive.xlsx](https://github.com/user-attachments/files/27717510/unit-test-coreid-CG0545-positive.xlsx)

RamilCDISC (approved):
The PR adds regex support for target_is _sorted_by_operator. The validation was done by: 

4. Ensuring relevant testing is updated. 
5. Ensuring documentation is updated.
6. Ensuring the regex support is properly implemented.
7. Verifying the regex implementation in the operator will handle edge cases.
8. Verifying the regex implementation is similar to other operators.
9. Running manual validation using dev editor using negative dataset.
10. Running manual validation using dev editor using positive dataset.

SFJohnson24 (approved):
this PR preserved base operator functionality while adding regex functionality.  Checked SEND354 and CG0620 in addition to rule associated with issue

PR #1709, opened 2026-04-24
fix codelist operations type inference for empty dataframes

### LINKED ISSUE #1651: New version of Core engine causes errors in USDM rule execution
### Standard

CDISC USDM

### Reference Rule ID(s)

DDF00141, DDF00093, DDF00094

### Conformance Rule ID(s) (if published in CORE)

CORE-000857, CORE-000873, CORE-000874

### JIRA Ticket

https://jira.cdisc.org/projects/CORERULES/issues/CORERULES-

### CLI Command Used (if applicable)

### Rule YAML

```YAML
Authorities: 
  - Organization: 
    Standards:
      - Name: 
        References:
          - Citations:
              - Cited Guidance: 
                Document: 
                Item: 
                Section: 
            Origin: 
            Rule Identifier: 
              Id: 
              Version: 
            Version: 
        Version:
Check:
  all:
    - name: 
      operator: 
      value: 
Core:
  Id: 
  Status: 
  Version: '1'
Description: 
Match Datasets:
  - Keys:
      - 
    Name: 
Operations:
  - id: 
    operator: 
Outcome:
  Message: 
  Output Variables:
    -
Rule Type: 
Scope:
Sensitivity:
```

### Attach any sample test data file(s)

Standard Fixed json version gives error for gender + client version where studyVersion.date is missing (which is allowed).

### Attach any output report and/or log file(s)

[Test_CoreMar26_fixedFile.xlsx](https://github.com/user-attachments/files/25799159/Test_CoreMar26_fixedFile.xlsx)

+ client output (cannot be shared) gives "EXECUTION ERROR" for rule DDF00093 and DDF00094

### Expected output

SUCCESS run - no issues.

### Any Additional Information

_No response_

gerrycampion:
Fixes the issue for CORE-000857 with #1651. I cannot reproduce the issue for CORE-000873, CORE-000874 without the proper data.

Previously, we would get an Execution Error because the operations input dataframe is empty (after match datasets) and the column datatypes cannot be inferred.
[Broken.xlsx](https://github.com/user-attachments/files/27057714/Broken.xlsx)
Now we get a skip.
[Fixed.xlsx](https://github.com/user-attachments/files/27057713/Fixed.xlsx)

Ran with following data and launch config:
[pilot_LZZT_narrative_2026MAR10.json](https://github.com/user-attachments/files/27057835/pilot_LZZT_narrative_2026MAR10.json)
```json
{
  "name": "USDM Rule",
  "type": "debugpy",
  "request": "launch",
  "program": "${workspaceFolder}/core.py",
  "console": "integratedTerminal",
  "args": [
    "validate",
    "-s",
    "usdm",
    "-v",
    "4-0",
    "-dp",
    "pilot_LZZT_narrative_2026MAR10.json",
    "-r",
    "CORE-000857",
    "-of",
    "XLSX",
    "-of",
    "json",
    "-l",
    "debug"
  ]
}
```

### DIFF
--- a/cdisc_rules_engine/operations/codelist_extensible.py
+++ b/cdisc_rules_engine/operations/codelist_extensible.py
@@ -26,6 +26,12 @@ class CodelistExtensible(BaseOperation):
             self.params.ct_package_type, unique_ct_versions
         )
         ct_df = self.evaluation_dataset.__class__.from_dict(ct_data)
+        ct_df = ct_df.astype(
+            {
+                "version": str,
+                "codelist_code": str,
+            }
+        )
         if self.params.codelist_code in self.evaluation_dataset.columns:
             is_extensible = self.evaluation_dataset.merge(
                 ct_df.data,
--- a/cdisc_rules_engine/operations/codelist_extensible.py
+++ b/cdisc_rules_engine/operations/codelist_extensible.py
@@ -28,8 +28,8 @@ class CodelistExtensible(BaseOperation):
         ct_df = self.evaluation_dataset.__class__.from_dict(ct_data)
         ct_df = ct_df.astype(
             {
-                "version": str,
-                "codelist_code": str,
+                "version": "string",
+                "codelist_code": "string",
             }
         )
         if self.params.codelist_code in self.evaluation_dataset.columns:
--- a/cdisc_rules_engine/operations/codelist_terms.py
+++ b/cdisc_rules_engine/operations/codelist_terms.py
@@ -60,8 +60,8 @@ class CodelistTerms(BaseOperation):
         ct_df = self.evaluation_dataset.__class__.from_dict(ct_data)
         ct_df = ct_df.astype(
             {
-                "version": str,
-                "codelist_code": str,
+                "version": "string",
+                "codelist_code": "string",
             }
         )
         if self.params.codelist_code in self.evaluation_dataset.columns:
--- a/cdisc_rules_engine/operations/codelist_terms.py
+++ b/cdisc_rules_engine/operations/codelist_terms.py
@@ -58,6 +58,12 @@ class CodelistTerms(BaseOperation):
             self.params.ct_package_type, unique_ct_versions
         )
         ct_df = self.evaluation_dataset.__class__.from_dict(ct_data)
+        ct_df = ct_df.astype(
+            {
+                "version": str,
+                "codelist_code": str,
+            }
+        )
         if self.params.codelist_code in self.evaluation_dataset.columns:
             result = self.evaluation_dataset.merge(
                 ct_df.data,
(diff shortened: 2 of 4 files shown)


RamilCDISC (changes requested):
Would it be better to add a unit or regression test for this case?

gerrycampion:
> Would it be better to add a unit or regression test for this case?

Yes, I added unit tests that fail on main but pass on this branch

cdisc_rules_engine/operations/codelist_extensible.py
```diff
@@ -26,6 +26,12 @@ def _handle_multiple_versions(self) -> pd.Series:
             self.params.ct_package_type, unique_ct_versions
         )
         ct_df = self.evaluation_dataset.__class__.from_dict(ct_data)
+        ct_df = ct_df.astype(
```
RamilCDISC, line 29: The PR only converts the ct_df to string but evaluation dataset is still not made safe for merge. if it has empty columns the merge can fail. I have noticed that in the test the evaluation dataset was made safe manually by converting it to string. That let me to think that we need to make the evaluation dataset safe here too to merge without errors.

cdisc_rules_engine/operations/codelist_extensible.py
```diff
@@ -26,6 +26,12 @@ def _handle_multiple_versions(self) -> pd.Series:
...
+            {
+                "version": str,
```
RamilCDISC: using str to covnert to string here can convert missing values like nan to "nan" which could create issue in downstream flow. One approach can be to convert like: 
`{

       "version": "string",
       "codelist_code": "string"

}
`

cdisc_rules_engine/operations/codelist_terms.py
```diff
@@ -58,6 +58,12 @@ def _handle_multiple_versions(self) -> pd.Series:
             self.params.ct_package_type, unique_ct_versions
         )
         ct_df = self.evaluation_dataset.__class__.from_dict(ct_data)
+        ct_df = ct_df.astype(
+            {
+                "version": str,
```
RamilCDISC: same here too

RamilCDISC (changes requested):

cdisc_rules_engine/operations/codelist_extensible.py
gerrycampion: i believe that the fix for the types allowed us to have empty evaluation datasets as well. I've removed the conversions from the tests.

RamilCDISC (approved):
The PR fixes the bug for empty dataframe in codelist operations. The PR was validated by: 

2. Reviewing the PR in accordance with the AC.
3. Reviewing the updated tests for logic and coverage.
4. Reviewing the code for maintaining quality and downstream safe checks.
5. Running validation using positive dataset in dev editor.
6. Running validation using negative dataset in dev editor.
7. Running validation using edge case of empty dataframe.
8. Ensuring all unit and integration testing pass.

PR #1711, opened 2026-04-28
1699: get_parent_model_column_order calculation made only for unique rdomains and merged by

### LINKED ISSUE #1699: CORE-000783
CORE-000783 appears to be holding up a validation for a stakeholder--the logs made it clear this was due to the fact that they have a very large SUPP-- dataset.  
"
Since multiple processes are running simultaneously, in the logs the completion does not always immediately follow the start; log records from other rules can appear in between.

Validation runs up to 99% — after that I made a comparison of:
all rules that should run
all rules for which validation was started
all rules for which validation was completed
Conclusion: all rules are started, but one rule gets stuck, namely CORE-000783
(Raise an error when SUPP--.QNAM is present in the dataset, but the value of SUPP--.QNAM is equal to a variable name defined in the corresponding SDTM version.)
If I start the validation without this rule, the validation does complete successfully.

JSON
Check:

YAML

all:
- name: QNAM
operator: is_contained_by
value: $model_variables

Core:

Id: CORE-000783
Status: Published
Version: "1"
Description:

Raise an error when SUPP--.QNAM is present in the dataset, but the value of SUPP--.QNAM is equal to a variable name defined in the corresponding SDTM version.
Executability:
 Fully Executable
Operations:

YAML

- id: $model_variables
operator: get_parent_model_column_order

Outcome:

• Message:
SUPP--.QNAM is present in the dataset, but the value of SUPP--.QNAM equals a variable name defined in the corresponding SDTM version.
• Output Variables:
QNAM
$model_variables
Rule Type:
 Record Data
Scope:

If I remove SUPPLB, or if I ensure that SUPPLB contains only 10 records, validation also completes successfully.
I suspect this is related to the more than 64,000 records in SUPPLB → it probably performs the check for each QNAM, so I suspect that a distinct on QNAM should be applied first.
"

https://github.com/cdisc-org/sdtm-adam-pilot-project/tree/master/updated-pilot-submission-package/900172/m5/datasets/cdiscpilot01/tabulations/sdtm - pilot supplb data

we will need to find the bottleneck:
i believe it is the is_contained_by operatordoing iloc() in a for loop, causing significant call overhead from the indexing cost of doing that on 64k calls.  We could convert columns to lists and enumerate to iterate over plain Python values instead. This would eliminate the repeated pandas indexing overhead by paying the conversion cost once upfront via .tolist().  (assuming this is the bottleneck but it does appear to be the most memory intensive step of this)

alexfurmenkov:


SFJohnson24 (changes requested):
I am unfortunately not getting the hang up described in the issue locally-- that said, the changes you made dont appear to impact the performance.  The original was O(n) for the number of supp-- rows since _get_parent_variable_names_list was called once per unique RDOMAIN value via set(rdomain_values) loop.  Your refactor is also O(n) since the setdefault call inside the generator still iterates over all n rows of rdomain_values and the expensive _get_parent_variable_names_list call is still only made once per unique RDOMAIN value, just lazily rather than upfront.  I may be off here if you were able to reproduce the hangup?  I still believe the bottleneck is likely the is_contained_by as it is O(n*m) presently -- number of rows (64k in this case) x the length of the comparator list.  It may not be the iloc but that adds extra overhead of an already expensive loop.

-  I think reducing n by using a distinct
- we could also use a set on comparison_data before the loop so each is_in check is O(1) instead of O(m)

What do you think?

alexfurmenkov:
The setdefault approach inside a generator does not actually cache the results of _get_parent_variable_names_list. In Python, function arguments are evaluated eagerly — meaning the second argument to setdefault is fully evaluated before setdefault is even called. So _get_parent_variable_names_list gets called once per row (n times total), not once per unique RDOMAIN value.
```Python
d = {}
def expensive(x):
    print(f"called with {x}")
    return x * 10

for v in ["A", "A", "A"]:
    d.setdefault(v, expensive(v))
# Output:
# called with A
# called with A
# called with A   ← 3 times, not 1
```
So the original setdefault version is O(n rows), while the explicit set() loop version:
```Python
for rdomain in set(rdomain_values):
    rdomain_names_list[rdomain] = self._get_parent_variable_names_list(...)
```
is O(k unique RDOMAIN values), which is a real improvement, not just a cosmetic refactor.

SFJohnson24 (approved):
This PR correctly preserves get_parent_model_column_order while optimizing it for performance, going from O(n) I/O calls down to O(u) by deduplication with set().  Neither dev was able to reproduce the hangup using the pilot data for both the full SDTMIG and just the one problematic rule.  Should the hangup persist in v0.16 we will look to is_contain_by and work to optimize there

PR #1720, opened 2026-05-05
update $id

SFJohnson24:
merge schema: https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/25388257044

### DIFF
--- a/resources/schema/rule-merged/CORE-base.json
+++ b/resources/schema/rule-merged/CORE-base.json
@@ -191,7 +191,7 @@
       "description": "Can reference either a dataset variable name or an operation result"
     }
   },
-  "$id": "https://cdisc.org/CORE-base.json",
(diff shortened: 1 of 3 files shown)


gerrycampion (commented):
Good, though I noticed now 2 things:
- `...refs/heads/main...` are just `...main...` in the open rules md files. These should probably be updated for consistency: https://github.com/cdisc-org/cdisc-open-rules/blob/809d0f52b6b10390229a91291f68f236e66557e7/docs/rule_type.md?plain=1
- Should we use rule-merged here in case the validator can give more context? https://github.com/cdisc-org/cdisc-open-rules/blob/809d0f52b6b10390229a91291f68f236e66557e7/.github/workflows/validate-yaml-schema.yml#L35

gerrycampion (approved):

PR #1724, opened 2026-05-06
fix the versions for library to prevent installing incompatible version

RamilCDISC:


gerrycampion:
We removed rule editor e2e testing altogether

PR #1729, opened 2026-05-11
1023: Ensure UTF Support

### LINKED ISSUE #1023: Ensure UTF Support in Report Generator

DmitryMK:
Addresses #1023 
- Added tests to verify that reporting service support UTF in both Excel and JSON.
- Added additional tests to check different encodings (latin1, cp1252, utf8) are reported correctly.


tests/QARegressionTests/test_Issues/test_CoreIssue1023.py
RamilCDISC, line 1: Could we please move this test under regression testing instead? Regression testing folder has similar tests where we run the engine using subprocess.

RamilCDISC (changes requested):
JsonReport.write_report() still uses a `with open(self._output_name, "w")` and `json.dump(report_data, f)` without explicit encoding or `ensure_ascii=False` the json.dump ensure_ascii defaults to True. This can escape non ASCII characters instead of writing them as it is.

RamilCDISC (approved):
The PR ensures UTF-8 encoding for writing the reports and that non-asciii characters are handled properly. The PR was validated by:

4. Ensuring the testing is placed in proper pace and covers the AC.
5. Ensuring the report writing defaults to 'UTF-8'
6. Ensuring both formats (excel and json) both reports handle multilingual characters and non-ASCII characters.
7. Ensuring non-ASCII characters are displayed and intractable in reports.

PR #1737, opened 2026-05-18
date bug

### LINKED ISSUE #1742: date bug
see attached PR

SFJohnson24:
SEND354 and CG0620 produce a float cannot be iterated issue--this PR updates the catch in the beginning of date check to not only catch None but also catch non-strings as floats are getting passed this first line and causing the rules to fail to execute

[CORE-Report-2026-05-18T08-54-41.xlsx](https://github.com/user-attachments/files/27961803/CORE-Report-2026-05-18T08-54-41.xlsx)
[CORE-Report-2026-05-18T09-04-05.xlsx](https://github.com/user-attachments/files/27961804/CORE-Report-2026-05-18T09-04-05.xlsx)
result pre- and post- fix


RamilCDISC (approved):
The PR adds an additional type check to the date is valid helper in check operators. This check for string dates successfully resolves the bug reported for the Rules in attached tickets. See attached reports for results in the PR description.

PR #1738, opened 2026-05-19
removed underscore logic from rules

### LINKED ISSUE #1732: Remove underscore logic
There is remaining underscore logic for rules left over from editor.  https://github.com/cdisc-org/cdisc-rules-engine/blob/46ef2f2b3e6efd3221bdd34f6b5a3cb31e71d358/cdisc_rules_engine/models/rule.py#L85  This is not needed and it would be good to transition out of it--the parsing logic as well as the references to paramaters with underscores in their names will need to be phased out.

this ticket spurred this one: https://github.com/cdisc-org/cdisc-library-api/issues/16

alexfurmenkov:


cdisc_rules_engine/models/rule.py
```diff
@@ -35,11 +35,21 @@ def __init__(self, record_params: dict):
         self.output_variables: dict = record_params.get("output_variables")
         self.grouping_variables: List[str] = record_params.get("grouping_variables", [])
 
+    @classmethod
+    def _get_key(cls, obj: dict, space_key: str, default=None):
+        """Get a value by key, supporting both space-format ('Rule Type') and
+        underscore-format ('Rule_Type') for backward compatibility with the
```
RamilCDISC, line 41: While I understand having backward compatibility is good but the ticket AC is to remove underscore format, which was for rule editor. @alexfurmenkov @SFJohnson24 do you agree to keep backward compatibility or fully remove it? From the ticket AC we should remove this fully.

RamilCDISC (changes requested):

cdisc_rules_engine/models/rule.py
SFJohnson24: @RamilCDISC yes we would like to remove the underscore logic.  This PR should be merged after the API changes https://github.com/cdisc-org/cdisc-library-api/issues/16 as it will no longer have underscores

cdisc_rules_engine/models/rule_validation_result.py
```diff
@@ -48,12 +48,17 @@ def _get_rule_ids(self, rule: Rule, org: str) -> str:
         return ", ".join(
             sorted(
                 {
-                    reference.get("Rule_Identifier", {}).get("Id")
+                    (
+                        reference.get("Rule Identifier")
+                        or reference.get("Rule_Identifier")
```
SFJohnson24, line 53: note this line will be removed once github api logic is merged

PR #1739, opened 2026-05-19
define env

### LINKED ISSUE #1740: define-xml open rules
the define in the .env is not a path but a filename; logic needs to be added to accept both in engine

SFJohnson24:
- updates env.example to .env.example; updated format
- adds logic where if there is -dxp and -dep, checks if the define_xml is coming from the .env and if it is a complete path or a filename; if filename, uses the location of the .env to construct a path
- update docs to reflect above behavior


core.py
```diff
@@ -600,6 +600,13 @@ def validate(  # noqa
     load_dotenv(dotenv_path)
     validate_dataset_files_exist(dataset_path, logger, ctx)
 
+    if define_xml_path and dotenv_path and os.getenv("DEFINE_XML"):
+        dxp = Path(define_xml_path)
+        if not dxp.is_absolute() and dxp.parent == Path("."):
+            resolved = Path(dotenv_path).parent / define_xml_path
```
gerrycampion, line 606: Nice, I didn't know path had this slash operator

gerrycampion (approved):

PR #1757, opened 2026-06-05
Test/filipps update deps

SFJohnson24:
per the original PR:
Summary: expand dependency constraints and turn requirements.txt from a requirements file into a lockfile. This will allow the PyPi version of CORE to be used as a library in a larger project where other dependency constraints exist.

List of changes:

    Define dependency constraints in pyproject.toml
        Remove requirements-dev.txt
        Repurpose requirements.txt as a lockfile
    Change constraints from exact pinned versions to ranges
    Dependency range upgrades:
        Remove CLI defaults that clash with required to support a behavior change/fix in click 8.3.0
        Edit DatasetXPTMetadataReader.read to support a behavior change in pyreadstat 1.2.9
        Edit USDMDataService.__get_full_path to support a behavior change in jsonpath-ng 1.8.0
        Remove unnecessary self capture in ContentsDefineVLMDatasetBuilder to support dask 2024.8.1
        Remove __setitem__ reindexing in DaskDataset to fix errors surfaced in dask 2025.4.0
        Fix test_dataset_metadata_define_dataset_builder to support a sort behavior fix in pandas 2.2.0

Tested scenarios:

 Unit tests
 Verification succeeds with [CDISC_Pilot_Study_v4_FIXED.json](https://github.com/cdisc-org/cdisc-jsonata-rules/blob/main/testdata/CDISC_Pilot_Study_v4_FIXED.json)

This PR also removes requirements.txt and uses pyproject.toml for requirements-- tweaking dependency constraints as requirements was version pinned and adapting some methods to new pandas type coercion behavior in the updated versions.
Fix to_dict(orient="records") in DaskDataset to use compute() directly instead of map_partitions, which no longer implicitly computes in Dask 2025.x.
Fix __setitem__ in DaskDataset to handle both list and pd.Series values by computing to pandas before assignment, avoiding broken task graphs from da.Array → dd.DataFrame column assignment in newer Dask.
tweaked all git actions and docs to reflect this new workflow; tweaked test suite 
Cast Codelist / codelist extensible and evaluation dataset merge key columns to str explicitly to prevent pandas 2.x merge failures caused by stricter type matching.

Note: the USDM template will need to be updated upon merge for the one result


tests/unit/test_dataset_builders/test_dataset_metadata_define_dataset_builder.py
```diff
@@ -143,15 +142,24 @@ def test_dataset_metadata_define_dataset_builder(dataset_path):
...
+        .astype(object)
+        .sort_values("dataset_location")
+        .reset_index(drop=True)
+    )
 
-    result_df = result.data[expected_df.columns].reset_index(drop=True)
+    result_df = (
+        result.data[expected_df.columns]
+        .sort_values("dataset_location")
+        .reset_index(drop=True)
```
gerrycampion: I think we should enforce a predictable sort order inside the builder instead of here.

pyproject.toml
```diff
@@ -4,12 +4,49 @@ build-backend = "setuptools.build_meta"
...
+
+[dependency-groups]
+dev = [
+    "black >=24.10.0",
+    "flake8 >=6.1.0",
+    "pre-commit >=2.20.0",
+    "pytest >=7.4.0, <8.0.0",
+    "pytest-asyncio >=0.21.0",
+    "pytest-cov >=6.0.0",
+]
```
gerrycampion, line 53: I think requirements-rule-tester should be removed and azure-functions should go here

gerrycampion (changes requested):

gerrycampion (approved):
i updated the rule tester dependency. looks good

PR #1759, opened 2026-06-08
Fixed published open rules for usdm

### LINKED ISSUE #1758: Fix Published Open Rules for USDM
Running open rules against usdm rules is currently broken due to some issues with the CSV reader and report.

- [x] Fix variable length int conversion exception
- [x] Fix CSV metadata reader to recognize entity names, which are case sensitive, not upper
- [x] output path column

gerrycampion:
Fixes the tasks detailed in the attached issue


cdisc_rules_engine/services/csv_metadata_reader.py
```diff
@@ -32,59 +33,87 @@ def __init__(
         )
 
     def read(self) -> dict:
-        dataset_name = Path(self.file_name).stem.lower()
```
RamilCDISC, line 35: Previously the name of the dataset was lowered case which helps in comparison below. The new logic does not lower the case. This could cause incorrect results.

cdisc_rules_engine/services/csv_metadata_reader.py
```diff
@@ -32,59 +33,87 @@ def __init__(
...
-                "categorization_scheme": {},
-                "w_indexes": {},
-                "period": {},
-                "selection_algorithm": {},
-            },
+        return {
+            "dataset_name": (
+                single_match["Dataset Name"]
+                if "Dataset Name" in datasets_df.columns
+                else str(single_match["Filename"]).upper()
```
RamilCDISC, line 89: Maybe I am misunderstanding the intent here. The update here is according to AC "Fix CSV metadata reader to recognize entity names, which are case sensitive, not upper" but for the case when dataset name is missing it still returns uppercase of filename.

cdisc_rules_engine/services/csv_metadata_reader.py
```diff
@@ -95,7 +124,11 @@ def __get_variable_metadata(
             zip(variable_names, dataset_meta_df["type"])
         )
         variable_name_to_size_map = {
-            var: (int(length) if pd.notna(length) else None)
+            var: (
+                int(length)
+                if pd.notna(length) and (isinstance(length, int) or length.isdigit())
```
RamilCDISC: If a column has a missing/blank value in csv, it can cause pandas to read the whole column as float instead of int. So for row where value is present the check of length.isdigit() will cause error because floats don't have isdigit()

RamilCDISC (changes requested):

cdisc_rules_engine/services/csv_metadata_reader.py
gerrycampion: usdm uses case-sensitive dataset names. We don't want to force lower. do you have an example how it would cause incorrect results?

cdisc_rules_engine/services/csv_metadata_reader.py
gerrycampion: we don't know whether it will be definitely be usdm or sdtm. When it is usdm, hopefully the _datasets contains a "dataset name" column, so we can use that name. If that column doesn't exist, our best option is to assume it is not usdm, in which case, the name is the upper of the filename.

cdisc_rules_engine/services/csv_metadata_reader.py
gerrycampion: why would it read the whole column as float?

cdisc_rules_engine/services/csv_metadata_reader.py
RamilCDISC: I raised concern because of how old code was handling it. I understand the intent now.

cdisc_rules_engine/services/csv_metadata_reader.py
RamilCDISC: Here on pandas documentation it states it: [link](https://pandas.pydata.org/docs/user_guide/integer_na.html)

tests/unit/test_csv_reader.py
RamilCDISC: The value in the fixture now is changed to be 'patients' from 'patients.csv'. This line here still replaces the value 'patients.csv' with its upper case version, which will basically update nothing. This defeats the test purpose judging by its name.

RamilCDISC (changes requested):

cdisc_rules_engine/services/data_services/local_data_service.py
```diff
@@ -144,6 +144,9 @@ def get_dataset(self, dataset_name: str, **params) -> DatasetInterface:
             basename(full_path).split(".")[1].upper()
         )
         df = reader.from_file(full_path)
+        # Build a simulated json pointer for the case where we are simulating json data.
+        if self.standard == "usdm":
```
SFJohnson24, line 148: I think this and the excel data service using string literals for usdm are a bit fragile-- I ran a validation using USDM in capitals versus usdm in lowercase and got different results for the same data.

SFJohnson24 (changes requested):
[lowercase.json](https://github.com/user-attachments/files/29060908/lowercase.json)
[uppercase.json](https://github.com/user-attachments/files/29060909/uppercase.json)

See comment -- there is a discrepency in issues produced with standard name when capitalized vs lowercase.  When I add a lower() call to the standard, it reverts back to the lowercase 3 issues.

gerrycampion:
> [lowercase.json](https://github.com/user-attachments/files/29060908/lowercase.json) [uppercase.json](https://github.com/user-attachments/files/29060909/uppercase.json)
> 
> See comment -- there is a discrepency in issues produced with standard name when capitalized vs lowercase. When I add a lower() call to the standard, it reverts back to the lowercase 3 issues.

@SFJohnson24 These files aren't accessible. Can you reattach as .txt files?

cdisc_rules_engine/services/data_services/local_data_service.py
gerrycampion: This was actually a bug with the recently-added caching. I've resolved the bug and you should now see consistent results. Let me know if you still have other concerns about the capitalization.

SFJohnson24 (approved):
This PR correctly resolves the issues in the attached task.  It also resolves the differences in report with the same data but different capitalization

PR #1761, opened 2026-06-09
update regression tests and run them in CI tests

### LINKED ISSUE #1686: Regression testing
this issue is an extension of #688  

A/C update and make regression tests to be part of CI pipeline (these could be their own git action to run in parallel to unit tests, etc)
Some tests are skipped--these should not skip

RamilCDISC:
The PR updates the broken regression tests and removes the regression markers from all the tests to ensure all QARegressionTests are run during the CI pipeline.


tests/QARegressionTests/test_core/test_validate.py
```diff
@@ -163,7 +163,7 @@ def test_validate_all_options(self):
             "--meddra",
             os.path.join("tests", "resources", "dictionaries", "meddra"),
             "-r",
-            os.path.join("tests", "resources", "Rule-CG0027.json"),
+            "CORE-000237",
```
SFJohnson24: can we remove the rule from resources if we are no longer using it for the tests

tests/QARegressionTests/test_Issues/test_CoreIssue1345.py
```diff
@@ -1,167 +1,167 @@
...
+            "resources",
+            "CoreIssue1345",
+        ),
+        "-lr",
+        os.path.join(
+            "tests",
+            "resources",
+            "CoreIssue1345",
+        ),
+        "-r",
```
SFJohnson24: I am a bit confused on this as -r is only for published rules and this is not a CORE-ID.  Is CORE silently ignoring the -r and executing and that is what is being testing?

tests/QARegressionTests/test_Issues/test_CoreIssue1487.py
```diff
@@ -1,54 +1,54 @@
...
+        # Run the command in the terminal
+        command = [
+            f"{get_python_executable()}",
+            "-m",
+            "core",
+            "validate",
+            "-s",
+            "sdtmig",
+            "-v",
+            "3-4",
```
SFJohnson24: this test seemed to want a version of SDTMIG that doesnt contain a rule and was rewritten to not do that --was there a reason for this?

tests/QARegressionTests/test_Issues/test_CoreIssue1487.py
```diff
@@ -1,54 +1,54 @@
...
+        assert "Rules Report" in workbook.sheetnames
+        rules_sheet = workbook["Rules Report"]
+        target_row = None
+        for row in rules_sheet.iter_rows(min_row=2, values_only=True):
+            if row[0] == "CORE-000354":
+                target_row = row
+                break
+        assert target_row, "Rule CORE-000354 not present in 'Rules Report' sheet."
+        assert (
+            target_row[4] in target_row[4]
```
SFJohnson24: this assertion will always be true.  target_row[4] will never not be in target_row[4]

SFJohnson24 (changes requested):
a few questions--see comments

tests/QARegressionTests/test_Issues/test_CoreIssue1345.py
RamilCDISC: Its the rule file actually has this 
```
Core:
  Id: CDISC.SDTMIG.CG0019
```

So I think it is matching correctly the core id.

tests/QARegressionTests/test_Issues/test_CoreIssue1487.py
RamilCDISC: Yes, I forgot to update that part. The real intention was to discuss this test. CoreIssue1487 required that we skip a rule if that is not part of the selected version. Now the engine raises. 

`cdisc_rules_engine.exceptions.custom_exceptions.LibraryMetadataNotFoundError: No library metadata found for standard 'sdtmig' version '5.0'.`

We can either update the test to confirm for this exception or do we want to restore engine's functionality to skip instead of an exception?

tests/QARegressionTests/test_core/test_validate.py
RamilCDISC: It is still being used in other tests, thats why I did not remove it.

tests/QARegressionTests/test_Issues/test_CoreIssue1487.py
SFJohnson24: if you are running for an invalid standard, i would expect an exception

tests/QARegressionTests/test_Issues/test_CoreIssue1345.py
SFJohnson24: yes but why is the -r there if it executing from the local rule file?  Is it not running the rule specified with -r purposefully?  Trying to understand that line specifically as it relates to the test

RamilCDISC: I have changed it so -lr directly points to that rule file, instead of using -r for selecting the rule.

tests/QARegressionTests/test_Issues/test_CoreIssue1487.py
RamilCDISC: updated the test

tests/QARegressionTests/test_core/test_validate.py
SFJohnson24: I think we should continue to use the local instance of this as COREZ-000237 is an SDTMIG rule.  With the split for standards with Verisian--this will have downstream issues when we have their engine for SDTM.

SFJohnson24: see lower comment-- I think this should be reverted to the local rule file and not -r from library

tests/QARegressionTests/test_Issues/test_CoreIssue1345.py
SFJohnson24: that is not the line I am referring to,  -r CDISC.SDTMIG.CG0019 is what I am referring to.
-r is only for published rule IDs and CDISC.SDTMIG.CG0019 is not a CORE ID so is having that as an argument the reason for the test?  Looking at the ascertions, the rule runs so I assume it is just not using the -r command so can we remove that line

SFJohnson24 (changes requested):

SFJohnson24 (changes requested):
2 changes:
A.  there are a few rules still using library and the -r command in the regression tests, can we switch them to -lr and store these rules in resources to have the rule locally so any changes to library do not break the regression.
B.  can we add comments to each test case stating what it is testing and if we are not sure what it is testing/if it is redundant, can we prune the test to optimize the runtime for the tests and remove unneeded tests. This will help in the future as functionality changes/tests get stale to understand what each test is for

SFJohnson24 (approved):
looks good

PR #1768, opened 2026-06-18
csv reader to use _variables metadata file

### LINKED ISSUE #46: Resolve issues with type conversion
An example in the detailed report:

## CORE-000004 — negative / 01

**Expected:** 2  **Actual:** 2

```diff
DIFF_FOUND for negative/01:
  Row 1: expected=('EC', '7', 'ECDOSE', '0') -> actual=('EC', '7', 'ECDOSE', '0.0')
```

I'm not sure which reflects the actual source data, but we need to resolve these float/int issues.

SFJohnson24:
this PR brings _variables.csv into the reader class and uses that file for datatyping the corresponding columns.  It falls back to pandas inference should the column not be defined in the file/has parity with the excel reader.


cdisc_rules_engine/interfaces/data_reader_interface.py
```diff
@@ -8,14 +8,19 @@ class DataReaderInterface:
     """
 
     def __init__(
-        self, dataset_implementation=PandasDataset, encoding: str = DEFAULT_ENCODING
+        self,
+        dataset_implementation=PandasDataset,
+        encoding: str = DEFAULT_ENCODING,
+        variables_csv_path: str = None,
```
gerrycampion: this line is implementation-specific and should go in the csv_reader class, not the interface

gerrycampion (changes requested):

gerrycampion (approved):

gerrycampion:
issues with tests, otherwise looks good

PR #1733, opened 2026-05-13
normalize standards

SFJohnson24:

PR #1770, opened 2026-06-21
Add ADaM product standards to StandardTypes CLI gate

antonybevan:
## Summary

`StandardTypes` is the allow-list the CLI uses to validate `-s/--standard`. When
`--custom-standard` is not set, `core.py` rejects any value not in
`StandardTypes.values()` and exits with code 2. The enum lists `adamig` but not the
other six ADaM products defined in `ADAM_PRODUCTS`
(`cdisc_rules_engine/constants/adam_products.py`): `adam-adae`, `adam-md`,
`adam-nca`, `adam-occds`, `adam-tte`, `adam-poppk`.

Those six are already handled by `normalize_standard_input()` in
`cdisc_rules_engine/utilities/utils.py`, which maps any value in `ADAM_PRODUCTS` to
`("adam", "<product>-<version>")`. Because the gate does not list them, a command
such as `-s adam-tte -v 1-0` is rejected before the engine runs, even though the
engine supports it.

This adds the six products to `StandardTypes`, completing the normalization started
in #1733 (which added `adamig` and the `sendig-*` products). The values mirror
`ADAM_PRODUCTS`.

## Reproduction (current `main`)

```
core validate -s adam-tte -v 1-0 -d <datasets> -ft xpt ...
# Standard 'adam-tte' is not a supported standard. ... (exit code 2)
```

## Changes

- `cdisc_rules_engine/enums/standard_types.py`: add the six `adam-*` products,
  following the pattern used for the `sendig-*` products.
- `tests/unit/test_utilities/test_standard_types_adam.py`: assert that every
  `ADAM_PRODUCTS` entry is a valid `StandardTypes` value, so the gate and the
  product list cannot drift apart again.

## Testing

- `black` and `flake8` pass on both files.
- The new test passes.


SFJohnson24 (approved):
this PR correctly implements the missing adam products in the standard types

PR #1745, opened 2026-05-28
Support pandas 3.0

filippsatverily:
## Summary

Upgrades cdisc-rules-engine to support pandas 3.0. Stacked on top of #1713 (relax dependency constraints) — please merge that first.

- Drop the `pandas <3.0` upper bound and add `pytz` (no longer a pandas transitive dep)
- Replace `applymap()` with `map()` (removed in pandas 3.0)
- Replace `inplace=True` mutation patterns (pandas 3.0 Copy-on-Write)
- Handle pandas 3.0 default `StringDtype` in comparison operators
- Handle extension arrays in `DaskDataset.__setitem__`
- Remove unsupported `dd.DataFrame` type annotation in `parquet_reader`
- Remove `method=` and `downcast=` kwargs removed in pandas 3.0
- Replace Dask GroupBy `.apply(set)` path in `Distinct` operation

PR #1777, opened 2026-06-22
fix: handle pandas 3.0 default StringDtype

filippsatverily:
Pandas 3.0 changes the default string dtype from `object` to `StringDtype`, which requires these changes:

1. Regex operators: `.map()` now returns nullable `BooleanDtype`, where `pd.NA & True` raises instead of returning `False`. Adds a `_map_regex()` helper that normalizes to numpy bool via `.fillna(False).astype(bool)`, used by all prefix/suffix/matches regex operators.

2. Case-insensitive comparisons: `.lower()` on a non-string value (e.g. `pd.NA`) raises `AttributeError`. Guards with `isinstance(target_val, str)` before calling `.lower()`.

3. Empty-column detection in record_count: checks `dtype == "object"` to identify string columns, which misses `StringDtype`. Uses `pd.api.types.is_string_dtype()` instead.

4. Date validation: simplifies the `is_valid_date` guard to `not isinstance(date_string, str)`, which already handles `None`, `pd.NA`, and any other non-string type.

Tested scenarios:
- Full pytest suite: 1746 passed, 11 skipped, 0 failed (pandas 2.3.3, dask 2025.12.0)
- Ran validation on [CDISC_Pilot_Study_v4_FIXED.json](https://github.com/cdisc-org/cdisc-jsonata-rules/blob/main/testdata/CDISC_Pilot_Study_v4_FIXED.json): 201 SUCCESS, 6 SKIPPED, 0 errors

### DIFF
--- a/cdisc_rules_engine/check_operators/dataframe_operators.py
+++ b/cdisc_rules_engine/check_operators/dataframe_operators.py
@@ -228,8 +228,16 @@ class DataframeType(BaseType):
             target_val = custom_str_conversion(target_val)
             comparison_val = custom_str_conversion(comparison_val)
         if case_insensitive:
-            target_val = target_val.lower() if target_val else None
-            comparison_val = comparison_val.lower() if comparison_val else None
+            target_val = (
+                target_val.lower()
+                if isinstance(target_val, str) and target_val
+                else None
+            )
+            comparison_val = (
+                comparison_val.lower()
+                if isinstance(comparison_val, str) and comparison_val
+                else None
+            )
             return target_val == comparison_val
         return target_val == comparison_val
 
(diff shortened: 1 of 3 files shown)


filippsatverily:
@SFJohnson24 another commit from #1745

SFJohnson24:
@filippsatverily it looks like this is failing our linter for dataframe_operators, can you reformat?

filippsatverily:
> @filippsatverily it looks like this is failing our linter for dataframe_operators, can you reformat?

@SFJohnson24 done

SFJohnson24 (approved):

SFJohnson24 (approved):

PR #1781, opened 2026-06-23
fix: replace Dask GroupBy path in Distinct (.apply(set) fails in pandas 3.0)

filippsatverily:
Pandas 3.0 changes the behavior of `.apply(set)` / `.apply(list)` on a GroupBy, returning a DataFrame instead of a Series. This breaks the Dask `else` branch in `Distinct._execute_operation`.

Instead of adding a Dask-specific code path, this materializes dask DataFrames to pandas at the top of the method, letting the existing pandas groupby/agg logic handle everything. This eliminates the Dask-specific `else` branch entirely and also resolves 11 dask GroupBy `FutureWarning` instances (145 → 134 warnings).

Tested scenarios:
- Full pytest suite: 1746 passed, 11 skipped, 0 failed (pandas 2.3.3, dask 2025.12.0)
- Ran validation on [CDISC_Pilot_Study_v4_FIXED.json](https://github.com/cdisc-org/cdisc-jsonata-rules/blob/main/testdata/CDISC_Pilot_Study_v4_FIXED.json): 201 SUCCESS, 6 SKIPPED, 0 errors

PR #1786, opened 2026-06-24
Filipps/pandas3 fix distinct groupby

SFJohnson24:
see https://github.com/cdisc-org/cdisc-rules-engine/pull/1781 for change explanations

### DIFF
--- a/cdisc_rules_engine/operations/base_operation.py
+++ b/cdisc_rules_engine/operations/base_operation.py
@@ -131,7 +131,6 @@ class BaseOperation:
             return self.evaluation_dataset
 
     def _handle_grouped_result(self, result):
-        # Handle grouped results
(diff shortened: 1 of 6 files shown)


RamilCDISC (approved):
The PR fixes the dask groupby path in Distinct which was failing in pandas 3.0 The PR fixes it by handling the datatype returned by GroupBy. Passing of all CI checks proves proper integration and no breaking changes.

filippsatverily:
@SFJohnson24 I did some testing on this PR and found that `test_grouped_distinct[data1-expected1-None]` fails _some_ of the test runs with `Error: assert [18, 12] == [12, 18]` when I bump pandas to `3.0.0+`.

SFJohnson24:
@filippsatverily I will set up a second venv with pandas 3.xx and test

PR #1787, opened 2026-06-24
Added type_insensitive parameter for equals_string_part operator

### LINKED ISSUE #1665: Allow type_insensitive parameter for does_not_equal_string_part and equals_string_part operators
### Feature Description

Comparing Char and Num variables (eg --ORNRLO versus --STNRLO) using the "does_not_equal_string_part" and "equals_string_part" operators currently does not work. Also "type_insensitive: true" which can be used in combination with "equal_to" operator doesn't work with the "does_not_equal_string_part" and "equals_string_part" operators.

For our custom rules, we want to compare Char and Num variables using the "does_not_equal_string_part" and "equals_string_part" operators (for example: extract numeric part of --ORNRLO (char) and compare to --STNRLO (num)). Would it be possible to allow type_insensitive in combination with these operators? Thanks for looking into this!

Image

aniemes:


resources/schema/rule/Operator.md
```diff
@@ -156,17 +156,21 @@ Text-based operations including regex pattern matching, substring operations, pr
...
 
-Complement of `equals_string_part`
+Complement of `equals_string_part`. Also has the optional parameter 'type_insensitive'.
 
 ### equals_string_part
 
 Checks that the values in the target column equal the result of parsing the value in the comparison column with a regex
+Has optional parameter:
+
+- 'type_insensitive' when true, both values are converted to strings before comparison to handle type mismatches between string and numeric data. NOTE: all trailing zeroes will be removed in both strings and floats.
```
RamilCDISC: I think we should mention the behavior for leading zeros too along with trialing zeros.

tests/unit/test_check_operators/test_string_comparison.py
RamilCDISC, line 1: The PR looks great to me. I will recommend minor updates to the tests for covering the edge cases of 
1. No regex match
2. Empty string/null target
3. Case for deos_not_equal_string_part False
4. Test for leading zero behavior.

RamilCDISC (changes requested):

tests/unit/test_check_operators/test_string_comparison.py
aniemes: Hi Ramil,
Thank you for reviewing it!  I will work on the changes now to cover the edge cases and update the Operator.md file as you said above.
Thanks!
Adam

aniemes: @RamilCDISC 
So I am having some trouble with the 'no regex' as this seems to be expected.  I added this into a new test case:
`    if operator == "equals_string_part":
        result = dataframe_type.equals_string_part(
            {
                "target": "target",
                "comparator": comparator,
                "regex": "(.*)",
                "type_insensitive": True,
            }
        )
`

But this is still using regex.  Can you suggest how to test this without using regex?  I am also have trouble with the empty string/null targets.

I tried this:
`
        (
            {"VAR2": ["", ""], "target": [0, 0]},
            "VAR2",
            "equals_string_part",
            PandasDataset,
            [True, True],
        )
`
And an empty target array but neither worked.

Pasted test file for reference.

[test_string_comparison.py](https://github.com/user-attachments/files/29356957/test_string_comparison.py)

RamilCDISC: My intention with 'no regex match' was to ask for a test where the regex does not return any match, not to test without a regex. I see that my comment was not very clear. The edge cases I requested is to have a case where the passed regex does not return any match. Passing a regex is required.

For the null/empty target case, your example has empty values in VAR2 not in the target. target is '[0,0]'. I would test it with explicit empty/null values in target columns.

aniemes: Ok this makes sense, I will work on these changes.  Thank you!

aniemes: I updated and I think they are working properly.  Just let me know if anything needs adjusting...thanks!

aniemes: After updating and syncing the check_python_files script is failing but this was not the case before.  Is this from my changes?

RamilCDISC:
The PR looks fine to me now. Before approving could you please merge the updated schema with  markdown descriptions. and also fix the lint issues so lint tests pass in CI?

gerrycampion (approved):

PR #1791, opened 2026-06-27
feat: add does_not_start_with and does_not_end_with operators

### LINKED ISSUE #1153: Please also add "does_not_start_with" and "does_not_end_with" operators
As a complement to "starts_with" and "ends_with", can also "does_not_start_with" and "does_not_end_with" added as operators?

We can use that very well for custom rules.

Rdag15:
## Summary

Adds two new string operators, `does_not_start_with` and `does_not_end_with`, the negated complements of the existing `starts_with` / `ends_with` operators.

Closes #1153.

## Implementation

The operators follow the negation pattern already established in `dataframe_operators.py` (e.g. `does_not_contain` is `~self.contains(...)` and `does_not_equal_string_part` is `~self.equals_string_part(...)`):

```python
@log_operator_execution
def does_not_start_with(self, other_value):
    return ~self.starts_with(other_value)

@log_operator_execution
def does_not_end_with(self, other_value):
    return ~self.ends_with(other_value)
```

## Changes

- **`cdisc_rules_engine/check_operators/dataframe_operators.py`** — the two operator methods
- **`resources/schema/rule/Operator.json`** — operator enum entries
- **`resources/schema/rule/Operator.md`** — operator documentation
- **`resources/schema/rule-merged/Operator.json`** — regenerated with `python scripts/merge_schema_markdown.py` followed by `npx prettier ... --write` (not hand-edited), so it stays in sync with `check-schema-markdown`
- **`tests/unit/test_check_operators/test_string_comparison.py`** — parametrized tests for both operators across `PandasDataset` and `DaskDataset`

## Testing

- `pytest tests/unit/test_check_operators/test_string_comparison.py` — 77 passed (incl. 4 new cases)
- `black --check` — clean
- `flake8` — clean
- merged schema regenerated via the project script, so `check-schema-markdown` stays green


cdisc_rules_engine/check_operators/dataframe_operators.py
```diff
@@ -857,6 +857,16 @@ def ends_with(self, other_value):
...
+    @log_operator_execution
+    @type_operator(FIELD_DATAFRAME)
+    def does_not_start_with(self, other_value):
+        return ~self.starts_with(other_value)
+
+    @log_operator_execution
+    @type_operator(FIELD_DATAFRAME)
+    def does_not_end_with(self, other_value):
+        return ~self.ends_with(other_value)
+
```
RamilCDISC, line 878: Starts_with and ends_with currently returns raw result of pandas string methods. These methods can produce nullable/objects results when target values is missing. This can mix badly with use of '~'. Other operators safely use '~' when positive has already normalized its output to boolean series. Please make starts_with and ends_with null safe.

tests/unit/test_check_operators/test_string_comparison.py
RamilCDISC, line 1: Could you please add cases for when 'value_is_literal=True'?

RamilCDISC (changes requested):

cdisc_rules_engine/check_operators/dataframe_operators.py
Rdag15: Fixed in f865a42. Confirmed: with a missing target value, `starts_with` returned `[True, None, False]` (object dtype) and `does_not_start_with` then raised `TypeError: bad operand type for unary ~: 'NoneType'`. Both `starts_with` and `ends_with` now pass `na=False` to `str.startswith`/`str.endswith`, so the result is always a clean boolean Series (missing -> False) that negates safely. Verified on PandasDataset and DaskDataset.

tests/unit/test_check_operators/test_string_comparison.py
Rdag15: Added in f865a42: `test_starts_with_literal` and `test_ends_with_literal` cover `value_is_literal=True` for both operators and their negations, plus `test_does_not_start_with_is_null_safe` / `test_does_not_end_with_is_null_safe` for the missing-value case — all parametrized across PandasDataset and DaskDataset.

RamilCDISC (approved):
The PR adds does_not_start and does_not_end with operators to CORE. These operators are negatives of starts_with and ends_with operators. The PR was validated by: 

2. Reviewing the PR in accordance to the AC.
3. Reviewing the code to verify it follows the previous implemented code standards.
4. Ensuring the documentation is updated.
5. Ensuring the documentation mentions the use and explanation correctly.
6. Ensuring the schema is updated.
7. Ensuring previous tests are updated and new tests are added for the new operators.
8. Ensuring the new tests cover null and empty cases.
9. Ensuring the existing operators are updated to work with '~' in the negative operators.
10. Ensuring all unit and regression tests pass.

tests/unit/test_check_operators/test_string_comparison.py
```diff
@@ -134,6 +134,121 @@ def test_ends_with(data, comparator, dataset_type, expected_result):
...
+
+
+@pytest.mark.parametrize(
+    "data,comparator,dataset_type,expected_result",
+    [
+        (
+            {"target": ["Att", "Btt", "Ctt"], "VAR2": ["tt", "B", "D"]},
+            "VAR2",
+            PandasDataset,
+            [False, False, False],
```
gerrycampion, line 374: Am I misunderstanding, or shouldn't these be `[False, True, True]`

gerrycampion (changes requested):

tests/unit/test_check_operators/test_string_comparison.py
Rdag15: Good question — I had to check this one carefully. The expectation matches actual behavior, but the behavior itself is surprising.

When the comparator is a **column**, `ends_with` converts it to a tuple, and `Series.str.endswith(tuple)` matches "ends with **any** element of the tuple" rather than comparing row-wise:

```python
comparison_data: Tuple[str] = tuple(comparison_data)
results = self.value[target].str.endswith(comparison_data, na=False)
```

With `target = ["Att","Btt","Ctt"]` and `VAR2 = ["tt","B","D"]`, every target ends with `"tt"`, so `ends_with` is `[True, True, True]` and `does_not_end_with` is `[False, False, False]`.

`[False, True, True]` would be right under row-wise semantics — which is what I'd expect too. That's the pre-existing behavior of `starts_with`/`ends_with` (the tuple conversion and its comment predate this PR), and the negations just mirror it.

I've added a comment on the tests explaining this. If you'd prefer row-wise semantics for column comparators, that's a behavior change to the positive operators as well — happy to open a separate issue rather than fold it into this PR. Let me know which you'd like.

PR #1794, opened 2026-06-30
Added TUMOR to allowed datasets

### LINKED ISSUE #950: Rule blocked: CORERULES-9356 (FB8401) and several others - tumor.xpt - add support
A number of FDA business rules that are specific for SEND (FB8401, FB8101, FB7901, FB7801, ...) use "tumor.xpt" dataset.
"TUMOR" is however not a CDISC/SEND domain, so I cannot add it to "Domains - Include".

aniemes:


RamilCDISC (approved):
The PR adds 'TUMOR' to list of allowed datasets in the schema. The schema is updated   properly and will allow users to specify 'TUMOR' as a dataset option.

gerrycampion (approved):

PR #1799, opened 2026-07-07
don't strip results values

gerrycampion:
This fixes unnecessary stripping as seen in [CORE-000019](https://github.com/cdisc-org/cdisc-open-rules/pull/60/changes#diff-f7b8a4fc6936acc0f0a2e5c8b8087c540cd85a66c65bdcac35a1724f2a46cb6d)

Error count went down here: https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/28888155574
(319->315)

### DIFF
--- a/cdisc_rules_engine/services/reporting/base_report_data.py
+++ b/cdisc_rules_engine/services/reporting/base_report_data.py
@@ -49,8 +49,8 @@ class BaseReportData(ABC):
             if value is None:
                 processed_values.append(null_placeholder)
                 continue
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
The PR removes unnecessary stripping which was failing the rules that required leading or trialing spaces. The update fixes the behavior to correctly identify the case where stripping is necessary and leaves as it is otherwise. The reduction in error count on the execution of the rules supports the fix.

Issue #168, opened 2022-10-13
not_exists for variable presence only works when there is a second variable check


JozefAerts:
When I have the code:

Check:
  all:
    - name: SPECIES
      operator: not_exists
    - name: USUBJID
      operator: exists

The check works correctly when "SPECIES" is not a column in the dataset.
When I then remove or comment out the last 2 lines (check on USUBJID), I get an error "Output variables: ['SPECIES'] not found in dataset".
It looks as in that case, the validator tries to provide the value of "SPECIES" in the output, but fails, as of course the column is not there - that is what we are testing for.
[SENDIG_105.yaml.txt](https://github.com/cdisc-org/cdisc-rules-engine/files/9776719/SENDIG_105.yaml.txt)
[unit-test-coreid-SENDIG105-negative.xlsx](https://github.com/cdisc-org/cdisc-rules-engine/files/9776727/unit-test-coreid-SENDIG105-negative.xlsx)

gerrycampion:
From related ticket:
Links to related JIRA Tickets

https://jira.cdisc.org/browse/CORERULES-5852
Rule Information

Standard: SDTMIG
Rule ID: SD1109
Rule Description: Exposure (EX) dataset should be included in every submission.
Describe the bug
Refer to SD1109 and corresponding negative test data.

Rule Type: Domain Presence Check
Check:
  all:
    - name: EX
      operator: not_exists
The issue is that the Domain Presence Check creates a dataframe variable for each dataset that exists in the study. But we are looking for a domain that is missing from the study. So when the check fires an issue, it automatically includes EX as an Output Variable even though it does not exist in the dataframe.

Error returned from Rule Engine

{
"error":"An unknown exception has occurred"
"message":"Output variables: ['EX'] not found in dataset"
}
Expected behavior
There needs to be a way to run this type of rule without an error. One possibility is to provide a default value of "Missing" when an output variable is missing.

PR #1800, opened 2026-07-07
Show target var for not_exists

gerrycampion:
Related to:
https://github.com/cdisc-org/cdisc-rules-engine/issues/168
https://github.com/cdisc-org/cdisc-rules-engine/pull/714

But these are fixed in a better way now, by providing a value of "Not in dataset".
This should fix the issues found in Rule [CORE-000029](https://github.com/cdisc-org/cdisc-open-rules/pull/60/changes#diff-e38b7f98bfecc4bf6f6e94acbb58bb5902e0a26b9a649fbae78868a8d4003586)

Needs this test suite PR: https://github.com/cdisc-org/CORE_Test_Suite/pull/101

Action run here: https://github.com/cdisc-org/cdisc-rules-engine/actions/runs/28894280239

### DIFF
--- a/cdisc_rules_engine/utilities/rule_processor.py
+++ b/cdisc_rules_engine/utilities/rule_processor.py
@@ -683,8 +683,6 @@ class RuleProcessor:
         seen: set[str] = set()
         conditions: ConditionInterface = rule["conditions"]
         for condition in conditions.values():
-            if condition.get("operator") == "not_exists":
-                continue
             target: str = condition["value"].get("target")
             if target is None:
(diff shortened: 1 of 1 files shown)


RamilCDISC (approved):
The PR removes a temporary fix made previously for missing columns in output variables when using not_exists operator as a better solution of putting 'Not in dataset' is already implemented in the engine. The removal of this fix is validated by running a validation against a rule mentioned in the description. Passing of all unit and regression tests supports the review.

PR #1802, opened 2026-07-09
2 fixes

SFJohnson24:
this PR fixes 2 issues found in the test data--
- catches when no rules are selected to execute (in test, usually due to .env vs rule standard/version scoping; instead return an error)
- if -d is given for USDM and multiple json files are found this it will execute the first json file, now logs an error and returns blank.

removes 1 QA test that was not executing any rules and was now returning an error; added unit test


core.py
```diff
@@ -154,8 +154,86 @@ def _validate_csv_data_paths(
...
+    """Gathers provided dataset paths plus any sibling datasets.csv/variables.csv files."""
+    all_files_in_dp = []
+    for dp in dataset_path:
+        all_files_in_dp.append(dp)
+        dp_path = Path(dp)
+        all_files_in_dp.extend(
+            [
+                str(p)
+                for p in dp_path.parent.glob("*")
+                if p.is_file() and p.name in {"datasets.csv", "variables.csv"}
```
gerrycampion: should these have leading underscores?

gerrycampion (changes requested):

core.py
SFJohnson24: resolved

scripts/run_validation.py
```diff
@@ -194,6 +194,11 @@ def run_validation(args: Validation_args):
         engine_logger.info(
             f"Running {len(rules)} rules against {len(datasets)} datasets"
         )
+        if len(rules) == 0:
```
RamilCDISC: The check is happening after the engine_logger has already logged length of the rules, that would show running 0 rules against number of datasets. It also happens after the datasets has been loaded/converted to parquet if needed. With zero rules it will just end up getting aborted. If we move this check a bit higher, it can help save extra work done on datasets.

scripts/run_validation.py
```diff
@@ -194,6 +194,11 @@ def run_validation(args: Validation_args):
...
+            raise ValueError(
```
RamilCDISC: From the content of the error message it seems the intention is to print one line instead of full trace. But raise ValueError here will actually print whole python trace as it is not caught or handled anywhere in this file or core.py file. If full trace is intentional please let me know.

RamilCDISC (changes requested):

scripts/run_validation.py
SFJohnson24: @RamilCDISC i think a stacktrace here is fine.  It will be 2 lines core.py -> run_validation then the actual log.

PR #1803, opened 2026-07-10
csv report errors

SFJohnson24:
fixes bug where errors were not reported on CSV and a successful execution versus an error execution looked identical

[CORE-Report-2026-07-10T14-58-03.csv](https://github.com/user-attachments/files/29903135/CORE-Report-2026-07-10T14-58-03.csv)
[CORE-Report-2026-07-10T15-13-33.csv](https://github.com/user-attachments/files/29903136/CORE-Report-2026-07-10T15-13-33.csv)

### DIFF
--- a/cdisc_rules_engine/services/reporting/sdtm_report_data.py
+++ b/cdisc_rules_engine/services/reporting/sdtm_report_data.py
@@ -363,15 +363,16 @@ class SDTMReportData(BaseReportData):
                         dataset = dataset_val.removesuffix(".csv")
                         csv_value = error.get("values") or error.get("message") or ""
                         rows.append([dataset, "", "EXECUTION_ERROR", csv_value])
-                else:
-                    for issue in self._issue_details(validation_result, result):
-                        dataset = (issue.get("dataset") or "").removesuffix(".csv")
-                        record = str(issue.get("row", ""))
-                        variables = issue.get("variables") or []
-                        values = issue.get("values") or []
-                        for variable, value in zip(variables, values):
-                            csv_value = "" if value in (None, "null") else value
-                            rows.append([dataset, record, variable, csv_value])
+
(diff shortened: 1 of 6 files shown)


cdisc_rules_engine/services/reporting/sdtm_report_data.py
```diff
@@ -350,6 +350,20 @@ def _generate_error_details(
     def _get_csv_rows(self) -> tuple[list[str], list[list[str]]]:
         header = ["Dataset", "Record", "Variable", "Value"]
         rows = []
+        for validation_result in self._results:
+            for result in validation_result.results or []:
+                if (
+                    result.get("executionStatus")
+                    == ExecutionStatus.EXECUTION_ERROR.value
+                ):
+                    for error in self._error_details(validation_result, result):
```
RamilCDISC: the _error_details() function is already called once per result in _generate_error_details() function during __init__ (to build data_sheets["Issue_details"] and is now called here again for the same result. Could we compute it once and cache/resuse it instead of running it twice?

RamilCDISC: This loop also processes the EXECUTION_ERROR entries that already exist in the ISSUE_DETAILS, but we only skip re-adding them because their variables/attributes field happen to be "" so zip() produces nothing,. If this empty field behavior ever changes this will silently start double counting the execution errors in the csv.

RamilCDISC (changes requested):

RamilCDISC (approved):
The PR resolves some csv report errors. The validation was done by: 

3. Reviewing the PR for updated tests.
4. Reviewing the attached csv reports.
5. Ensuring all unit and regression testing pass. 
6. Ensuring results are coherent after update.
7. Ensure the update made is not fragile and ensures proper working for edge cases.
8. Ensure full integration in existing system.

PR #1806, opened 2026-07-20
1443 Send49 Rule Blocked Fix

### LINKED ISSUE #1443: Rule blocked: SEND49
[unit-test-coreid-SENDIG_49.xlsx](https://github.com/user-attachments/files/23627949/unit-test-coreid-SENDIG_49.xlsx)

In SEND49 we need to test whether, when the SENDIG states that there is a codelists associated to a variable, it must be declared in the define.xml, either on the variable level or at the valuelist level.
So, we want to first check on the variable level, where we use:
- name: 'library_variable_ccode'
   operator: non_empty
- name: define_variable_ccode
   operator: not_equal_to
   value: library_variable_ccode
if the result of the second is "false", we need to further check on the valuelist level.
So we change/add (full code is in the rule-editor):
- name: 'library_variable_ccode'
      operator: non_empty
  - any:
      - name: define_vlm_ccode
         operator: not_equal_to
         value: 'test'
      - name: define_variable_ccode
         operator: empty
      - name: define_variable_ccode
         operator: not_equal_to
         value: library_variable_ccode

but then we get a message "define_vlm_ccode":"Not in dataset"
As "define_vlm_ccode" is defined as "ValueListDef.ItemDef.CodeList.Alias.Name" it looks as it acts on a single ValueList-ItemDef. But as we start by checking on a variable ItemDef, it looks as the system does not recognize that a "ValueList" is added to the variable (ItemDef). 

The request is to develop something so that we can check both on the variable level as on the valuelist level, i.e. when we start at the variable level, it is recognized when there is a valuelist attached to that variable, and that for each item in the valuelist, we can check on presence of a codelist and whether this codelist is the correct one to be used.

I attach a test file "define_negative_VSORRESU_ValueList_test.xml" where for VSORRESU, the codelist has been "delegated" to the ValueList level ItemDefs.
If more test files are needed, please let me know.

[define_negative_VSORRESU_ValueList_test.xml](https://github.com/user-attachments/files/23627935/define_negative_VSORRESU_ValueList_test.xml)

aniemes:
Addressing [#1443](https://github.com/cdisc-org/cdisc-rules-engine/issues/1443)

Added Value Level Metadata support for the Variable Metadata Check against Define XML and Library Metadata.

Extended VariablesMetadataWithDefineAndLibraryDatasetBuilder to extract and aggregate VLM data from Define xml.
Added metadata variables: define_vlm_present, define_vlm_item_count, define_vlm_ccodes, define_vlm_has_codelist_any, define_vlm_has_codelist_all, define_vlm_ccode_missing_any, define_vlm_ccode_matches_library_any, define_vlm_ccode_matches_library_all

Updated the schema documentation
Created regression test for the issue and tested flagging and none flagging scenarios.


cdisc_rules_engine/dataset_builders/variables_metadata_with_define_and_library_dataset_builder.py
```diff
@@ -95,6 +104,103 @@ def build(self):
...
+        vlm_summary = vlm_summary.rename(
+            columns={"define_variable_name": "variable_name"}
+        )
+        final_dataframe = final_dataframe.merge(
+            vlm_summary,
+            how="left",
+            on="variable_name",
+        )
+        final_dataframe = final_dataframe.drop(
+            columns=["define_variable_name_y"], errors="ignore"
```
RamilCDISC: Given vlm-summary is renamed to variable_name. There is no define_variable_name column collision left at this point. This drop looks like a leftover form earlier iteration. Could you please remove it or add a comment explaining if its covering any edge case?

RamilCDISC (changes requested):

RamilCDISC (approved):
The PR adds updates to unblock rule Send49. The updates were validated by: 

1. Ensuring the PR has changes in accordance with AC.
2. Ensuring the PR does not have any unwanted code or comments.
3. Ensuring the updated code integrates with existing flow.
4. Ensuring previous coding standards for  implementation of rule type and operation updates are followed.
5. Ensuring new tests has been added.
6. Ensuring all previous unit and regression testing pass.
7. Ensuring no artifacts from old code remains.
8. Running manual validation using positive and negative datasets through the branch.
9. Validating the output of the rule execution to ensure correct results.

PR #1810, opened 2026-07-23
Added regex_find_replace operation for Issue587 coverage

### LINKED ISSUE #587: Rule blocked: CORERULES-1689
- 
- ADaM Rule 64 Jira ID CORERULES-1689

**Rule Information**
- **Standard**: ADaM 
- **Rule ID**: 64
- **Rule Description**: The rule checks for presence of TRTxxAN and non presence of TRTxxA in ADSL

**Describe the problem**
How to handle the subscripts on the variable names metadata, which xx ranges from 01-99. If the numerical variable is present e.g TRT01AN then TRT01A must also be present. 

**Describe the solution**
Solution was discussed at weekly meet, where variables would be extracted and compared to/matched to a regular expression to confirm presence.  This would use the exisiting reg_match operator to compare against general pattern e.g  'value: ^TRT[0-9]{2}AN*' .  This part currently exists.
  
However the solution proposes a new operation using regex via a find and replace operation using regex capture buffers to extract the xx value from the primary variable and use this to create the shorten name (minus the "n") of the second variable.  This will then be used to define the second predicate in the rule logic.

**Proposed rule logic**
```yaml```

Check:
  all:
    - name: variable_name # e.g. TRT11AN
      operator: matches_regex # test 1 exist clause
      value: ^TRT[0-9]{2}AN*$
#new test clause using the new operation
    - name: $generated_variable_name # TRT11A
      operator: is_not_contained_by
      value: variable_name # e.g. returns column vector [TRT11AN, ABCDE, FGHIJ]
Operations:
  - find: ^TRT([0-9]{2})AN*$
    id: $generated_variable_name
    name: variable_name
    operator: regex_find_replace
    replace: TRT\1A # uses capture buffer to extract xx and create secondary variable.

**Screenshots**

### PR DESCRIPTION
**Pull request by aniemes**
Fix for issue [587](https://github.com/cdisc-org/cdisc-rules-engine/issues/587)
- added the regex_find_replace operation and wired it into operation parsing and registration
- updated rule schema support for regex_find_replace
- added unit coverage for the new operation
- added CoreIssue587 regression coverage with dataset fixtures for paired, missing, boundary, mixed, and noise cases

### CODE REVIEW

File: resources/schema/rule-merged/CORE-bundled.json
Comment (RamilCDISC) [inline resources/schema/rule-merged/CORE-bundled.json:1]: Could you please mention why is this bundled version being added here?

File: cdisc_rules_engine/operations/regex_find_replace.py [hunk 1]
```diff
@@ -0,0 +1,98 @@
...
+            raise OperationError(f"Invalid regex pattern '{find}': {exc}") from exc
+
+    def _transform_value(self, value, pattern, replace, on_no_match):
+        if value is None or (isinstance(value, float) and pd.isna(value)):
+            return None
+
+        text = str(value)
+        match = pattern.search(text)
+        if match:
+            return pattern.sub(replace, text)
```
RamilCDISC: here if replace references a capture group that does not exist in find. The error will be propagated back as a raw error instead of `OperationError`. Could you please wrap the call same way as `_compile_pattern` is wrapped?

RamilCDISC (changes requested):

resources/schema/rule-merged/CORE-bundled.json
aniemes: Hi @RamilCDISC!
I added this file by running the bundle_json_schema.py script since Operations.json was updated to define the new regext_find_replace operator and the bundled file has to be in sync with the source file.

aniemes:
This was created with the incorrect HEAD, it has been replaced by PR 1814.

PR #1813, opened 2026-07-28
fixes

SFJohnson24:
this PR fixes CORE-000757 expected not matching actual.  DataFrame.iterrows() silently upcasts blank RELREC values (like an all-empty USUBJID column) from None to float NaN, and since filter_if_present treated NaN as a real filter to apply, the RELREC-to-CM merge in CORE-000757 wiped out the whole dataset

to test: 
run CORE-000757 on this branch with test data--no longer gives skip; gives expected result from dataset
also fixes CORE-000744 negative 01-04


CORE-Report-2026-07-28T14-00-32.csv
RamilCDISC, line 1: The CSV files seems like debug artifacts. Could you please remove them or mention in a comment why they were made part of the PR?

cdisc_rules_engine/utilities/data_processor.py
```diff
@@ -196,7 +199,7 @@ def merge_relrec_datasets(
             DataProcessor.merge_on_relrec_record(
                 relrec_row, left_dataset, datasets, dataset_preprocessor, wildcard
             )
-            for _, relrec_row in relrec_for_domain.iterrows()
+            for relrec_row in relrec_for_domain.to_dict("records")
```
RamilCDISC, line 202: The merge on relrec record function still has 
`def merge_on_relrec_record(
        relrec_row: pd.Series,
...`

the type hint is outdated as it is changed to type dict here now.

RamilCDISC (changes requested):

CORE-Report-2026-07-28T14-00-32.csv
SFJohnson24: they are artifacts-- I added https://github.com/cdisc-org/cdisc-rules-engine/pull/1802 the csv in the gitignore but i also added it to this branch

RamilCDISC (approved):
The PR fixes .itterrows() bug identifies and explained in the PR description. The updated code was validated by: 

1. Ensuring the PR is in accordance with AC.
2. Ensuring the PR doe snot have any unwanted code or comments.
3. Ensuring the PR updated code integrates with existing flow.
4. Ensuring the type hints and function calls chain to be maintained.
5. Ensuring all unit and regression tests pass.
6. Running manual validation on the tests data, using positive and negative datasets.
7. Ensuring the outputs of the rule execution are valid.

PR #1814, opened 2026-07-30
587 Core Rules 1689 Blocked Fix

### LINKED ISSUE #587: Rule blocked: CORERULES-1689

aniemes:
Created


aniemes:
This will replace PR #1810 as it was created with the incorrect HEAD

cdisc_rules_engine/operations/regex_find_replace.py
```diff
@@ -0,0 +1,103 @@
+import re
+import pandas as pd
+
+from cdisc_rules_engine.operations.base_operation import BaseOperation
+from cdisc_rules_engine.exceptions.custom_exceptions import OperationError
+
+
+class RegexFindReplace(BaseOperation):
+    _NO_MATCH_POLICIES = {"keep_original", "set_null", "set_empty", "error"}
+    _FLAG_MAP = {
```
RamilCDISC: Do you think it will be better to move these under constants folder to follow code practice of the repository?

cdisc_rules_engine/operations/regex_find_replace.py
```diff
@@ -0,0 +1,103 @@
...
+        "i": re.IGNORECASE,
+        "m": re.MULTILINE,
+        "s": re.DOTALL,
+    }
+
+    def _execute_operation(self):
```
RamilCDISC, line 13: Could you please add a docstring describing the operation as other implementations of _execute_operation does?

resources/schema/rule/Operations.json
RamilCDISC, line 1: Could you please also update the operations.md file along with this file?

RamilCDISC (changes requested):

RamilCDISC (approved):
The PR adds regex_find_replace operation to unblock CORE_Rule 1689. The updated code was validated by: 

1. Review the PR for any unwanted code or comments.
2. Review the PR is in accordance with AC.
3. Review the PR for new implemented logic.
4. Review the PR for repository code practices.
5. Review the PR for already followed practice for addition of an operation.
6. Review the PR for added tests.
7. Ensure all unit and regression testing pass.
8. Ensure new regression and unit test cover the AC.
9. Ensure all relevant documentation is updated.
10. Ensuring the schema is updated to include this new operation.

PR #1815, opened 2026-07-30
Fix

SFJohnson24:
Bug 1 (crash): In dataset_preprocessor.py, the SUPP-domain child-merge logic (_merge_standard_then_filter_idvar/_merge_idvar_only) returned child_records.__class__.from_records(results) instead of a plain pd.DataFrame, causing a PandasDatase to wrap another PandasDataset  that crashed future function invocations due to the double wrap. 

Bug 2 (wrong data): In rule_processor.py, resolving an operation's "SUPP--" domain wildcard always picked the first startswith("SUPP") dataset in the study rather than the one actually being validated, so validating SUPPLB could silently compute against SUPPDM's data instead.

Test against CORE-000712 (positive/01, SUPPLB/SUPPDM).


RamilCDISC (approved):
The PR resolves bug caused by wrong return statement, and less strict check of SUPP datasets. The two changes made in PR fixes the bugs. All unit and regression tests pass confirming proper integration.

PR #1819, opened 2026-08-06
To fix issue 930: adding is_ordered_set_operator

### LINKED ISSUE #930: New operator request to check sequential ordering and absence of skips.
[unit-test-sequential SEQ_Negative.xlsx](https://github.com/user-attachments/files/17979812/unit-test-sequential.SEQ_Negative.xlsx)

Problem:
The operator 'is_ordered_set' currently doesn't check skips in numbering. It does check if for example: --SEQ is completed sequentially, but it doesn't check that a number is missing in the sequence. See attached test data: LB has a number skip in --SEQ but it is not picked up if you would write:

Check:
  all:
    - name: --SEQ
      operator: is_not_ordered_set
      value: USUBJID

Possible solution
Would it be possible to create an additional operator that is an extension on 'is_ordered_set/is_not_ordered_set' so that includes checking for skips as well?

aniemes:
To address issue #930

### DIFF
--- a/cdisc_rules_engine/check_operators/dataframe_operators.py
+++ b/cdisc_rules_engine/check_operators/dataframe_operators.py
@@ -1402,6 +1402,172 @@ class DataframeType(BaseType):
     def is_not_ordered_set(self, other_value):
         return ~self.is_ordered_set(other_value)
 
+    def _coerce_sequence_value_to_int(self, value):
+        """
+        Attempts to coerce a sequence value to an integer for continuity checks.
+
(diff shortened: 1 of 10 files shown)


aniemes:
The failing Deploy Rule Tester Endpoint step appears unrelated to code changes. The same step passed earlier with pip 26.1.2 and later failed with pip 26.2.x using the same pip-tools 7.6.0, failing inside pip-tools import logic. This looks like external toolchain drift.

PR #1820, opened 2026-08-10
484

SFJohnson24:
small PR for CORE-000484; it is dropping blank USUBJID but blank USUBJID in RELREC are allowed from my understanding.

for the negative/01 open rules test data it drops all the data where usubjid is blank and only reports the last 4 rows.  It should report all the rows for the test data as the RELID per USUBJID are unique


pendingintent (approved):
When validating against CORE conformance rules for a specific implementation, a rule author should double-check if the rule flags null USUBJID in RELREC as an error by default — some validation engines default to flagging it unless explicitly configured to allow the trial-design-relationship case.

The engine should allow null USUBJID so any rule should be authored considering which domains are being validated.

PR #1832, opened 2026-08-20
core-000107

SFJohnson24:


pendingintent (approved):
I updated the engine submodule in my cdisc-open-rules development environment and executed against my new test cases for CORE-000107 and I now see all of the expected results.

PR #1850, opened 2026-09-03
removed old code

### LINKED ISSUE #1846: execute_rule: triggered rules crash with stock PyPI deps (business_rules_enhanced passes `results=`; COREActions expects record-wise iteration)
## Environment

- `cdisc-rules-engine==0.17.1` (PyPI), Python 3.12
- `business_rules_enhanced==1.4.8` (pulled by the `>=1.4.8` requirement)
- Reproduced on Windows and Linux

## Summary

`RulesEngine.execute_rule` evaluates conditions correctly (row-wise, observed via a `PandasDataset` + `DatasetVariable`), but **any rule whose conditions trigger crashes in the action-dispatch path**, making the engine unable to complete a triggered rule with stock PyPI dependencies.

## Finding 1 ??action signature mismatch

`business_rules/engine.py` (`do_actions`, line ~119) calls
`method(**params, results=results)` unconditionally, but `COREActions.generate_record_message(self, message, target=None)` (`cdisc_rules_engine/models/actions.py`, ~line 38) does not accept `results` ??`TypeError` on every triggered rule.

## Finding 2 ??missing record-wise iteration

After working around Finding 1 (subclass adding `**kwargs` and forwarding), `generate_record_message` fails with:

```
AttributeError: 'PatchedCOREActions' object has no attribute 'record'
```

The action dereferences `self.record.get('row')`, i.e. it expects the runtime to set `.record` per row before invoking actions. The PyPI `business_rules_enhanced` 1.4.8 `do_actions` does not perform record-wise iteration.

## Minimal reproducer

```python
# pip install cdisc-rules-engine==0.17.1 pandas
import pandas as pd

from cdisc_rules_engine.models.dataset import PandasDataset
from cdisc_rules_engine.models.rule_conditions.condition_composite_factory import (
    ConditionCompositeFactory,
)
from cdisc_rules_engine.models.sdtm_dataset_metadata import SDTMDatasetMetadata
from cdisc_rules_engine.rules_engine import RulesEngine

df = pd.DataFrame({"STUDYID": ["S1", "S2", "S3"], "A": ["1", "", "3"]}).fillna("")

rule = {
    "core_id": "REPRO-1",
    "conditions": {
        "all": [
            {
                "name": "get_dataset",
                "operator": "exists",
                "value": {"target": "A", "comparator": "A"},
            }
        ]
    },
    "operations": [],
    "actions": [
        {
            "name": "generate_record_message",
            "params": {"message": "A is empty", "target": "A"},
        }
    ],
}

rule["conditions"] = ConditionCompositeFactory.get_condition_composite(rule["conditions"])
results = RulesEngine().execute_rule(
    rule,
    PandasDataset(df.copy()),
    SDTMDatasetMetadata(name="DM", filename="dm.xpt", record_count=len(df)),
)
print(results)
# -> TypeError: COREActions.generate_record_message() got an unexpected keyword argument 'results'
```

(Data path note: the same behavior was observed feeding a real SDTM domain via `pandas.read_sas(path, format="xport")`.)

## Question

Is the CORE application shipping a patched/forked `business_rules` runtime that provides the record-wise iteration and tolerates the `results` kwarg? If so, would you consider pinning that runtime as a wheel dependency so the MIT engine is runnable from PyPI as published? Happy to turn the reproducer into a unit test if useful.

SFJohnson24:
This PR removes 2 pieces of dead code in actions so as to prevent confusion with end users that these are maintained, valid actions.

### DIFF
--- a/cdisc_rules_engine/models/actions.py
+++ b/cdisc_rules_engine/models/actions.py
@@ -37,13 +37,6 @@ class COREActions(BaseActions):
         self.rule = rule
         self.value_level_metadata = value_level_metadata
 
-    @rule_action(params={"message": FIELD_TEXT, "target": FIELD_TEXT})
-    def generate_record_message(self, message, target=None):
-        full_message = f"Error in row {self.record.get('row')} " + message
-        if target:
-            full_message = full_message + f" Actual value: {self.record.get(target)}"
-        self.output_container.append(full_message)
-
     @rule_action(params={"message": FIELD_TEXT})
     def generate_dataset_error_objects(self, message: str, results: pd.Series):
         # leave only those columns where errors have been found
@@ -69,10 +62,6 @@ class COREActions(BaseActions):
                 error.dataset = DomainPresenceValues.DATASET.value
                 error.row = DomainPresenceValues.RECORD.value
         self.output_container.append(error_object.to_representation())
-
-    @rule_action(params={"message": FIELD_TEXT})
-    def generate_single_error(self, message):
-        self.output_container.append(message)
 
     def _get_target_names_from_list_values(
         self, target_names: List[str], rows_with_error: pd.DataFrame
--- a/tests/unit/test_rules_engine.py
+++ b/tests/unit/test_rules_engine.py
@@ -42,14 +42,6 @@ def test_get_schema():
     }
     assert dataset_check_schema in schema["variables"]
     assert "actions" in schema
-    generate_record_message_metadata = {
-        "name": "generate_record_message",
-        "label": "Generate Record Message",
-        "params": [
-            {"label": "Message", "name": "message", "fieldType": "text"},
-            {"label": "Target", "name": "target", "fieldType": "text"},
-        ],
-    }
     generate_dataset_error_objects_metadata = {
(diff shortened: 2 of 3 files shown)


pendingintent (approved):

cdisc_rules_engine/models/actions.py
OGarcia11, line 1: Removing `generate_record_message` and `generate_single_error` still leaves references to them in the test suite, which is why the `test` check is currently failing:

- `tests/unit/test_rules_engine.py` still asserts on `generate_record_message_metadata` and `generate_single_error_metadata` in the actions schema.
- `tests/unit/test_utilities/test_rule_processor.py` still has a test case built around `generate_record_message`.

If these are truly dead/unmaintained actions, we'll need to update or remove those tests as part of this PR. But it's also worth double-checking there's no rule content elsewhere (e.g. in `cdisc-open-rules`) still invoking these actions by name — deleting an action that's still referenced by a published rule would break that rule at runtime, not just in tests.

OGarcia11 (commented):
Flagging a couple of things before this merges — see inline comment.

PR #1854, opened 2026-09-10
Ap identification

### LINKED ISSUE #1837: APxx datasets giving false positives, even when -v 3-4 is given.
### Standard

CDISC SDTM/SDTMIG

### Reference Rule ID(s)

CORE-00376 CORE-000575 CORE-000701 CORE-000575 CORE-000852 CORE-000902

### Conformance Rule ID(s) (if published in CORE)

_No response_

### JIRA Ticket

_No response_

### CLI Command Used (if applicable)

### Rule YAML

```YAML
N/A
```

### Attach any sample test data file(s)

When working on an SDTMIG-3.4 submission for a customer that contains APMH (which is based on 3.2 I presume), we got a number of false positives for the above CORE rules.
Some of the rules even treat it as a "custom domain".
Essentially, as I gave -v 3-4, APMH should not have been evaluated at all, as it is based on 3.2, and also as it is a very old standard even breaking some of the principles of SDTM-1.4 on which SDTMIG-3.2 is based.

For users not realizing this, these false positives can be very confusing, even tempting them to "correct" their datasets.

My proposal is to, when an APxx dataset is in the list of datasets submitted to the engine, then these are excluded from the validation.
In the Excel tab "Dataset Details" , these could be marked as "Not evaluated".

I presume this is a just 1 or a few lines in the Python code.

### Attach any output report and/or log file(s)

N/A

### Expected output

See above

### Any Additional Information

_No response_

SFJohnson24:
This PR updates logic for identifying AP class which used to seek out the parent dataset then ID based on that.  This PR still uses that logic, if it fails, it attempts to use the AP suffix--testing it in the case of both standard and custom domains attached to AP--.

### DIFF
--- a/cdisc_rules_engine/models/sdtm_dataset_metadata.py
+++ b/cdisc_rules_engine/models/sdtm_dataset_metadata.py
@@ -32,6 +32,8 @@ class SDTMDatasetMetadata(DatasetMetadata):
     | APXX     | APXX         | False   | APXX   | XX                   | None    | True  | XX        | False            | XX             | True                     |
     | SQAPXX   | SQAPXX       | True    | None   |                      | APXX    | True  |           | False            | XX             | True                     |
     | FA       | FA           | False   | FA     | FA                   | None    | False |           | False            |                |                          |
+    | APRELSUB  | APRELSUB    | False   | None   | RELSUB               | None    | True  | RELSUB    | False            |                |                          |
+    | APRELSPEC | APRELSPEC   | False   | None   | RELSPEC              | None    | True  | RELSPEC   | False            |                |                          |
     """  # noqa: E501 W291
 
     @property
@@ -90,6 +92,8 @@ class SDTMDatasetMetadata(DatasetMetadata):
     def ap_suffix(self) -> str:
         """
         Returns the 2-character suffix (characters 3-4) from AP domains.
+        Falls back to the dataset name when there's no DOMAIN variable
+        (e.g. APRELSUB, APRELSPEC)
         Returns empty string if not an AP domain or for supp datasets.
         """
         if not self.is_ap:
@@ -98,4 +102,10 @@ class SDTMDatasetMetadata(DatasetMetadata):
             return ""
         if isinstance(self.domain, str) and len(self.domain) >= 4:
             return self.domain[2::]
+        if (
+            isinstance(self.name, str)
+            and len(self.name) >= 4
+            and self.name.upper().startswith("AP")
+        ):
+            return self.name[2::]
         return ""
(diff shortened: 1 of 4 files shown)


pendingintent (commented):
AP-- will not have an associated domain other than APRELSUB. MH will not have a relationship to APMH.  All AP domains are only to record details for a person associated with a USUBJID.  There is no parent/child relationship with AP--.  The suffix only indicates the type of data collected in the dataset.  In this case, this is medical history for the associated person.  The APID in APMH will join to APRELSUB.APID and the APRELSUB.USUBJID will indicate to which subject the person is associated.

SFJohnson24:
I removed the logic for AP that I preserved, that was built on a false premise.
I also updated ap_suffix for APRELSUB/APRELSPEC as it would return "" for ap_suffix.  In returning this, the the suffix is used to ID the dataset as RELSUB/RELSPEC and thus relationship.  It would not impact wildcard replacement as there are not wildcard variables in either dataset from what I can tell from library

I updated tests as the ap_suffix had several failing tests where there was no metadata on the dataset--meaning a dataset was submitted but completely empty.  It no longer returns "" and instead would error out given the total lack of metadata.  I would expect this given the blank data which would force a rule skip for that dataset and rule

SFJohnson24:
[AP-Domains.xlsx](https://github.com/user-attachments/files/32211768/AP-Domains.xlsx)
I tested CORE-000376 which used the custom_domain logic.  There was another bug there with AP that I resolved.  The attached data should be positive data and return no bugs running this rule.

pendingintent (commented):
[AP-Domains-negative.xlsx](https://github.com/user-attachments/files/32246139/AP-Domains-negative.xlsx)
[CORE-Report-2026-09-15T10-13-10.xlsx](https://github.com/user-attachments/files/32246142/CORE-Report-2026-09-15T10-13-10.xlsx)


I have executed the API-Applicable rules against the datasets in _AP-Domains-negative.xlsx_.

```bash
python core.py validate -s sdtmig -v 3-4 -r CORE-000181 -r CORE-000233 -r CORE-000234 -r CORE-000235 -r CORE-000778 -r CORE-000180 -r CORE-000510 -r CORE-000201 -r CORE-000376 -dp /Users/dmoreland/Downloads/AP-Domains.xlsx
```

I am seeing one issue.  CORE-000778 is showing a false positive - _Associated Persons non-supplemental qualifier dataset associated with a split dataset does not have a dataset name with a length greater than 4 and less than, or equal to, 6_. There is no split dataset included in the test data.

SFJohnson24:
note: the test suite failure is expected given the changes in scope (rules involving split and supp are failing)

PR #1860, opened 2026-09-21
fix CORE-000034

SFJohnson24:
updates the comparable logic for 

| Exp/Act       |   Result Row | Dataset   |   Record | Variable   | Value      |
|---------------|--------------|-----------|----------|------------|------------|
| Expected only |            4 | DS        |       21 | DSDECOD    | DEATH      |
| Expected only |            5 | DS        |       21 | DSSTDTC    | 2019-01-10 |
| Expected only |            6 | DS        |       21 | DTHDTC     |            |

which is missed with the new rule logic date_not_equal_to

Fixed compare_dates so that when one date is blank and the other has a value, date_equal_to/date_not_equal_to correctly return False/True respectively (they're definitively unequal, not just "uncertain"), while two blank dates still return False for both since there's nothing to compare. This resolves CORE-000034 under-reporting (6/9 rows) by properly flagging cases like record 21 where DSSTDTC has a value but DTHDTC is blank.


pendingintent (commented):
Fixes date_not_equal_to false-negatives (CORE-000034).  However, I ran this through a skill, and two potential underlying issues were identified:

1. [helpers.py:385](vscode-webview://1ui5b36rek1pubh5r7r7098ga319n2ed027nf6ubp78saqbbv8vc/cdisc_rules_engine/check_operators/helpers.py#L385) — when both dates are float('nan'), not target and not comparator is False because bool(nan) is True in Python. The code falls through to return operator_func is operator.ne, returning True unconditionally. Any row where both date fields are simply missing gets flagged as an error. Verified directly: compare_dates('auto', float('nan'), float('nan'), operator.ne) returns True on the PR branch.
2. [helpers.py:387](vscode-webview://1ui5b36rek1pubh5r7r7098ga319n2ed027nf6ubp78saqbbv8vc/cdisc_rules_engine/check_operators/helpers.py#L387) — two non-blank but unparseable date strings that are actually identical (e.g. both 'UNK') get reported as not-equal, because _dates_are_comparable returns False for either, and the fallthrough returns True for ne regardless of whether the raw values match. This conflates "missing" with "present but unparseable" (a different case).

You are more familiar with the code, but I thought I would add these issues for your review. I would be interested in knowing what you think about the skill findings.

pendingintent (approved):
Change looks fine and our discussion addressed the questions I had posted.

